PR #1, opened 2016-03-10
Added scripts, and a few features

SchrodingersGat:
Hi @pointhi I have added some scripts for creating various connectors.

Also I have added a few features / shortcuts, e.g.

> Added "addMountingHole" function for added NPTH holes
> Extended functionality of "createNumberedPadsTHT"
 - can provide an x_offset and y_offset for the starting position
 - can provide a starting pin number (1 = default)
 - can provide a pin-number increment (1 = default)

Let me know what you think. I'm quite liking your script - I had one in the works for exactly the same purpose but yours is much further along so I'm just going to extend yours where I can :)


kicad_mod/kicad_mod.py
```diff
@@ -209,21 +247,27 @@ def __str__(self):
 
         for pad in self.pad_array:
             output += self._savePad(pad)
+            
+        if (self.model):
```
pointhi: in python, you don't need brackets for such simple conditions

kicad_mod/kicad_mod.py
```diff
@@ -179,24 +207,34 @@ def _savePad(self, data):
...
-
+        
+    def _saveModel(self):
+        
+        #model path
+        output = "(model {model}\n".format(model=self.model)
+        #model position
+        output += "\t(at (xyz {x} {y} {z}))\n".format(x=self.model_pos['x'],y=self.model_pos['y'],z=self.model_pos['z'])
+        output += "\t(scale (xyz {x} {y} {z}))\n".format(x=self.model_scale['x'],y=self.model_scale['y'],z=self.model_scale['z'])
+        output += "\t(rotate (xyz {x} {y} {z})))\n".format(x=self.model_rot['x'],y=self.model_rot['y'],z=self.model_rot['z'])
```
pointhi: replace ```\t``` with the correct amount of spaces (look onto files which were created by kicad which means 2 spaces per indention level, and the last closing bracket is located in a new line).
The script output should be the same as kicad does it.

scripts/Hirose/df13_angled_pth.py
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env python
+
+import sys
+sys.path.append(r'''C:\kicad\fp-gen\kicad_mod''')
```
pointhi: this is an absolute path, which causes problems on every computer with another file hierarchy or non Windows Operating System.

Does my default code for loading the correct directory via relative path not working on Windows?

```python
sys.path.append("../../kicad_mod") # load kicad_mod path
```

SchrodingersGat:
Hey @pointhi , I've fixed all the import paths to be relative, however I had to do it a little bit differently to you, because I've been calling the scripts a bit differently to the way to do.

Basically, I cd to the .pretty directory, and call "python C:\kicad\fp-gen\scripts\jst\example_.py" which dumps all the scripts into the current directory. (which may be C:\kicad\share\kicad\pretty\Connectors_JST.pretty for example)

However when I do this, calling "sys.path.append("../../kicad_mod")" means that the directory relative to the current dir (which is the .pretty directory) is used, which is wrong.

So, my include line is now:

sys.path.append(os.path.join(sys.path[0],"..","..","kicad_mod"))

which means that a relative include will function correctly, no matter where the script is called from.

SchrodingersGat:
@pointhi I'm not going to persist with this PR as KicadModTree is so much better!

PR #4, opened 2016-03-17
Point grid

SchrodingersGat:
One of requirements for the courtyard is that it falls on an 0.05mm grid. I've added a "grid=" keyword (optional) to point/line/polygoneline/rectline to force the points of the lines to fall on the grid.


pointhi:
I also had similar ideas, but I think I do the implementation by myself.

Those functionality could be implemented in a simpler, and much more flexible way.

SchrodingersGat:
Did you ever implement the "grid" functionality - if so, where in the code is it implemented? I would like to transfer my scripts to using *your* grid implementation if it exists :)

pointhi:
Please rebase to the current master and remove those parts with default line width & co.

I do not longer know what idea I had before, and after rereading over this proposal it seems like a good way to implement this feature.

pointhi:
Furthermore, this code should be moved into the render method, otherwise there could be rounding errors.

SchrodingersGat:
I have fixed conflicts!

I have left the grid code where it is - I tried putting it in the render method but it was not getting called! I think somewhere a new Point object is being made and the grid data is lost...

SchrodingersGat:
Fixed conflicts with master branch

pointhi:
@SchrodingersGat
I added a ```round_to``` method to the Point class to support such things like aligning to a grid (as long as no transformations are applied). Is this Pull Request still of relevance?

https://kicad-footprint-generator.readthedocs.io/en/latest/KicadModTree.html#KicadModTree.Point.Point2D.round_to

SchrodingersGat:
That sounds sufficient, thanks :)

PR #5, opened 2016-03-18
Added my scripts

SchrodingersGat:
Various component generator scripts
Some use the old kicad_mod lib
Some use the new KicadModTree lib


pointhi:
thanks.

Could you also create a PR for your changes in ```kicad_mod```? Then I can add the little fix to have correct 3D-Model output.

thx

PR #6, opened 2016-04-05
Arc fix

SchrodingersGat:
Fix to the Arc node type, text rendering had 'center' instead of 'start'

### DIFF
--- a/KicadModTree/KicadFileHandler.py
+++ b/KicadModTree/KicadFileHandler.py
@@ -126,7 +126,7 @@ class KicadFileHandler(FileHandler):
 
     def serialize_Arc(self, node):
         render_strings = ['fp_arc']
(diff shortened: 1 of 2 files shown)


pointhi:
Hy, thanks for this fix, but probably it would be better to rename the attributes of the class to ```center``` and ```start```, which would describe the actual behaviour.

Although KiCad uses ```start``` and ```end``` in the file structure, it shouldn't confuse users of this library so I would propose to rename it to ```center``` and ```start```. What do you think?

SchrodingersGat:
I agree. I didn't know that KiCAD used 'start' to designate arc center and
it was a bit confusing. I'll adjust the PR :)

pointhi:
thx

PR #11, opened 2016-08-17
Added default line width for the circle

SchrodingersGat:
Simple PR that adds a default line width when adding a circle


pointhi:
this feature is already implemented in the file handler:
https://github.com/pointhi/kicad-footprint-generator/blob/master/KicadModTree/KicadFileHandler.py#L23

PR #20, opened 2017-01-24
Enable double pin numbering

Frankkkkk:
Useful when creating pin headers ;-)

I just saw that kicad_mod is deprecated, so …


pointhi:
thanks. But I don't think there is no much usefulness of this feature in a deprecated framework :)

PR #21, opened 2017-02-13
Added klc.py

SchrodingersGat:
Contains KLC constants for various dimensions
Scripts can reference these such that they do not need to be updated in multiple scripts


pointhi:
This feature is already implemented, at least for lines:
https://github.com/pointhi/kicad-footprint-generator/blob/master/KicadModTree/KicadFileHandler.py#L23

SchrodingersGat:
Oh, fantastic :)

PR #22, opened 2017-02-13
Added various scripts for Molex connectors

SchrodingersGat:
This PR adds scripts for generating various Molex connectors

Most of these have been accepted into https://github.com/KiCad/Connectors_Molex.pretty


pointhi:
Thanks for those scripts

PR #25, opened 2017-02-13
Added scripts for various inductors

SchrodingersGat:
This PR adds various inductor scripts


pointhi:
Are those footprints already merged in the library?

SchrodingersGat:
Yup

PR #26, opened 2017-02-14
New scripts for generating jst ph series connectors conforming to klcv1.2

poeschlr:
Reimplementd scripts using kicad mod tree.
Footprints created by this script conform to klcv1.2 (there is an option to create footprints conforming to klcv1.1)


pointhi:
thanks for upgrading and improving my scripts :)

PR #28, opened 2017-02-23
Better phoenix scripts (for new klc)

poeschlr:
The new phoenix contact connector scripts for the current footprints.

Merged footprints from this new scripts:
https://github.com/KiCad/Connectors_Phoenix.pretty/pull/44
https://github.com/KiCad/Connectors_Phoenix.pretty/pull/45


pointhi:
thanks for this improvements.

As sidenote: you don't have to specify the width of lines, because the script automatically selects parameters valid to KLC (see #21).

poeschlr:
The width is specified because these scripts are highly configurable and can generate scripts for any standard. (Including my own with slightly different line widths. I made my standard back when kicad still was at klc version 1.0)

pointhi:
That's a valid argument.

I wan't to ask you, what do you think about my approach to unifying/simplifying the definitions of models.
Current you define them inside the python scripts, I selected them by arguments and bashscripts and I think there are other approaches too.

As proposal, I developed a first version of ModArgparse, which handles all those reading tasks for you and more (importing definitions from .csv and .yml files, type conversation, specifying of default values, required and optional arguments, generation of example definition files,...)
* https://github.com/pointhi/kicad-footprint-generator/blob/master/KicadModTree/ModArgparser.py
* https://github.com/pointhi/kicad-footprint-generator/blob/master/KicadModTree/examples/argparse_example.py

What do you think about it? Would you use it, or does this concept lack at some points

poeschlr:
Looks good to me.
I would use it (for new designs).
This could make the creation of footprints a lot easier to understand for outsiders.

As an example (if i understand your effort correctly) for my phoenix footprints i would use maybe one csv/yml file for series specific parameters and one file per standard. (the stuff that is currently done in the options manager)

I'm sure i will understand it better once i used it. (or if someone else made some footprints with it.)

pointhi:
well, in the current implementation you would have one file for the footprints, and the standard would likely be defined as default parameters and/or by some self made functionality.

A good approach would probably adding a functionality to allow a global overwrite of parameters which change line width, courtyard distance,... of all related footprints. It would be useful to specify how such a overwrite would look like and what variables it contain, so everyone can use their overwrite on all scripts.

I also made some initial documentation, where we can put such stuff in the future: http://kicad-footprint-generator.readthedocs.io/en/latest/

PR #31, opened 2017-04-12
Update molex kk 6140 script

hackscribble:
Converted script to use KicadModTree.


pointhi:
thanks

pointhi:
@hackscribble: please recheck this script. It looks like F.Fab is outside the silkscreen layer

molex_ffab

To be accurate: F.Fab should represent actual part dimensions, and the silkscreen should be visible even when the part is mounted on the pcb. Look into the KLC for a more details about that.

hackscribble:
Thanks @pointhi I will check which one is actual component size and swap them round.

hackscribble:
It was the outline that was the wrong size.  I have moved it to the correct position, and will submit a new PR with the updated script.

Question: the original author of the script included detail on the silkscreen representing the feet of the part and the locking ramp.  Because the part is raised on feet, these details are partially visible when the component is installed.  Would you advise on removing the detail inside the silkscreen outline and replacing it with a pin 1 marker outside the outline?

image

pointhi:
thanks. I would keep the silkscreen as it is, to simplify hand placement.

A pin-1 marker wouldn't be much help in this case, because it's not really clear by looking onto the part what is pin-1. Nevertheless, a pin 1 marker would be beneficial for testing and should be additionally added.

hackscribble:
OK, will do.

hackscribble:
Vertical bar added to silkscreen as pin 1 marker.  I will put through a PR with the updates.

image

PR #42, opened 2017-06-07
added ${KISYS3DMOD}/ is it the right place to add it?

Shackmeister:


poeschlr:
You might want to check if the filename given already has ${KISYS3DMOD}/ in it.
(Otherwise you end up with ${KISYS3DMOD}/${KISYS3DMOD}/)

And i would not add it by default. What if someone wants to use this generator scripts to generate their own footprints. (I'm using the scripts for exactly this.) Such a user might need a different prefix. It needs to be configurable

I think the handling of this should go into the node not the file handler.
In this case i would give the node a member variable called prefix which is initialized with ${KISYS3DMOD}/ if the node is not created with a prefix.
Something like the offset in the rectline node. (Maybe there is a some error handling necessary. But simplified it could look like this.)

```
if kwargs.get('prefix'):
  self.prefix = kwargs['prefix']
else:
  self.prefix = "${KISYS3DMOD}/"
```

Of course the serialize function still needs to use this prefix.
(And the render function in the node should also use this prefix.)

Shackmeister:
makes sense, Ill close this

PR #43, opened 2017-06-22
MECF Socket and Footprint

matyro:
This is the script to create Samtec MECF Footprint for sockets and pcb edge connector used in the https://github.com/KiCad/Connectors_Samtec.pretty/pull/5


pointhi:
thank you for your contribution

PR #66, opened 2017-10-09
Molex 52991: Fix h_body_length calculation

evanshultz:
Currently half the body width is calculated as `h_body_length = (pad_x_span / 2.0) + 2.15 + 0.35`. I believe this is wrong.

Looking at the datasheet image, we see:
image

0.35mm is between the last pin and the inner locking/locating feature, but between the edge of the body and the last pin is is only 2.15mm. The 0.35mm should not be included.

This is also reflected in the table of dimensions, where C is 4.3mm (2.15mm * 2) larger than A:
image

Am I mis-reading this or is the calculation in error? If I'm right, the footprints are 0.7mm wider than part and should be re-generated.


pointhi:
can you give me a reference to a PR where this change was commited?

evanshultz:
@pointhi 
Sorry, I'm not sure what you're asking for.

Documents and other relevant information:
- Footprints were added to the repo by https://github.com/KiCad/Connectors_Molex.pretty/pull/38.
- The script is at https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/Connectors_Molex/slimstack-52991.py with the line I mentioned above being line 53.
- Datasheet is at http://www.molex.com/pdm_docs/sd/529910208_sd.pdf.

The comment above describes the calculation I think may be wrong. I believe `h_body_length = (pad_x_span / 2.0) + 2.15 + 0.35` should be `h_body_length = (pad_x_span / 2.0) + 2.15`, removing the ` + 0.35`.

evanshultz:
The script was added to this repo by https://github.com/pointhi/kicad-footprint-generator/pull/35.

I would like someone else to verify the calculation is wrong before doing anything about it. @hackscribble This was your script. Could you take a look as well?

poeschlr:
I will include these changes in my larger rework of all molex generator scripts.
(By the way a lot of the molex scripts had very wrong calculations. Not only this series. So it might take some time until everything is finished.)

Thanks @evanshultz. This will save me some time searching.

See commit: https://github.com/pointhi/kicad-footprint-generator/pull/81/commits/64988e5d6f1d12cfa3a7ce6595fabdcdd2b9f8ed

evanshultz:
Fixed by @poeschlr

PR #73, opened 2017-10-31
Add 501331, 502426, 502430 scripts

evanshultz:
These were scripts used for merged PRs at https://github.com/KiCad/Connectors_Molex.pretty/pull/59 and https://github.com/KiCad/Connectors_Molex.pretty/pull/55. Argaparser interface is used, but simplified over the Molex scripts I started with since the pincount is derived from the first two digits of the PN suffix.


pointhi:
thanks

PR #74, opened 2017-11-01
Fix 502426 script

evanshultz:
As Oliver noticed, my pin 1 marks were in the wrong place. The PR to fix footprints is at https://github.com/KiCad/Connectors_Molex.pretty/pull/60, and this updates the script.


pointhi:
thanks for your fix

PR #75, opened 2017-11-05
New script for generating two terminal smd components

poeschlr:
This script generates the footprints for this pull request: https://github.com/KiCad/kicad-footprints/pull/3

For some reason i did not see the generic generator when i started with it. I did keep it though because it might still serve some purpose.

My script makes heavy use of yaml files. 

 - one yaml file for configuration information, 
 - one file that holds ipc specifications, 
 - one file that determines what footprints will be created and 
 - many files that hold footprint measurement information.

This should make it easy to generate slightly different footprints (example if the KLC does change).
It should also be quite easy to add more footprints by either extending the size definition files or by adding a new group to the command file.

Do not merge before the footprints have been merged. (I need a pull request though because i want to reference it in an issue.)


poeschlr:
@pointhi: The footprints generated by these scripts have been merged.

PR #77, opened 2017-11-06
Fix format string in _serializer_arc (angle currently has a lot of precision.)

poeschlr:
While regenerating my phoenix connectors i noticed that the arc angle was created with lots of precision.

The reason might simply be the missing '.' in the format string.

If it is intended to print with this much precision, then you can ignore this PR.


pointhi:
good catch. But the current solution would cause loss of precision if someone want's to specify the angle with an accuracy more than 0.1 degree.

That could be the case when the arc is used for Edge.Cuts,  where points need to match very accuratly on top of each other.

pointhi:
command back, it seems kicad limits to one trailing digit.

It would make sense to match KiCad as close as possible. But I suspect allowing more precision would cause problems. @poeschlr what do you think?

poeschlr:
I really don't know the kicad side that well. Maybe we should ask about it on the mailing list.
I just noticed that the current implementation of your script creates files that differ from older implementations and `{1f}` looked suspicious to me.

pointhi:
I fixed it in the way to not output useless decimal, which should be better than the current behaviour.

PR #79, opened 2017-11-10
Add TE 8495x script

evanshultz:
This is the script for footprints merged by https://github.com/KiCad/Connectors_TE-Connectivity.pretty/pull/7.

This script has a different argument parser interface than I've seen used. There are two parameters, and either two or zero arguments must be provided. I would have liked to either accept no input or use a default list of all elements to be generated. I tried all permutations of `nargs=`, and `default=`, and `type=`, and `action=`, but I was unable to find a better way than this. I can either get a single item or a list, and with two parameters I could find a better way to combine them. Here is the relevant code:
```
if __name__ == '__main__':

    # handle arguments
    parser = argparse.ArgumentParser()
    parser.add_argument('pincount', help='number of pins of the connector', type=int, action='append', nargs='?')
    parser.add_argument('partnumber', help='suffix of 502426 series number (e.g. 0810)', type=str, action='append', nargs='?')
    parser.add_argument('-v', '--verbose', help='show extra information while generating the footprint', action='store_true')
    parser.parse_args()
    args = parser.parse_args()
    
    # determine if the user entered valid arguments
    if ((args.pincount == [None]) and (args.partnumber == [None])):
        # no arguments provided, so generate all footprints
        pincounts = range(4, 31)
        partnumbers = ["84952","84953"]
    elif ((args.pincount == [None]) or (args.partnumber == [None])):
        # only one argument, so bail
        sys.exit("Error: enter zero or two arguments")
    else:
        # both arguments provided, so generate on that one footprint
        pincounts = args.pincount
        partnumbers = args.partnumber
    
    # with pincount(s) and partnumber(s) to be generated, build them all in a nested loop
    for partnumber in partnumbers:
    
        for pincount in pincounts:
```

### DIFF
--- a/scripts/Connectors_TE-Connectivity/84952-84953.py
+++ b/scripts/Connectors_TE-Connectivity/84952-84953.py
@@ -48,7 +48,6 @@ if __name__ == '__main__':
     parser.add_argument('pincount', help='number of pins of the connector', type=int, action='append', nargs='?')
     parser.add_argument('partnumber', help='suffix of 502426 series number (e.g. 0810)', type=str, action='append', nargs='?')
     parser.add_argument('-v', '--verbose', help='show extra information while generating the footprint', action='store_true')
-    parser.parse_args()
     args = parser.parse_args()
     
     # determine if the user entered valid arguments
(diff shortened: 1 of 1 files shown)


pointhi:
Thank you for your contribution

PR #80, opened 2017-11-10
Fix drawing_tools.py to run on Windows

evanshultz:
When trying to build the pot footprints, I got the following error (the second run is after this change):
image

I found that the degree symbol was causing the issue, I assume only on my machine. I have:
- Windows 7 x64
- Python 2.7.13 (included with KiCad install)

Anyway, this avoids any other changes to the file and does not require defining formatting which could possibly break other things. So it seems the minimal fix. I also corrected a typo.


terjeio:
You can also add
```
# -*- coding: utf8 -*-
```
as the first line in the file to avoid the issue

evanshultz:
As stated above, I chose this method because it specifically does not require defining the encoding.

terjeio:
@evanshultz Sorry about my comment, I have to read comments more closely - especially when late.

poeschlr:
I don't think most of the scripts are compatible to python 2.x anyway (difference in division behavior, file handler behaves differently, ...)
It would be a lot of work to ensure that all scripts run smoothly under all versions of python. Why not just state the footprint generator is only guaranteed to work under 3.x (maybe even 3.6+)

If we want the scripts to run smoothly under 2.x we at least need to ensure division is imported from future. (And no files can include non ascii chars as you already discovered)

Let me know if we want 2.x compatible scripts. (Then i need to rewrite most of my scripts. Again)

Btw: python 3 is now nearly 10 years old. (Released 2008) And the last 2.x release came out mid 2010.

evanshultz:
KiCad ships with Python 2.7, which was my reason to make the few changes needed to this file. I didn't feel this small change to one file that doesn't break anything else (that I could find) would cause any issues.

I'm not advocating or enforcing one way or the other, but Windows users won't have Python installed by default and thus may rely on KiCad's installed version. Personally I'd prefer Python 3 as well.

pointhi:
the change is not problematic, so I will merge it as is.

@poeschlr The generator scripts can be python2 or python3, I don't really enforce rules for them. The Framework itself should be stay to be python2&3 compatible as long as possible. (At the moment Python 2.7 and Python 3.3-3.6 are tested in travis). Sadly, Python 3 support is still not the standard for all libraries and runtimes, so I think supporting the old version is still a good idea.

pointhi:
@poeschlr When your scripts are Python3 only, I would suggest to change the header to: ```#!/usr/bin/env python3``` for your scripts to enforce it on linux at least, and help windows users identifying the problem :)

PR #82, opened 2017-11-14
New script for socket strips

terjeio:
Single row SMD left/right sockets for 2.54 and 2.00 mm added. Updated
for KLC3 and new naming convention.

NOTE: most dimensions are from legacy footprints since there are many
variants available. There are references to dimensions and sources that
I have found that could be useful for adding more variants.

Also included are two wrapper classes in a separate module, this for
some of the core primitives. It allows simpler coding in a somewhat
similar style as CadQuery provides for 3D modeling.

A pull request will soon be made for the footprints, I will refer to that when it is done. Also, I have scripted 3D models at hand that I will make a PR for when the footprints are approved.

I have not updated the existing scripts as code is shared with pin headers and I did not want to affect those. The main reason is that the generator code is nearly 100% new and based on my wrapper classes.

### DIFF
--- a/scripts/Conn_PinSocket/canvas.py
+++ b/scripts/Conn_PinSocket/canvas.py
@@ -28,14 +28,29 @@
 #
 # (C) 2017 by Terje Io, <http://github.com/terjeio>
 
+#
+# NOTE:
+# The code for the the Keepout class is based on code from drawing_tools.py
+# Currently circular and (rounded) rectangular keepout zones are respected
+# for horizontal and vertical lines, only basic support (bounding box only) for diagonal lines.
+# Arcs and circles are not yet handled. On TODO: list.
+# More testing needs to be done for edge cases, such as for lines inside
+# the bounding box but extending outside the keepout area.
+# Preferably the private _Keepout class should encapsulate all the line processing?
+#
+
 import sys
 import os
 import math
 
 sys.path.append(os.path.dirname(os.path.realpath(__file__)) + "\\..\\..")
 
-from KicadModTree.nodes.base import Line, Circle, Text, Pad
+from collections import namedtuple
+
+from KicadModTree import Point
+from KicadModTree.nodes.base import Line, Arc, Circle, Text, Pad
 from KicadModTree.nodes.specialized import RectFill
+from KicadModTree.util.kicad_util import formatFloat
 
 class Layer:
 
@@ -57,7 +72,9 @@ class Layer:
     def __init__(self, footprint, layer='F.Fab', origin=(0,0), line_width=None, offset=None):
         self.footprint = footprint
         self.layer = layer
+        self.polyline_start = None
         self.gridspacing = None
+        self.keepout = None
         self.setOrigin(origin[0], origin[1])
         self.txt_size = [1.0] * 2
         self.setTextDefaults()
@@ -72,6 +89,8 @@ class Layer:
         self.setLineWidth(self._DEFAULT_LINE_WIDTHS.get(layer, self._DEFAULT_LINE_WIDTH) if line_width == None else line_width)
         if layer.split('.')[1] == 'CrtYd':
             self.setGridSpacing(self._DEFAULT_GRID_CRT)
+#        else:
+#            self.setGridSpacing(0.001)
 
     @staticmethod        
     def getBevel(width, height):
@@ -90,6 +109,9 @@ class Layer:
         if self.auto_offset:
             self.offset = self.line_width / 2.0
         return self
+
+    def getMinPadDistance(self):
+        return self._DEFAULT_MIN_PAD_DISTANCE
 
     def setTextDefaults(self, max_size=None, min_size=None):
         if max_size == None and min_size == None:
@@ -131,6 +153,14 @@ class Layer:
         self.y = self._align(y)
         return self
 
+    def gotoX(self, x):
+        self.x = self._align(x)
+        return self
+
+    def gotoY(self, y):
+        self.y = self._align(y)
+        return self
+
     def jump(self, x, y):
         self.x += self._align(x)
         self.y += self._align(y)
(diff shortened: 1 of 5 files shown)


terjeio:
PR for footprints: [https://github.com/KiCad/kicad-footprints/pull/41](https://github.com/KiCad/kicad-footprints/pull/41)

jkriege2:
Just a note: In https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/tools/drawing_tools.py there are functions for "keepouts", which are a list of rectangles that may not contain lines. These tools allow to set a list of these keepouts that cover all pads. Then you can use the mre "natural" drawig functions and what they draw will mind these keepouts. This allows to lgically specify what you want to draw and have the scripts handle the intricacies, like silkscreen not crossing pads and their mask-clearance ...

Would that help?

terjeio:
@jkriege2 I have seen the code for keepout handling and my plan is to wrap it in a class and make it even more simple to use. I have to investigate a bit more but I think it should be possible to build (parts of?) the keepout areas directly from the pads list with a single method call and then enhance the layer class I've made.

The workflow I am thinking of is something like this: create layer canvas > add pads > add keepout instance and bind to layer > define keepouts (eg. from pads) > draw using the normal canvas primitives which then observe the keepouts. But then, maybe I am a bit crazy...

It is when I see code like this I start thinking:
```
keepout=keepout+addKeepoutRect((x - 1) * rm,0,pad[0]+2*min_pad_distance+2*lw_slk, pad[1]+2*min_pad_distance+2*lw_slk)
```
I would rather like it to be somewhat like this:
```
silk = Layer(...) # create silk layer instance
pads = Pads(...) # create pads layer
pads.add(...) # add pads
....
keepout = Keepout(silk) # create keepout instance and bind to silk layer
keepout.addPads(pads) # add pads to keepout
...
```
then use the drawing primitives via the Layer class transparently. Well, maybe not exactly like this but I think it should be possible to get rid of more code/math in the footprint scripts and thus make them both more readable and more easy to code.

So, for now I will go ahead without using the drawing tools and just stick with your excellent primitives.

If this is a crazy idea just tell me, I am sure I have not grasped alle the complexities related to this field yet.
Anyway, I see this as a nice challenge as part of learning to code in Python which is a new language for me.

BTW: I see that someone wants you to separate the core code from the footprint scripts, I second that because that was somewhat confusing me when I started looking into this.

Terje

terjeio:
Last PR update includes new code for simple to program keepout processing for vertical and horizontal lines:

Test with 200 lines vertically, 150 horizontally:

bilde

Keepout areas:

bilde

Areas are based on a rounded rectangles and their bounding box.

Code (not included in PR):

```python
Keepout.DEBUG = 1

footprint_name = "test"
lib_name = "keepout"

print "###################"
print footprint_name, "in", lib_name

# init kicad footprint
kicad_mod = Footprint(footprint_name)
kicad_mod.setDescription("keepout test")
kicad_mod.setTags("keepout test")

kicad_modg = Translation(0.0, 0.0)
kicad_mod.append(kicad_modg)

# create layer canvases
silk    = Layer(kicad_modg, 'F.SilkS')
fab     = Layer(kicad_modg, 'F.Fab')
crt     = Layer(kicad_modg, 'F.CrtYd')
pads    = PadLayer(kicad_modg, [1.0, 1.0], Pad.TYPE_THT, Pad.SHAPE_OVAL, shape_first=Pad.SHAPE_RECT, drill=0.5)
keepout = Keepout(silk)

height = 4.0
width  = 6.0

pads.add(0.0, 0.0)\
    .add(0.0, 2.0)\
    .add(2.0, 0.0, size=[1.2, 0.7])\
    .add(2.0, 2.0, shape=Pad.SHAPE_RECT, size=[1.2, 0.7])\
    .add(4.0, 0.0, size=[0.7, 1.2])\
    .add(4.0, 2.0, shape=Pad.SHAPE_RECT, size=[0.7, 1.2])

silk.setOrigin(-1.0 - silk.offset, -1.0 - silk.offset)\
    .rect(width + silk.offset * 2.0, height + silk.offset * 2.0, origin="topLeft")\
    .goHome()

fab.setOrigin(-1.0, -1.0)\
   .rect(width, height, origin="topLeft")

crt.setOrigin(-1.0 - crt.offset, -1.0 - crt.offset)\
   .rect(width + crt.offset * 2.0, height + crt.offset * 2.0, origin="topLeft")

# add pads to silk keepout
keepout.addPads()
keepout.debug_draw()

silk.goHome()

test = 1 + 2

if test & 1:
    for y in Keepout.frange(0.0, height, height / 150):
        silk.gotoY(-1 + y)\
            .right(width)\
            .left(width, draw=False)
            
if test & 2:
    silk.goHome()
    for x in Keepout.frange(0.0, width, width / 200):
        silk.gotoX(-1 + x)\
            .down(height)\
            .up(height, draw=False)

if test & 4:
    silk.goHome()
    for x in Keepout.frange(width / 2.0, width + 2, width / 10):
        silk.gotoX(x - 2.5)\
            .to(3, height)\
            .gotoX(6.5 - x)\
            .to(-3, -height)

if test & 8:
    silk.goHome()
    for x in Keepout.frange(width / 2.0, width + 2, width / 10):
        silk.gotoX(x - 2.5)\
            .to(-3, height)\
            .gotoX(6.5 - x)\
            .to(3, -height)   
                     
# write file
file_handler = KicadFileHandler(kicad_mod)
file_handler.writeFile(root_dir.saveTo(lib_name) + footprint_name + '.kicad_mod')

```

terjeio:
Note: New yaml parameter file is shared with 3D model generator script to avoid dimensional mismatches and to make it easy to keep both in sync. 3D script and models will be PR'ed soon.
Description is updated with datasheet references where relevant, parts where dimensions are from legacy footprints are referenced to KiCad 4.0.7. Some updates for better compliance with datasheets.

pointhi:
I will merge as is.

@terjeio greate work for implementing a keepout. It would be cool to implement it in the main framework in the future, if you are interested. It could get handy for other footprints like capacitors,... too.

terjeio:
@pointhi 
Feel free to implement it in the main framework as you please. Currently the classes I have made stands on top of your base primitives and "overrides" some of the specialized ones. They can already be used "out of the box" by importing canvas.py.

I plan to add complete angled line support and support for arcs/circles in the keepout processor when I get time. Also, I want to add inset/outset for polylines too - I think this will be on top of my TODO list.

PR #84, opened 2017-11-27
Update hirose scripts

poeschlr:
This contains the updates for the hirose scripts.
The generators for df33 straight and df63 angled have no footprints in the current lib. (I can't even find DF33 on the webside.)
I did not touch these for this reason.


poeschlr:
@pointhi wait with merging this i will add some more hirose connectors. In addition i will use this PR to do a bit more cleanup of my scripts.

poeschlr:
@pointhi i am done here. (This can now be merged.)
Changes: I moved all my connector generators in a new sub folder Connector and i unified the naming of scripts. (Reason: they share some assets that i now have at a more sensible location.)
I also made better new scripts for hirose connectors. This needed the addition of a function to add a hatched rectangle on dwgs.user for the denotation of keepout zones. (I plan to generalize it such that it also works for any polygon. But this will happen later)

PR #86, opened 2017-11-29
Naming convention tantal caps

poeschlr:
Move letter code to back to avoid impression that these are manufacturer specific.

This can be merged as soon as https://github.com/KiCad/kicad-footprints/pull/136 is accepted.

Edit: I also change the smd 2 terminal generator such that it now uses the global KLC config file. (If the naming convention change is not accepted i will revert that part of this PR.)


poeschlr:
The footprints have been merged. -> this is ready to merge

PR #87, opened 2017-11-29
Samtec MECF pcb side

poeschlr:
kicad-footprint pull request: https://github.com/KiCad/kicad-footprints/pull/137

I will only make minimal updates such that this uses the config files.


poeschlr:
The footprints have been merged. -> this is ready to merge now.

PR #91, opened 2017-12-01
New scripts for jae ly20

poeschlr:
footprint pull request: https://github.com/KiCad/kicad-footprints/pull/147


poeschlr:
The footprints have been merged -> ready to be merged

pointhi:
I see you always add the same helper function. Probably we should add support for that into the core framwork. Could it be added to the Point class?

poeschlr:
In my symbol [generator i added support for a grid to the point class](https://github.com/KiCad/kicad-library-utils/tree/master/schlib/autogen/KiCadSymbolGenerator). This seems to work so far.

PR #92, opened 2017-12-03
Make manage.sh use greadlink on OS X

twam:
readlink on OS X is different (https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/readlink.1.html) than on Linux. Use the GNU's readlink available through homebrew (https://brew.sh/index_de.html) instead.


pointhi:
thanks to add osx support to the script

PR #83, opened 2017-11-24
reworked THT diodes

jkriege2:
2017-11-25 12_28_41-footprint editor no active library
2017-11-25 12_28_23-footprint editor no active library
2017-11-25 12_49_49-footprint editor no active library
2017-11-25 12_47_28-footprint editor no active library
2017-11-25 12_47_18-footprint editor no active library
2017-11-25 12_47_08-footprint editor no active library
2017-11-25 12_46_52-footprint editor no active library
2017-11-25 12_54_35-footprint editor no active library
2017-11-25 12_56_21-3d viewer
2017-11-25 12_56_06-3d viewer
2017-11-25 12_55_44-3d viewer
2017-11-25 12_54_55-footprint editor no active library
2017-11-25 12_54_35-footprint editor no active library

PR #93, opened 2017-12-04
Fix replacement of fab text on small THT Resistor packages

grob6000:
(1) Changed the directory name from "Resistors_THT" to "Resistor_THT" (to match footprint and 3d library new convention = no plural)
(2) Adjusted placement of fab ref text to avoid clash (see https://github.com/KiCad/kicad-footprints/issues/74)

A few examples of the improved layout for previously troubled parts here:
Image of footprints


grob6000:
Just noticed https://github.com/KiCad/kicad-footprints/issues/76 - will need to have a look at 3d references too. Hang on...

grob6000:
Also have now updated the 3d references to work. Checked with latest 3d library (see https://github.com/KiCad/kicad-packages3D/pull/180).

grob6000:
Some additional problems have presented when looking at the diff with the footprint library models. Investigate and return...

grob6000:
Made a few changes - the description is now better aligned with the previous, but the naming is correct and model associations work. Got a bit bogged down in the description field. Footprint update PR is at https://github.com/KiCad/kicad-footprints/pull/151. Hope this is useful.

poeschlr:
@jkriege2 could you take a look at this?

I assumed that most of this should have been fixed by https://github.com/pointhi/kicad-footprint-generator/pull/83 (After all the diode and resistor tht generators use the same functions in the tools directory.)

Shackmeister:
@Pointhi the generated footprints have been merged a long time ago, can you have a look if it's merge worthy? Im gonna be working on the same file for inductors and this might fix parts of it

pointhi:
ok, thanks

PR #94, opened 2017-12-04
Add option to shift center point to match datasheet

robertlong13:
I only checked for the picoblade series, but must be done for all.
Also fixed a bug causing "JST" to be listed in description of all files.


robertlong13:
To clarify, this PR is to bring the connectors into compliance with KLC F6.2

pointhi:
thanks for your contribution

PR #97, opened 2017-12-07
Fix permissions on .py files in Connector/*

twam:


pointhi:
thanks for fixing this

PR #104, opened 2018-01-03
Add SOP-18.

twam:
Adds SOP-18 footprint generator.
Footprint has been merged into kicad-footprints in https://github.com/KiCad/kicad-footprints/pull/215.


pointhi:
Thanks

PR #109, opened 2018-01-21
Add support for solder mask margin

tvannoy:
I added support for defining solder mask margin/clearance.


pointhi:
thanks, looks good to me

PR #112, opened 2018-02-10
Add Stackpole CSSx resistors script

strombom:
Stackpole CSS and CSSH
https://www.seielect.com/catalog/sei-css_cssh.pdf

Current sense resistors. 

image


strombom:
Added fabrication refdes, centered component

image

image

image

poeschlr:
The footprint side was taken over by @Misca1234 and does not use a script for it. I therefore assume this can be closed.

PR #115, opened 2018-02-23
added 3D-path-prefix to resistor scripts

jkriege2:
This simply adds the prefix ${KISYS3DMOD}/ to the 3D model paths

made while fixing this issue: https://github.com/KiCad/kicad-footprints/issues/393


pointhi:
@jkriege2: I was not able to find  a PR where this changed any footprints (probably someone did the fix without changing the script already). Thanks

PR #119, opened 2018-04-04
I had slightly wrong formulas and the wrong ipc tables for some parts

poeschlr:
Issue on the footprint libs: https://github.com/KiCad/kicad-footprints/issues/466

Footprints have not yet been merged. (Will do it tomorrow as i am a bit too tired right now. This is such that the reporters can review my fix)

### DIFF
--- a/scripts/SMD_chip_package_rlc-etc/SMD_chip_package_rlc-etc.py
+++ b/scripts/SMD_chip_package_rlc-etc/SMD_chip_package_rlc-etc.py
@@ -62,13 +62,15 @@ class TwoTerminalSMDchip():
         if 'terminator_spacing_max' in device_params:
             Tmin = (device_params['body_length_max'] - device_params['terminator_spacing_max'])/2
             Tmax = (device_params['body_length_min'] - device_params['terminator_spacing_min'])/2
-            Ttol = (Tmax-Tmin)/2
+            Ttol = (Tmax-Tmin)
             Stol_RMS = math.sqrt(Ltol**2+2*(Ttol**2))
             Smax_RMS = device_params['terminator_spacing_min'] + Stol_RMS
             Gmin = Smax_RMS - 2*ipc_data['heel'] - math.sqrt(Stol_RMS**2 + F**2 + P**2)
+            if device_params['body_length_min'] == 6.15:
+                print(Gmin)
 
         else:
-            Ttol = (device_params['terminal_length_max'] - device_params['terminal_length_min'])/2
+            Ttol = (device_params['terminal_length_max'] - device_params['terminal_length_min'])
             Stol_RMS = math.sqrt(Ltol**2+2*(Ttol**2))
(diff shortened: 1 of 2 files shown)


scripts/SMD_chip_package_rlc-etc/ipc_smd_two_terminal_chip.yaml
```diff
@@ -63,24 +63,24 @@ ipc_spec_smaller_0402:
...
+# Side (JS) |  -0.05  |  0.00     |  0.05
 # Courtyard |  0.1    |  0.25     |  0.5
 
 ipc_spec_larger_or_equal_0603:
     least:
         toe: 0.15
-        heel: -0.3
-        side: 0.01
+        heel: -0.05
+        -side: 0.05
```
sethhillbrand: I think this '-' is a typo

scripts/SMD_chip_package_rlc-etc/SMD_chip_package_rlc-etc.py
```diff
@@ -60,17 +60,22 @@ def calcPadDetails(self, device_params, ipc_data, ipc_round_base, footprint_grou
...
+            Tmax = (device_params['body_length_min'] - device_params['terminator_spacing_min'])/2
+            Ttol = (Tmax-Tmin)/2
+            Stol_RMS = math.sqrt(Ltol**2+2*(Ttol**2))
+            Smax_RMS = device_params['terminator_spacing_min'] + Stol_RMS
+            Gmin = Smax_RMS - 2*ipc_data['heel'] - math.sqrt(Stol_RMS**2 + F**2 + P**2)
+
         else:
-            terminal_tolerance = device_params['terminal_length_max'] - device_params['terminal_length_min']
-            spacing_tolerance = math.sqrt(length_tolerance**2+terminal_tolerance**2)
+            Ttol = (device_params['terminal_length_max'] - device_params['terminal_length_min'])/2
```
sethhillbrand: I suspect that Ttol should not be divided by 2 here. (or above on line 65)  It doesn't look like you are measuring from the center of the component.  These are absolute lengths

poeschlr: Yes it needs to be devided by two. The reason is that the tolerance is relative to the nominal size which is assumed to be between max and min.

poeschlr: Ok i re read the relevant parts of the standard. I think you might be right.

poeschlr:
I fixed all the mentioned errors.

The footprints are now added in PR https://github.com/KiCad/kicad-footprints/pull/469

poeschlr:
@pointhi the new footprints have been merged -> this is now also ready to be merged.

pointhi:
thanks, I will merge

PR #120, opened 2018-04-05
added TO-220F footprints and fixed y-offset of pins of of TO-220 

jkriege2:
matches: https://github.com/KiCad/kicad-footprints/pull/470

1. fixed TO-220 footprints: the y-offset of pins to metal-back was wrong, it has been modified to match https://www.vishay.com/docs/66542/to-220-1.pdf
2. added footpritns for TO-220F (isolated plastic package), see https://www.fairchildsemi.com/package-drawings/TO/TO220M03.pdf

2018-04-05 11_23_18-pcbnew d__kicad_devproject_devproject kicad_pcb
2018-04-05 11_40_24-3d viewer


pointhi:
@jkriege2 can I merge?

jkriege2:
yes, I think so ... the footprints have been merged ... only the 3D models are missing ;-)

jkriege2:
@Shackmeister could you have a look at this? The footprints have been merged, but the 3D models are still missing (and I feel they would be a good addition).

PR #123, opened 2018-04-22
Add pad shape offset support

agren:
Makes KicadFileHandler aware of pad offsets.

Since Kicad does not support specifying an offset for circular pads.
Circular pads will be converted to oval pads if an offset is specified.


KicadModTree/KicadFileHandler.py
```diff
@@ -225,7 +225,11 @@ def _serialize_Model(self, node):
         return sexpr
 
     def _serialize_Pad(self, node):
-        sexpr = ['pad', node.number, node.type, node.shape]
+        sexpr = ['pad', node.number, node.type]
+        if node.shape == Pad.SHAPE_CIRCLE and (node.offset.x != 0 or node.offset.y != 0):
```
pointhi: I don't think it is a good idea to implicitly convert the type

PR #124, opened 2018-04-23
Rename DFN/QFN generator, add new footprints

Ratfink:
The DFN/QFN library has been renamed to Package_DFN_QFN.pretty; this renames the generator script to match.

This also adds two new footprints: QFN-36-1EP_6x6mm_P0.5mm_EP4.1x4.1mm and QFN-48-1EP_7x7mm_P0.5mm_EP5.6x5.6mm, needed for STM32 chips.

External PRs:
- https://github.com/KiCad/kicad-footprints/pull/504
- https://github.com/KiCad/kicad-footprints/pull/505

PR #125, opened 2018-04-24
A generator for qfp packages (using ipc standards to derive the landing pattern)

poeschlr:
Work in progress. Will notify when done and the generated footprints are merged.


poeschlr:
@pointhi the footprints that are produced by this have been merged. -> this is ready to merge.

The error reported by codeclimate is not new.

poeschlr:
I merged #124 into this as i and @Ratfink both touched the same files. (I because i reorganized the generators and @Ratfink because he added some new footprints and updated the lib name from Housings -> Package)

poeschlr:
The dfn/qfn generator needs a bit more work to ensure that there always is some minimum clearance between the exposed pad and the outer pads.

poeschlr:
This can be merged for now, as better qfn footprints would require support for custom pads which i will implement in a separate PR.

KicadModTree/nodes/base/Pad.py
```diff
@@ -82,6 +82,7 @@ def __init__(self, **kwargs):
         self._initOffset(**kwargs)
         self._initDrill(**kwargs)  # requires pad type and offset
         self._initSolderPasteMargin(**kwargs)
+        self._initSolderPasteMarginRatio(**kwargs)
```
pointhi: please add docstring as well

poeschlr: Should be done

pointhi:
current status? And there are now merge conflicts

poeschlr:
I will make the qfn generator better. (Make a specialized node for exposed pads and for chamfered pads, use these nodes in the generator)

poeschlr:
@pointhi it seems the python version used by travis does not come with the futures package installed. The problem is without it the round function behaves differently in python 2 as compared with python 3. Should i just accept that difference (remove the test case plus the import) or is there a way to get futures package into the travis python version?

KicadModTree/nodes/specialized/ExposedPad.py
```diff
@@ -0,0 +1,254 @@
...
+                pads.append(Pad(
+                    number=self.number, at=self.at, size=self.bottom_size,
+                    shape=Pad.SHAPE_RECT, type=Pad.TYPE_SMT,
+                    layers=self.bottom_pad_Layers
+                ))
+
+        return pads
+
+    @staticmethod
+    def __roundToBase(value, base):
```
pointhi: Duplication of functionality? https://github.com/pointhi/kicad-footprint-generator/blob/master/KicadModTree/Point.py#L64

poeschlr: Your function only works for points, not for floats and other values like tubles and arrays.

pointhi: vector is a good idea. Such basic things like round_to should be moved to a utils class where they belong to

poeschlr: I changed the name from Point -> Vector and used this new class for all size and grid specifiers in the exposed pad class. This allowed me to remove the roundToBase member from it.

poeschlr:
I will add some further functionalists to the exposed pad class. Specifically to the automatic paste management for avoiding thermal vias.

KicadModTree/Point.py
```diff
@@ -16,282 +16,32 @@
...
-                        'y': self.y / other.y})
-
-    def __truediv__(self, obj):
-        return self.__div__(obj)
-
-    def __dict__(self):
-        return {'x': self.x, 'y': self.y}
-
-    def render(self, formatcode):
+        Vector2D.__init__(self, coordinates, y, z)
```
pointhi: where does the z comes from?

poeschlr:
Update: I have now implemented a first draft for generating paste automatically. I am happy with the resulting footprint but i do not like the implantation (4 deep nested for loops are extremely ugly)
To improve this situation i will add a few new classes that will improve this situation.

KicadModTree/nodes/specialized/ChamferedPad.py
```diff
@@ -124,17 +129,17 @@ def __getitem__(self, item):
...
-            self.bottom_right = value
+            self.bottom_right = bool(value)
         elif item in [3, CornerSelection.BOTTOM_LEFT]:
-            self.bottom_left = value
+            self.bottom_left = bool(value)
         else:
             raise IndexError('Index {} is out of range'.format(item))
 
-    def __dict__(self):
+    def to_dict(self):
```
pointhi: reason?

KicadModTree/nodes/specialized/ChamferedPad.py
```diff
@@ -0,0 +1,252 @@
...
+# KicadModTree is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with kicad-footprint-generator. If not, see < http://www.gnu.org/licenses/ >.
+#
+# (C) 2016 by Thomas Pointhuber, <thomas.pointhuber@gmx.at>
+# (C) 2018 by Rene Poeschl, github @poeschlr
```
pointhi: space ;)

KicadModTree/nodes/specialized/ChamferedPad.py
```diff
@@ -0,0 +1,252 @@
...
+class CornerSelection():
+    r"""Class for handling chamfer selection
+        :param chamfer_select:
+            * A list of bools do directly set the corners
+              (top left, top right, bottom right, bottom left)
+            * A dict with keys (constands see below)
+            * The integer 1 means all corners
+            * The integer 0 means no corners
+
+        :Constands:
```
pointhi: CONSTANTS

KicadModTree/nodes/specialized/FilledRect.py
```diff
@@ -28,9 +28,9 @@ class FilledRect(Node):
         See below
 
     :Keyword Arguments:
-        * *start* (``Point``) --
+        * *start* (``Vector``) --
```
pointhi: Type is actually Vector2D when I'm right

KicadModTree/nodes/specialized/PadArray.py
```diff
@@ -193,7 +199,25 @@ def _createPads(self, **kwargs):
             else:
                 kwargs['shape'] = padShape
 
-            pads.append(Pad(number=number, at=[x_pad, y_pad], **kwargs))
+            if i == 0 and kwargs.get('chamfer_corner_selection_first') is not None\
```
pointhi: 'chamfer_corner_selection_first' in kwars

poeschlr: We have a lot of places where we checked kwargs.get against None instead of checking 'key' in kwargs. I will go around and fix this at least in the places where i touched it in this PR.

KicadModTree/nodes/specialized/FilledRect.py
poeschlr: We did not specify the vector dimensions in any docstring. Fixed that (all instances are vector2d except the 3d model stuff)

KicadModTree/nodes/specialized/ChamferedPad.py
poeschlr: given in the commit message. __dict__ does not work in python2. Error message "'dictproxy' object is not callable"

I think __dict__ is reserved for a member that is a dict, not a function returning a dict.
https://docs.python.org/3/library/stdtypes.html#object.__dict__

poeschlr:
This is now ready for review. I am sure that i could clean up the code a lot if i would spend another few days on it but i think it should be good enough for now.
I plan to later add the option to generate exposed pads with the same via placement as the old qfn generator. (This would allow me to use the class inside that generator without significantly changing the generated files.)

poeschlr:
@pointhi poke

poeschlr:
@pointhi another poke

poeschlr:
@pointhi this PR is slowly turning into a monster. All the latest changes could have been in separate better documented PRs if this one would have gotten merged. So could you take a look at it? (Or give at least some form of feedback such that i know what the holdup is?)

pointhi:
I know, but at least until the next week, I likely don't have enough time to really review it.

Some time ago, I did a quick review, and it looked quite good. But I want to review such framework specific changes in detail before merging them.

If it is helpful for the future: you can still branch framework improvements from script improvements, and place them in different PR. Github should simplify the history when parts of an PR are merged upstream.

KicadModTree/nodes/specialized/ChamferedPad.py
pointhi: between header and code. But thats only style

KicadModTree/util/paramUtil.py
```diff
@@ -0,0 +1,166 @@
...
+          * Vector -> truncated to lenght
+
+    :param length: (``int``) -- default: 2
+        Defines the length of the resulting array
+
+    :min_value: (``int``) -- default: 1
+        Defines the minimum alowed value (raise value error if too low)
+        None -> No check
+    """
+    return toNumberArray(value, length, min_value, member_type=int)
```
pointhi: this should be ```float```.

pointhi:
@poeschlr: please fix the one bug above, and then I will merge.

pointhi:
Thanks for your hard work. That pull request was pretty huge

PR #126, opened 2018-04-24
Added BGA generator

Ratfink:
I was surprised to find that this repository lacked a BGA generator.  Since I needed to make a bunch of BGA footprints for https://github.com/KiCad/kicad-footprints/issues/374, I figured I should make one, so here it is, along with data for a bunch of BGAs.  This generator is heavily based on the QFN generator.

Still largely a work-in-progress, as the footprints it generates haven't been reviewed by another librarian yet.  I will update this when I believe it's ready to merge.

### DIFF
--- a/scripts/Package_BGA/bga.py
+++ b/scripts/Package_BGA/bga.py
@@ -1,0 +1,172 @@
+#!/usr/bin/env python3
+
+import math
+import os
+import sys
+
+# load parent path of KicadModTree
+sys.path.append(os.path.join(sys.path[0], "..", ".."))
+
+from KicadModTree import *
+
+def bga(args):
+    footprint_name = args["name"]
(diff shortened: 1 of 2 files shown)


poeschlr:
Is there some logic behind the die### suffixes? (can't find a reference to it in the first datasheet)

Ratfink:
Those suffixes are from ST's XML data that the symbol generator script uses.  Each chip has a die number, which for the WLCSP packages seems to be usable as a package identifier when combined with the number of balls.  There's a little more info about this [here](https://github.com/KiCad/Housings_CSP.pretty/issues/8#issuecomment-325613650).  I could add the pin pitches if you'd like, but leaving them out didn't create any name conflicts so I opted to make the names more concise.

Issue #117, opened 2018-03-06
Support for polygon pads?


evanshultz:
KiCad v5 adds polygon pads. Will KiCadModTree support them?

The RectFill function produces an approximation, but with many lines and only in rectangular shapes. Even when it works, it's more clunky.

pointhi:
You are right. Supporting polygons would simplify rectangular shapes for example by quite some amount. The Approximation always had the problem of being very prone to changes in line width.

I will look into the support of polygon lines as well as polygon pads if possible soon due to the fact they provide quite some improvement.

PR #128, opened 2018-04-30
Add support for kicad 5 primitives (roundrect, polygon, custom pad)

poeschlr:
**Work in progress** Feedback welcome

ToDo:
- [x] Add pad type roundrect with its additional attribute round radius
   - [x] Add initializers and pad type to Pad node
   - [x] Add support in pad serializer
      - My kicad nightly version is a bit inconsistent with where it puts whitespace in the fileformat. It does not put a space between layer info and round rect radius ratio. (Ignored for now.)
- [x] Add support for graphical polygon
   - [x] Find out if it is feasible to refactor the polyline node such that it and the graphical polygon can share code?
       - It is possible. Is now implemented
   - [x] Implement node for graphical polygon
   - [x] Implement serializer
- [x] Add support for custom pads
   - [x] Add support in pad node
   - [x] Add support in serializer (re use serializer of other nodes)
      - [x] polygon
      - [x] line
      - [x] arc
      - [x] circle
- [ ] Advanced features
  - [ ] cut, combine polygons (The later is not really necessary, but could result in nicer looking files)
     - [x] smaller polygon fully within larger polygon 
        - [ ] raise exception if not
     - [ ] polygons only need to partly overlap
         - [ ] check if the resulting polygon is still a single polygon (fail otherwise)
  - [x] support specialized nodes
    - [x] PolygoneLine
    - [x] Filled Rect (Filled arcs and circles might also be a good idea to have.)
  - [ ] check that all primitives intersect each other ("orphaned" primitives are not allowed)
    - kicad can read files that have pads with orphaned primitives. But i assume there is a reason why the devs build in a function not to allow making such a pad(Maybe DRC can not deal with it)

### DIFF
--- a/KicadModTree/Point.py
+++ b/KicadModTree/Point.py
@@ -16,6 +16,7 @@
 import warnings
 
 from KicadModTree.util.kicad_util import formatFloat
+from math import sqrt
 
 
 class Point2D(object):
@@ -135,6 +136,9 @@ class Point2D(object):
 
     def __str__(self):
         return "(x={x}, y={y})".format(**self.__dict__())
+
+    def distanceTo(self, other):
+        return sqrt((other.x - self.x)**2 + (other.y - self.y)**2)
 
 
 class Point3D(Point2D):
--- a/KicadModTree/Point.py
+++ b/KicadModTree/Point.py
@@ -76,6 +76,15 @@ class Point2D(object):
         return Point2D({'x': round(self.x / base) * base,
                         'y': round(self.y / base) * base})
 
+    def distance_to(self, value):
+        r"""Distance between selfe and another point
+
+        :param value: the other point
+        :return: distance between self and other point
+        """
+        other = Point2D.__arithmetic_parse(value)
+        return sqrt((other.x - self.x)**2 + (other.y - self.y)**2)
+
     @staticmethod
     def __arithmetic_parse(value):
         if isinstance(value, Point2D):
@@ -136,9 +145,6 @@ class Point2D(object):
 
     def __str__(self):
         return "(x={x}, y={y})".format(**self.__dict__())
-
-    def distanceTo(self, other):
-        return sqrt((other.x - self.x)**2 + (other.y - self.y)**2)
 
 
 class Point3D(Point2D):
--- a/KicadModTree/Point.py
+++ b/KicadModTree/Point.py
@@ -77,7 +77,7 @@ class Point2D(object):
                         'y': round(self.y / base) * base})
 
     def distance_to(self, value):
-        r"""Distance between selfe and another point
+        r"""Distance between this and another point
 
         :param value: the other point
         :return: distance between self and other point
--- a/KicadModTree/PolygonPoints.py
+++ b/KicadModTree/PolygonPoints.py
@@ -94,12 +94,12 @@ class PolygonPoints(object):
         return len(self.nodes)
 
     def findNearestPoints(self, other):
-        min_distance = self[0].distanceTo(other[0])
+        min_distance = self[0].distance_to(other[0])
         pi = 0
         pj = 0
         for i in range(len(self)):
             for j in range(len(other)):
-                d = self[i].distanceTo(other[j])
+                d = self[i].distance_to(other[j])
                 if d < min_distance:
                     pi = i
                     pj = j
--- a/KicadModTree/PolygonPoints.py
+++ b/KicadModTree/PolygonPoints.py
@@ -83,17 +83,17 @@ class PolygonPoints(object):
 
         return Node.calculateBoundingBox({'min': min, 'max': max})
 
-    def __iter__(self):
-        for n in self.nodes:
-            yield n
-
-    def __getitem__(self, idx):
-        return self.nodes[idx]
-
-    def __len__(self):
-        return len(self.nodes)
-
     def findNearestPoints(self, other):
+        r""" Find the nearest points for two polygons
+
+        Find the two points for both polygons that are nearest to each other.
+
+
+        :param other: the polygon points of the other polygon
+        :return: a tuble with the indexes of the two points
+                 (pint in self, point in other)
+        """
+
         min_distance = self[0].distance_to(other[0])
         pi = 0
         pj = 0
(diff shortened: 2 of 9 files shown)


pointhi:
graphical polygons are already implemented: https://github.com/pointhi/kicad-footprint-generator/issues/117, but it's at the moment not easily possible to create cutouts (I would like to abstract the hack KiCad currently does)

poeschlr:
Do you have any plans for this abstraction? I don't want to implement it in a way that does not fulfill your standards.

poeschlr:
I now have a first implantation of a data container for both currently implemented Polygon like Nodes.
As i am currently not very creative i named it PolyPoints for now. I might rename it if i can think of a better name.

To keep backwards compatibility this class accepts both the point list parameter of the Polygon and PolygoneLine Node (nodes and polygone)

This same container will then also be useful for the custom pad.

poeschlr:
I have a first implementation for the custom pad. For now only  the polygon primitive is supported. (For others i need to do some testing to find out how they work.)

pointhi:
> Do you have any plans for this abstraction? I don't want to implement it in a way that does not fulfill your standards.

Well, my only thought about this at the moment would be that a Polygone can be given an abritrary number of cutouts and this is internally converted to a single polyline as it's done in "Bitmap-to-component".

Allowing only one cutout for now should be enough for most use-cases (like capacitor-footprints) and this should be achivable without much hassle (search the closest points between inner and outer layer -> duplicate points and connect them together). Allowing multiple cutouts would require more complex algorithm where I don't know currently one.

poeschlr:
You can take a look at what i have so far. Everything except the more advanced features is now done.

pointhi:
The code changes look good so far. Only some of the style issues need to to be fixed.

:+1:

KicadModTree/PolygonPoints.py
```diff
@@ -0,0 +1,88 @@
...
+        self._initNodes(**kwargs)
+
+    def _initNodes(self, **kwargs):
+        self.nodes = []
+        if 'nodes' in kwargs:
+            for n in kwargs['nodes']:
+                self.nodes.append(Point2D(n))
+            if 'polygone' in kwargs:
+                raise KeyError('Use of "nodes" and "polygone" parameter at the same time is not supported.')
+        elif 'polygone' in kwargs:
```
pointhi: Add deprecation warning:

```python
warnings.warn(
            "polygone argument is deprecated, use nodes instead",
            DeprecationWarning
)
```

poeschlr: Good idea. Added. (Will be fun running one of the larger generator scripts. There will be a lot of warnings displayed.)

pointhi:
I would ignore this for now. This is not a user application and we can define constraints like "cutout is not allowed to overlap with the outline or other cutouts".

poeschlr:
I have now implemented a simple cut feature. The more complex nodes are now also supported as primitives for the custom pad.
As i make no checks at all i added a warning to the function call. I think we can implement such checks later on. (And make the cut function more resilient)

poeschlr:
@pointhi could you take another look at this? The functionality provided by this would help to create better qfn generators.

KicadModTree/Point.py
```diff
@@ -136,6 +137,9 @@ def __repr__(self):
     def __str__(self):
         return "(x={x}, y={y})".format(**self.__dict__())
 
+    def distanceTo(self, other):
```
pointhi: Move up to ```round_to``` function, rename to ```distance_to``` for name concistency and use ```Point2D.__arithmetic_parse``` to get the value for ```other```

pointhi: And add a docstring

KicadModTree/nodes/base/Pad.py
```diff
@@ -166,3 +216,6 @@ def _getRenderTreeText(self):
         render_text += '({})'.format(' '.join(render_strings))
 
         return render_text
+
+    def addPrimitive(self, p):
```
pointhi: add docstring

KicadModTree/PolygonPoints.py
```diff
@@ -0,0 +1,126 @@
...
+                    pi = i
+                    pj = j
+                    min_distance = d
+
+        return (pi, pj)
+
+    def getPoints(self):
+        return self.nodes
+
+    def cut(self, other):
```
pointhi: add docstring

KicadModTree/PolygonPoints.py
```diff
@@ -0,0 +1,126 @@
...
+        for n in self.nodes:
+            yield n
+
+    def __getitem__(self, idx):
+        return self.nodes[idx]
+
+    def __len__(self):
+        return len(self.nodes)
+
+    def findNearestPoints(self, other):
```
pointhi: add docstring

KicadModTree/nodes/base/Polygon.py
```diff
@@ -76,3 +67,6 @@ def _getRenderTreeText(self):
         render_text += "]"
 
         return render_text
+
+    def cut(self, other):
```
pointhi: add docstring

KicadModTree/PolygonPoints.py
```diff
@@ -108,9 +108,26 @@ def findNearestPoints(self, other):
...
     def cut(self, other):
+        r""" Cut other polygon points from self
+
+        As kicad has no native support for cuting one polygon from the other,
+        the cut is done by connecting the nearest points of the two polygons
+        with two lines on top of each other.
+
+        This function assumes that the other polygon is fully within this one.
+        It also assumes that connecting the two nearest points creates a valid
+        polygon. (There are no geomtry checks)
```
pointhi: geomtry -> geometry

KicadModTree/Point.py
```diff
@@ -76,6 +76,15 @@ def round_to(self, base):
         return Point2D({'x': round(self.x / base) * base,
                         'y': round(self.y / base) * base})
 
+    def distance_to(self, value):
+        r"""Distance between selfe and another point
```
pointhi: selfe -> this and another point

pointhi:
Thanks for your hard work. Should be file now 👍

PR #133, opened 2018-06-07
Add newer ipc definitions for smd chip packages

poeschlr:
As the title says, this adds the ipc definitions from the newer ipc7315b standard. The old 7315 definition is unchanged (other than removing a duplicate unneeded definition and adding an explanation about the fact that castelated exchanges the meaning of toe and heel in the ipc document.)

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/642


poeschlr:
Fixed the comments

poeschlr:
This will be included in my larger pull request #125 as i plan to further update the generators to IPC-7351C (Using rounded rectangle pads.)

PR #135, opened 2018-06-19
Fix terminal block NPTH plus some minor improvements

poeschlr:
- NPTH holes should have the pad size equal to the drill size. (was
drillsize 0)
- Remove pad numbers from NPTH
- Resulting footprints are stored inside a .pretty lib instead of the
root directory.
- Automatic codestyle changes (Whatever atom does automatically.)

footprints are added with https://github.com/KiCad/kicad-footprints/pull/673


poeschlr:
Footprints have been merged. This should be ready to go.

pointhi:
Thanks

PR #136, opened 2018-06-20
Script to generate Micro-Fit 3.0 THT Side Dual and Single row.

jneiva08:
Created two scripts to create Micro-Fit 3.0 tht 90º dual and single row


jneiva08:
fixed this:
* Changed mounting_pad name to ""
* Now have right part_code on single row,
* In double row have right number of pins

now shoud be right

jneiva08:
changed drill size to 1.02mm like datasheet says

pointhi:
Thanks for your contribution

PR #141, opened 2018-07-01
Fix footprints of RECOM DC/DC converters.

rschaeuble:
Some had their pins towards the front instead of towards the back, some had a wrong  pin 1 offset, and other had wrong widths.

Also refactored the script to put different families of controllers into different functions.

See https://github.com/KiCad/kicad-footprints/pull/688 for the original pull request which triggered this one.


rschaeuble:
The footprint changes have been merged, so this one can now also be merged.

rschaeuble:
@pointhi: as the footprints have already been merged, could you please merge this pull request?

rschaeuble:
@pointhi Is there anything that still needs to be done before this can be merged?

As the footprints have already been merged, it would be great if the generator changes could be merged soon as well.

pointhi:
@rschaeuble I will merge now. Thanks for your contribution

PR #149, opened 2018-07-16
Add further qfn dimensions

poeschlr:

PR #150, opened 2018-07-19
Mask/paste global footprint support & new BGA entries

evanshultz:
When added the footprints at https://github.com/KiCad/kicad-footprints/pull/718, @Ratfink mentioned that I should use his generator. He was right.

I added the footprint entries, then found two things:
1. KiCad comes with Python 2.7 and I used that for broadest compatibility (why they've not gone to 3 I don't know...).
2. This framework didn't support the global settings for pad mask/paste.

This commit adds the BGAs info the YAML file, updates the BGA script to support global mask/paste settings, and also updates the framework to support the same.

Clay, you probably should take a look here. Does everything look OK to you? I realized it really make no sense to use per-pad mask/paste settings since it would be applied to every single pad in the footprint, and so this only made sense. I also aligned some of the variable terminology with the framework.

Quick per-file breakdown of changes:
bga.yml
- Added entries for 3 TI footprints

bga.py
- Made denominators floats for Python 2.7 compatibility
- Simplified CSP BGA 3D model repo determination
- Support CSP in keywords if appropriate
- Silkscreen offset to 0.11mm (was 0.12mm)
- Support mask/paste clearance parameters for whole footprint (not per pad)

Pad.py
- Fixed typo
- Minor wordsmithing
- Used a more explicit variable name since elsewhere is full words

Footprint.py / KicadFileHandler.py:
- Support mask margin, paste margin, and paste margin ratio

### DIFF
--- a/KicadModTree/nodes/Footprint.py
+++ b/KicadModTree/nodes/Footprint.py
@@ -30,7 +30,7 @@ ADVANTAGES:
 '''
 
 # define in which order the general "lisp" operators are arranged
-render_order = ['descr', 'tags', 'attr', 'fp_text', 'fp_circle', 'fp_line', 'pad', 'model']
+render_order = ['descr', 'tags', 'attr', 'solder_mask_margin', 'solder_paste_margin', 'solder_paste_ratio', 'fp_text', 'fp_circle', 'fp_line', 'pad', 'model']
 # TODO: sort Text by type
 
 
@@ -45,6 +45,9 @@ class Footprint(Node):
         self.description = None
         self.tags = None
         self.attribute = None
+        self.maskMargin = None
+        self.pasteMargin = None
+        self.pasteMarginRatio = None
 
     def setName(self, name):
         self.name = name
@@ -57,3 +60,15 @@ class Footprint(Node):
 
     def setAttribute(self, value):
         self.attribute = value
+        
+    def setMaskMargin(self, value):
+        self.maskMargin = value
+
+    def setPasteMargin(self, value):
+        self.pasteMargin = value
+    
+    def setPasteMarginRatio(self, value):
+        # paste_margin_ratio is a percentage between 0 and 1 but user could enter exact ratio or percentage
+        if value > 1:
+            value = value / 100.0
+        self.pasteMarginRatio = value
--- a/KicadModTree/nodes/Footprint.py
+++ b/KicadModTree/nodes/Footprint.py
@@ -70,8 +70,8 @@ class Footprint(Node):
         self.pasteMargin = value
 
     def setPasteMarginRatio(self, value):
-        # paste_margin_ratio is unitless between 0 and 1
-        # GUI uses percentage so account for that input
+        # paste_margin_ratio is unitless between 0 and 1 while GUI uses percentage
         if abs(value) > 1:
-            value = value / 100.0
-        self.pasteMarginRatio = value
+            raise ValueError("Solder paste margin must be between -1 and 1. {} is too large.".format(value))
+        else
+            self.pasteMarginRatio = value
--- a/KicadModTree/nodes/Footprint.py
+++ b/KicadModTree/nodes/Footprint.py
@@ -73,5 +73,5 @@ class Footprint(Node):
         # paste_margin_ratio is unitless between 0 and 1
         # GUI uses percentage so account for that input
         if value > 1:
-            value = value / 100.0
+            value = abs(value) / 100.0
         self.pasteMarginRatio = value
--- a/KicadModTree/nodes/Footprint.py
+++ b/KicadModTree/nodes/Footprint.py
@@ -30,7 +30,9 @@ ADVANTAGES:
 '''
 
 # define in which order the general "lisp" operators are arranged
-render_order = ['descr', 'tags', 'attr', 'solder_mask_margin', 'solder_paste_margin', 'solder_paste_ratio', 'fp_text', 'fp_circle', 'fp_line', 'pad', 'model']
+render_order = ['descr', 'tags', 'attr', 'solder_mask_margin',
+                    'solder_paste_margin', 'solder_paste_ratio', 'fp_text',
+                    'fp_circle', 'fp_line', 'pad', 'model']
 # TODO: sort Text by type
 
 
@@ -60,15 +62,16 @@ class Footprint(Node):
 
     def setAttribute(self, value):
         self.attribute = value
-        
+
     def setMaskMargin(self, value):
         self.maskMargin = value
 
     def setPasteMargin(self, value):
         self.pasteMargin = value
-    
+
     def setPasteMarginRatio(self, value):
-        # paste_margin_ratio is a percentage between 0 and 1 but user could enter exact ratio or percentage
+        # paste_margin_ratio is unitless between 0 and 1
+        # GUI uses percentage so account for that input
         if value > 1:
             value = value / 100.0
         self.pasteMarginRatio = value
--- a/KicadModTree/nodes/Footprint.py
+++ b/KicadModTree/nodes/Footprint.py
@@ -71,7 +71,6 @@ class Footprint(Node):
 
     def setPasteMarginRatio(self, value):
         # paste_margin_ratio is unitless between 0 and 1 while GUI uses percentage
-        if abs(value) > 1:
-            raise ValueError("Solder paste margin must be between -1 and 1. {} is too large.".format(value))
-        else:
-            self.pasteMarginRatio = value
+        assert abs(value) <= 1,"Solder paste margin must be between -1 and 1. {} is too large.".format(value)
+        
+        self.pasteMarginRatio = value
--- a/KicadModTree/nodes/Footprint.py
+++ b/KicadModTree/nodes/Footprint.py
@@ -72,6 +72,6 @@ class Footprint(Node):
     def setPasteMarginRatio(self, value):
         # paste_margin_ratio is unitless between 0 and 1
         # GUI uses percentage so account for that input
-        if value > 1:
-            value = abs(value) / 100.0
+        if abs(value) > 1:
+            value = value / 100.0
         self.pasteMarginRatio = value
--- a/KicadModTree/nodes/Footprint.py
+++ b/KicadModTree/nodes/Footprint.py
@@ -31,8 +31,8 @@ ADVANTAGES:
 
 # define in which order the general "lisp" operators are arranged
 render_order = ['descr', 'tags', 'attr', 'solder_mask_margin',
-                    'solder_paste_margin', 'solder_paste_ratio', 'fp_text',
-                    'fp_circle', 'fp_line', 'pad', 'model']
+                'solder_paste_margin', 'solder_paste_ratio', 'fp_text',
+                'fp_circle', 'fp_line', 'pad', 'model']
 # TODO: sort Text by type
 
 
@@ -71,6 +71,6 @@ class Footprint(Node):
 
     def setPasteMarginRatio(self, value):
         # paste_margin_ratio is unitless between 0 and 1 while GUI uses percentage
-        assert abs(value) <= 1,"Solder paste margin must be between -1 and 1. {} is too large.".format(value)
-        
+        assert abs(value) <= 1, "Solder paste margin must be between -1 and 1. {} is too large.".format(value)
+
         self.pasteMarginRatio = value
--- a/KicadModTree/nodes/Footprint.py
+++ b/KicadModTree/nodes/Footprint.py
@@ -73,5 +73,5 @@ class Footprint(Node):
         # paste_margin_ratio is unitless between 0 and 1 while GUI uses percentage
         if abs(value) > 1:
             raise ValueError("Solder paste margin must be between -1 and 1. {} is too large.".format(value))
-        else
+        else:
             self.pasteMarginRatio = value
--- a/scripts/Package_BGA/bga.py
+++ b/scripts/Package_BGA/bga.py
@@ -25,7 +25,7 @@ def bga(args):
     layoutY = args["layout_y"]
     rowNames = args["row_names"]
     rowSkips = args["row_skips"]
-    
+
     f = Footprint(footprint_name)
     f.setDescription(desc)
     f.setAttribute("smd")
@@ -33,9 +33,11 @@ def bga(args):
     f.setPasteMargin(pasteMargin)
     f.setPasteMarginRatio(pasteRatio)
     
+    package_type = 'CSP' if 'BGA' not in footprint_name and 'CSP' in footprint_name else 'BGA'
+    
     # If this looks like a CSP footprint, use the CSP 3dshapes library
     f.append(Model(filename="${{KISYS3DMOD}}/Package_{}.3dshapes/{}.wrl".format(
-                  'CSP' if 'BGA' not in footprint_name and 'CSP' in footprint_name else 'BGA', footprint_name)))
+                  package_type, footprint_name)))
 
     s1 = [1.0, 1.0]
     s2 = [min(1.0, round(pkgWidth / 4.3, 2))] * 2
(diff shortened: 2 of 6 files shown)


evanshultz:
Fixed some issues noted above. Other than the over-indent (which I can adjust if you tell me the style you prefer) the remaining 4 issues were already in the code before I started so I've left them alone. If you do want some change on those items, push to my branch or let me know and I'll do it. Thanks!

scripts/Package_BGA/bga.py
```diff
@@ -18,29 +18,24 @@ def bga(args):
...
-    pasteDiameter = args["paste_diameter"]
+    maskMargin = args["mask_margin"]
+    pasteMargin = args["paste_margin"]
+    pasteRatio = args["paste_ratio"] # '10.0' means paste ratio is 10% of pad size
     layoutX = args["layout_x"]
     layoutY = args["layout_y"]
     rowNames = args["row_names"]
     rowSkips = args["row_skips"]
-
+
```
Ratfink: No need for these spaces.

scripts/Package_BGA/bga.py
```diff
@@ -50,40 +45,37 @@ def bga(args):
...
+    xRightFab = xCenter + pkgWidth / 2.0
     xChamferFab = xLeftFab + chamfer
-    xPadLeft = xCenter - pitch * ((layoutX - 1) / 2)
-    xPadRight = xCenter + pitch * ((layoutX - 1) / 2)
-    xLeftCrtYd = crtYdRound(xCenter - (pkgWidth / 2 + crtYd))
-    xRightCrtYd = crtYdRound(xCenter + (pkgWidth / 2 + crtYd))
+    xPadLeft = xCenter - pitch * ((layoutX - 1) / 2.0)
+    xPadRight = xCenter + pitch * ((layoutX - 1) / 2.0)
+    xLeftCrtYd = crtYdRound(xCenter - (pkgWidth / 2 + crtYdOffset))
+    xRightCrtYd = crtYdRound(xCenter + (pkgWidth / 2 + crtYdOffset))
```
Ratfink: While you're forcing floating-point division elsewhere, you should probably update these lines too.

scripts/Package_BGA/bga.py
```diff
@@ -50,40 +45,37 @@ def bga(args):
...
+    yBottomFab = yCenter + pkgHeight / 2.0
     yChamferFab = yTopFab + chamfer
-    yPadTop = yCenter - pitch * ((layoutY - 1) / 2)
-    yPadBottom = yCenter + pitch * ((layoutY - 1) / 2)
-    yTopCrtYd = crtYdRound(yCenter - (pkgHeight / 2 + crtYd))
-    yBottomCrtYd = crtYdRound(yCenter + (pkgHeight / 2 + crtYd))
+    yPadTop = yCenter - pitch * ((layoutY - 1) / 2.0)
+    yPadBottom = yCenter + pitch * ((layoutY - 1) / 2.0)
+    yTopCrtYd = crtYdRound(yCenter - (pkgHeight / 2 + crtYdOffset))
+    yBottomCrtYd = crtYdRound(yCenter + (pkgHeight / 2 + crtYdOffset))
```
Ratfink: More places to force floating-point division.

scripts/Package_BGA/bga.py
```diff
@@ -149,11 +141,10 @@ def crtYdRound(x):
                          shape=padShape,
                          at=[xPadLeft + (col-1) * pitch, yPadTop + rowNum * pitch],
                          size=[padDiameter, padDiameter],
-                         layers=Pad.LAYERS_SMT,
-                         solder_paste_margin_ratio=pasteRatio))
-
-    f.setTags("BGA {} {}".format(balls, pitch))
+                         layers=Pad.LAYERS_SMT))
 
+    f.setTags("{} {} {}".format('CSP' if 'BGA' not in footprint_name and 'CSP' in footprint_name else 'BGA', balls, pitch))
```
Ratfink: This is the second time you wrote `'CSP' if 'BGA' not in footprint_name and 'CSP' in footprint_name else 'BGA'`.  I strongly suggest you only do that once and store the resulting string to a variable.

scripts/Package_BGA/bga.py
```diff
@@ -166,8 +157,9 @@ def crtYdRound(x):
     parser.add_parameter("pkg_height", type=float, required=True)
     parser.add_parameter("pitch", type=float, required=True)
     parser.add_parameter("pad_diameter", type=float, required=True)
+    parser.add_parameter("mask_margin", type=float, required=False, default=0)
+    parser.add_parameter("paste_margin", type=float, required=False, default=0)
     parser.add_parameter("paste_ratio", type=float, required=False, default=0)
-    parser.add_parameter("paste_diameter", type=float, required=False, default=0)
```
Ratfink: Since you changed this option, be sure to update the .yml files to match.  It's only used in csp.yml I believe.

evanshultz:
@Ratfink 
Thanks. Good suggestions all around. All are fixed in the last commit.

I also noticed in the 2nd CSP datasheet that they specify a mask opening size, but I was too lazy to go through all the datasheets and add that.

KicadModTree/nodes/Footprint.py
```diff
@@ -73,5 +73,5 @@ def setPasteMarginRatio(self, value):
         # paste_margin_ratio is unitless between 0 and 1
         # GUI uses percentage so account for that input
         if value > 1:
-            value = value / 100.0
+            value = abs(value) / 100.0
```
Ratfink: Calling `abs` is redundant here: the line is only run if value > 1 (and is therefore positive), so it's a no-op.

scripts/Package_BGA/bga.py
```diff
@@ -65,17 +67,17 @@ def crtYdRound(x):
     xChamferFab = xLeftFab + chamfer
     xPadLeft = xCenter - pitch * ((layoutX - 1) / 2.0)
     xPadRight = xCenter + pitch * ((layoutX - 1) / 2.0)
-    xLeftCrtYd = crtYdRound(xCenter - (pkgWidth / 2 + crtYdOffset))
-    xRightCrtYd = crtYdRound(xCenter + (pkgWidth / 2 + crtYdOffset))
+    xLeftCrtYd = crtYdRound(xCenter - (pkgWidth / 2 + float(crtYdOffset)))
+    xRightCrtYd = crtYdRound(xCenter + (pkgWidth / 2 + float(crtYdOffset)))
```
Ratfink: This still doesn't do floating-point division (I meant to write `pkgWidth / 2.0` instead of `pkgWidth / 2`).

scripts/Package_BGA/bga.py
```diff
@@ -65,17 +67,17 @@ def crtYdRound(x):
...
     yCenter = 0.0
     yTopFab = yCenter - pkgHeight / 2.0
     yBottomFab = yCenter + pkgHeight / 2.0
     yChamferFab = yTopFab + chamfer
     yPadTop = yCenter - pitch * ((layoutY - 1) / 2.0)
     yPadBottom = yCenter + pitch * ((layoutY - 1) / 2.0)
-    yTopCrtYd = crtYdRound(yCenter - (pkgHeight / 2 + crtYdOffset))
-    yBottomCrtYd = crtYdRound(yCenter + (pkgHeight / 2 + crtYdOffset))
+    yTopCrtYd = crtYdRound(yCenter - (pkgHeight / 2 + float(crtYdOffset)))
+    yBottomCrtYd = crtYdRound(yCenter + (pkgHeight / 2 + float(crtYdOffset)))
```
Ratfink: Same as above.

scripts/Package_BGA/csp.yml
```diff
@@ -12,7 +12,7 @@ ST_WLCSP-25_Die444:
   pkg_height: 2.325
   pitch: 0.4
   pad_diameter: 0.225
-  paste_diameter: 0.25
+  paste_margin: 0.25
```
Ratfink: `paste_margin` isn't simply a new name for `paste_diameter` - what you've done here makes paste openings that extend 0.25 mm beyond the edges of the pads, but they should only extend 0.0125 mm beyond the edges of the pad according to the datasheet.

scripts/Package_BGA/bga.py
```diff
@@ -50,40 +47,37 @@ def bga(args):
...
+    xRightFab = xCenter + pkgWidth / 2.0
     xChamferFab = xLeftFab + chamfer
-    xPadLeft = xCenter - pitch * ((layoutX - 1) / 2)
-    xPadRight = xCenter + pitch * ((layoutX - 1) / 2)
-    xLeftCrtYd = crtYdRound(xCenter - (pkgWidth / 2 + crtYd))
-    xRightCrtYd = crtYdRound(xCenter + (pkgWidth / 2 + crtYd))
+    xPadLeft = xCenter - pitch * ((layoutX - 1) / 2.0)
+    xPadRight = xCenter + pitch * ((layoutX - 1) / 2.0)
+    xLeftCrtYd = crtYdRound(xCenter - (pkgWidth / 2 + float(crtYdOffset)))
+    xRightCrtYd = crtYdRound(xCenter + (pkgWidth / 2 + float(crtYdOffset)))
```
Ratfink: I meant to change this to `pkgWidth / 2.0` to force floating-point division.  Sorry for not being clearer.

scripts/Package_BGA/bga.py
```diff
@@ -50,40 +47,37 @@ def bga(args):
...
+    yBottomFab = yCenter + pkgHeight / 2.0
     yChamferFab = yTopFab + chamfer
-    yPadTop = yCenter - pitch * ((layoutY - 1) / 2)
-    yPadBottom = yCenter + pitch * ((layoutY - 1) / 2)
-    yTopCrtYd = crtYdRound(yCenter - (pkgHeight / 2 + crtYd))
-    yBottomCrtYd = crtYdRound(yCenter + (pkgHeight / 2 + crtYd))
+    yPadTop = yCenter - pitch * ((layoutY - 1) / 2.0)
+    yPadBottom = yCenter + pitch * ((layoutY - 1) / 2.0)
+    yTopCrtYd = crtYdRound(yCenter - (pkgHeight / 2 + float(crtYdOffset)))
+    yBottomCrtYd = crtYdRound(yCenter + (pkgHeight / 2 + float(crtYdOffset)))
```
Ratfink: Same as above.

evanshultz:
Right. Thanks. All should be updated; I added specific mask sizes if the datasheet called for it and also used a more concise page-specific URL. All good with you now?

https://www.st.com/resource/en/datasheet/stm32f378vc.pdf#page=114 and https://www.st.com/resource/en/datasheet/stm32f469ni.pdf#page=198 don't give pad info; where did you get their pad sizes?

Ratfink:
Yep, looks good to me now!

The pads of those footprints are based on other packages with the same ball sizes.  For a bit more information, including IPC recommendations, see https://github.com/KiCad/kicad-footprints/pull/522 where those footprints were reviewed originally.

evanshultz:
@Ratfink 
I updated one of the packages mentioned above. Thanks for that info.

And I realized I missed a few CSPs. I updated those, but one of them matched the second link above in having 0.27mm ball diameter and no padstack info. They were the only ones with 0.24mm pads, so I suppose that's OK (89% of ball diameter) and those 2 are unique. However, there's no mask or paste info given. I set paste to 1nm since that's fine but I left mask blank. All other CSPs have 0.025 to 0.04mm mask margin, so 0.0325 would be a safe choice but do you think it should stay blank?

One last question: I copied your logic of `package_type = 'CSP' if 'BGA' not in footprint_name and 'CSP' in footprint_name else 'BGA'`, but why not just use `package_type = 'CSP' if 'CSP' in footprint_name else 'BGA'`? Wouldn't that be simpler? It works for me. Is there an edge case I'm missing?

Ratfink:
The problems you mentioned can happen for any pads.  The user should set the global mask clearance so they don't happen.  We don't know what the fab's tolerances are, so when the manufacturer has no recommendation, we don't know what's right, and making a guess probably won't help.  I don't think I have enough expertise to make a final decision on something like this, but that's my thought.

evanshultz:
All round pads and circular balls mean BGAs/CSPs center during reflow, which does not apply to all package types. Mask should always be bigger irrespective of anything the user does.

@poeschlr 
Are you OK with 0.0325mm mask oversize on the two CSPs mentioned above that don't specify the mask oversize?

poeschlr:
I do not really have any opinion on mask size for the two CSP footprints. (Not really my field of expertise.)

Your changes to the pad.py file (spelling changes) should have been made in a separate commit. Having it mixed in with other larger changes could potentially make merging a lot harder. (Luckily i did not touch the Pad class in any of my open pull requests so we should be good here. But keep that in mind for the future.)

evanshultz:
I've set the mask oversize on the two that didn't specify it.

@Ratfink 
Yes, but all circular pads with circular pins/balls/bumps mean these footprints are a special case when the reflow soldering process is considered..

@pointhi 
This should be ready to merge.

KicadModTree/nodes/Footprint.py
```diff
@@ -57,3 +62,16 @@ def setTags(self, tags):
...
+    def setMaskMargin(self, value):
+        self.maskMargin = value
+
+    def setPasteMargin(self, value):
+        self.pasteMargin = value
+
+    def setPasteMarginRatio(self, value):
+        # paste_margin_ratio is unitless between 0 and 1
+        # GUI uses percentage so account for that input
+        if abs(value) > 1:
```
pointhi: don't do such magic. Rather add an assert when users provide invalid values

evanshultz:
In the UI paste margin is entered in percentage up to 100, but a unit value matches the other code I saw. Changed to only allow up to -1 to 1 and raise an error otherwise. Let me know if you would like something else.

pointhi:
I would tend to -1 ... 1 because that seems more natural as an engineer for percent values.

evanshultz:
Me too. My last patch accepts only -1 to 1. Was there something about the wording or code that needs changed?

pointhi:
Please fix travis errors, than I can merge

evanshultz:
Duh. My fault. Fixed.

KicadModTree/nodes/Footprint.py
```diff
@@ -57,3 +62,16 @@ def setTags(self, tags):
...
+
+    def setMaskMargin(self, value):
+        self.maskMargin = value
+
+    def setPasteMargin(self, value):
+        self.pasteMargin = value
+
+    def setPasteMarginRatio(self, value):
+        # paste_margin_ratio is unitless between 0 and 1 while GUI uses percentage
+        if abs(value) > 1:
```
pointhi: input validation using asserts would be cleaner I think:

```python
assert abs(value) <= 1
```

this would also be possible :):

```python
assert -1 <= value <= 1
```

evanshultz:
Like this?

The previous commit following the convention I saw used in https://github.com/pointhi/kicad-footprint-generator/blob/master/KicadModTree/util/paramUtil.py.

pointhi:
@evanshultz ,

what's the current state? Please fix travis errors and merge conflicts if you want this merged

evanshultz:
The current state is that merging #149 created the conflict. Fixed now. I also addressed the Travis issues but let me know if something isn't to your satisfaction or the coding style you prefer.

scripts/Package_BGA/bga.py
```diff
@@ -18,8 +18,9 @@ def bga(args):
 
     pitch = args["pitch"]
     padDiameter = args["pad_diameter"]
-    pasteRatio = args["paste_ratio"]
-    pasteDiameter = args["paste_diameter"]
+    maskMargin = args["mask_margin"]
+    pasteMargin = args["paste_margin"]
+    pasteRatio = args["paste_ratio"] # '10.0' means paste ratio is 10% of pad size
```
pointhi: I think the comment is no longer correct

evanshultz: I deleted that comment.

pointhi:
@evanshultz, thanks for your contribution

PR #153, opened 2018-07-20
Add 3.81mm THT diode hairpin footprints

evanshultz:
Also removed a few special characters from drawing_tools.py which didn't have encoding specified.


evanshultz:
@pointhi 
Footprints have been merged.

pointhi:
thanks

PR #155, opened 2018-07-24
Fix MPN for conn samtec hle

poeschlr:
Issue reported in https://github.com/KiCad/kicad-footprints/issues/776


mikelawrence:
MPN HLE-150-XXXXXX footprints are missing. Looks like range for 'pins_per_row_range' is set to range(4,50) instead of range(4,51) on lines 68, 78, 88, 97, 106, 115, 122, 128 and 139.

poeschlr:
Fixed

poeschlr:
Will be included in #161

PR #157, opened 2018-08-05
Fix pin number letter case and fp name

poeschlr:
- connector types with full configuration now fit their respectife
generic symbol (other types need special symbols)
- footprint names now fit the scheme already used in the lib (seems
there was a manual update sometime back that did not make it into the
scripts)


poeschlr:
@pointhi the changes have been merged in the footprint repo so this would be ready to go.

pointhi:
thanks

PR #158, opened 2018-08-08
added QFN dimensions - Texas_RGE0024H

s-light:
added dimensions for Texas Instruments RGE0024H VQFN part.
seems to be a *common* housing if the [results of the ti-search](http://www.ti.com/sitesearch/docs/universalsearch.tsp?searchTerm=RGE0024H#linkId=3) are meaningful ;-)


poeschlr:
I will include this in my pull request https://github.com/pointhi/kicad-footprint-generator/pull/149 as soon as i get reliably internet again. (It seems to be too warm for my modem)

scripts/Packages/Package_DFN_QFN/size_definitions/qfn_texas.yaml
```diff
@@ -310,3 +310,55 @@ Texas_S-PVQFN-N64_EP:
...
   #include_suffix_in_3dpath: 'False'
+
+Texas_VQFN_RGE0024H:
+  device_type: 'QFN'
+  manufacturer: 'Texas'
+  #part_number: 'mpn'
+  size_source: 'http://www.ti.com/lit/ds/symlink/tlc5971.pdf#page=42&zoom=200,-13,779'
+  ipc_class: 'qfn' # 'qfn_pull_back'
+  #ipc_density: 'least' #overwrite global value for this device.
+  custom_name_format: 'Texas_RGE0024H{vias:s}'
```
poeschlr: I would suggest to change the name format to:
custom_name_format: 'Texas_RGE00{pincount}H_EP{ep_size_x:g}x{ep_size_y:g}mm{vias:s}'

scripts/Packages/Package_DFN_QFN/size_definitions/qfn_texas.yaml
```diff
@@ -310,3 +310,55 @@ Texas_S-PVQFN-N64_EP:
...
+  lead_len_min: 0.28
+  lead_len_max: 0.48
+
+  EP_size_x_min: 2.6
+  EP_size_x: 2.7
+  EP_size_x_max: 2.8
+  EP_size_y_min: 2.6
+  EP_size_y: 2.7
+  EP_size_y_max: 2.8
+  EP_paste_coverage: 0.78
```
poeschlr: Is there a reason for this much paste coverage? We normally aim for about 65%.
(Especially for the non Thermal vias option. The thermal vias option can have a higher coverage. This helps combat paste loss if the paste overlaps the vias like in this case.)

scripts/Packages/Package_DFN_QFN/size_definitions/qfn_texas.yaml
s-light: thanks for your feedback.
iam happy with integrating the EP size..
currently there are other TI parts that also lag it.. 
and i don't know if it is something that can change regarding the `RGE`prefix

for the same i have leaved out the pincount  - i found only results with 24 pin for RGE... so to me it seems this is a fixed term...

but as told iam happy changing it..

poeschlr: I would really suggest including the EP size for new footprints. (I recently changed the wording of the KLC rules to make this more clear.)
All script generated TI packages include it. I have not yet come around to rework all qfn footprints. Some of the still missing footprints are included in two larger pull requests of mine. (Adding the EP size to the existing footprint names might also require changes on the symbol side. This is why i did not yet finish all of them as it is quite easy to generate merge conflicts if the changes are too big.)

scripts/Packages/Package_DFN_QFN/size_definitions/qfn_texas.yaml
s-light: this comes directly from the datasheet...
http://www.ti.com/lit/ds/symlink/tlc5971.pdf#page=45&zoom=140,-151,749
> SOLDER PASTE EXAMPLE
BASED ON 0.125 mm THICK STENCIL
EXPOSED PAD
78% PRINTED COVERAGE BY AREA
SCALE: 20X

i don't know what the 'correct' way here is -
the datasheet also recommends `0.24mm pad width` - but the script calculations generate `0.25mm`

poeschlr: The coverage assumes that some solder will be lost to the suggested thermal vias. As i suggested above you can use a lower coverage for the non thermal via option.

Regarding pad width: If you want your part true to the datasheet then this script is the wrong approach. (You need to trust IPC and my implementation of IPC to use this script.)

s-light:
is there anything left todo i missed?
@poeschlr if i have seen this correct your #149 is merged. but as fare as i can see RGE0024H is not landed in master yet.

pointhi:
@poeschlr can this be merged?

pointhi:
@poeschlr I will merge it now. Seems the file was not included in #149

PR #161, opened 2018-08-17
Use round rect for pin 1 marking in THT footprints

poeschlr:
This pull request will eventually update all connector footprints to use rounded rectangle pads for pin 1 instead of rectangle pads. There will be a fallback mode for kicad4 compatibility.

**Important: This is based on top of #149 as i want to avoid merge conflicts in pad array.** I added some functionality to it in that particular pull request. This pull request also touches that file. (This means that this pull request should not be merged before #149 is merged.)


evanshultz:
One question: What is the reason/standard for doing this? Not that round rect is wrong, but IPC 7351C will recommend this pad shape and it is only for SMT. Is there a recommendation for this pad shape on THT pads also?

poeschlr:
As far as i know using rounded rectangle pads has some benefits for solder quality. (But i could be confused here and what i read was only true for SMD)

Using rounded rectangle everywhere (instead of rectangle) will also make KLC easier to write. (Would also make it easy for writing check scripts.)

I do not really see a downside to using rounded rectangle pads for THT as well. (Other than it being a bit of a tedious work. But i have now converted over all connector footprints so this argument is moot after the fact.)

poeschlr:
We use oval pads for all other pads so i don't really see a technical reason why the one pad could not be rounded rectangle. Exempt does not mean "should not use that" but more "no need to invest time".

---

The exception for thermal pads however is more of a concern. Right now we round them as well. (Would affect my other open pull request here)
Whats also strange: Why are DFN exempt even tough they have the same lead style as QFN which is explicitly shown on the previous page?
LGA is also exempt. But do they mean the things that are called LGA by manufacturers (ie the things we have in the lib) or do they mean the packages that are the real LGAs?

KicadModTree/nodes/specialized/PadArray.py
```diff
@@ -44,11 +46,11 @@ class PadArray(Node):
           y offset between rendered pads
         * *initial* (``int``) --
           name of the first pad
-        * *increment* (``int``) --
+        * *increment* (``int, function(previous_number)``) --
```
pointhi: hm, instead of ```function(previous_number)``` an iterator would make more sense

pointhi: or you should give an x/y index and the function returns the id. Would be more general I think and easier to understand.

evanshultz:
@poeschlr 
Can you make the changes above so that once #149 is merged this one can be too?

poeschlr:
I have very little time for working on kicad stuff right now. Will take a while

evanshultz:
@poeschlr 
Now that #149 is merged this is my next target. Any ETA on having time to address the review comment, branch conflicts, and Travis errors here?

evanshultz:
@poeschlr 
Do you have time for KiCad stuff now? I would like to see this merged so that I can submit a PR for https://github.com/KiCad/kicad-footprints/issues/911.

poeschlr:
Currently i am a bit limited in time as it is harvest season right now. (My parents have a farm and all hands are needed during this time)
Another time limiting factor is that the new study year starts with next month and i need to prepare some marketing stuff for the student organization i work at (we need this stuff for recruiting new members)
In short end of september till mid of october is extremely busy for me. If you are in a hurry you are free to fix the mentioned stuff in your pull request.

evanshultz:
I got some info related to the above discussion:
- Thermal pads can be rounded or rectangle. It seems that the manufacturer's recommendation is to be accepted and IPC does not have a preference.
- The DFN package above is for pins totally under the part, like many LGA packages. The traditional DFN we are thinking of falls into a big grouping of QFN according to IPC.

I think that covers the outstanding questions we had.

@poeschlr 
Are you able to wrap this up so it can be merged?

poeschlr:
In other words: That document calls what we call LGA as DFN and my current generators for DFN (which use QFN values) is ok. (This would be out of scope for this anyways as this is only meant to touch connector generators plus the required nodes.)

KicadModTree/nodes/specialized/PadArray.py
poeschlr: Pad arrays do not create a grid of pads. They only create one line. (Either in x or y direction or diagonal if you give both x and y spacing at the same time.) Meaning using x and y index does not really make sense.

Making it an iterator might be an option. I will look into that. (I however fear that iterators might be a bit above the skill level of the expected users of this library. Maybe i can support it in addition to the simpler function route.)

evanshultz:
Yes, that's correct. Here is the DFN footprint from the calculator tool:
image

So we are good! :)

Here are a couple more just for clarity:
image
image

Understood it's out of scope but I wanted to keep the conversation together.

Thanks for getting back to this PR.

evanshultz:
@pointhi 
Can you merge this now?

pointhi:
thanks, I will merge it now

PR #164, opened 2018-08-23
Script to generate AC/DC converter footprints.

Sthing:
Hi.

Thank you for work! It makes KiCAD better.

Here is a script to generate footprints for 3 AC/DC converters:
  Meanwell IRM-30
  Meanwell IRM-45
  Meanwell IRM-60

### DIFF
--- a/scripts/Converter_ACDC/make_converters.py
+++ b/scripts/Converter_ACDC/make_converters.py
@@ -1,0 +1,156 @@
+import sys
+import os
+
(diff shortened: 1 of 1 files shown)


Misca1234:
This can be discarded, Sthing have closed it and I have reponen the foot print push here
https://github.com/KiCad/kicad-footprints/pull/2044

by copying the foot prints directly

PR #165, opened 2018-08-23
Terminal blocks from Degson.

Sthing:
I have used these for PCBs - no problems discovered.


poeschlr:
We discovered in the footprint pull request that these are not really terminal blocks but connectors very similar to the phoenix MSTB series. It would therefore be better to extend that script to generate footprints for these parts as well.

tolot27:
To me it looks like Degson plagiarize Phoenix Contact and Wago connectors and terminal blocks.... But this shouldn't matter here much, only a in the sense introducing aliases. Does such a mechanism exists?

PR #170, opened 2018-09-05
Add JST PHD script

evanshultz:
http://www.jst-mfg.com/product/pdf/eng/ePHD.pdf

I used sorta-magic numbers in some calculations with comments explaining where they came from. I hope that's OK. Seems relatively common and at least I added attribution.

**Note: This is built off of #161 (which in turn needs #149 merged).**


pointhi:
related PR: https://github.com/KiCad/kicad-footprints/pull/915

PR #173, opened 2018-09-10
Fix TQFN-24 footprint name

evanshultz:
This is a simple one. As discussed at https://github.com/KiCad/kicad-footprints/pull/853.

### DIFF
--- a/scripts/Packages/Package_DFN_QFN/size_definitions/qfn-2x.yaml
+++ b/scripts/Packages/Package_DFN_QFN/size_definitions/qfn-2x.yaml
@@ -491,7 +491,7 @@ QFN-24-1EP_4x4mm_P0.5mm_EP2.7x2.7mm:
   #suffix: '_Pad{pad_x:.2f}x{pad_y:.2f}mm_HandSolder'
   #include_suffix_in_3dpath: 'False'
 
-TQFN-24-1EP_4x4mm_Pitch0.5mm_EP52.8x2.8mm_PullBack_ThermalVias:
+TQFN-24-1EP_4x4mm_Pitch0.5mm_EP52.8x2.8mm_PullBack:
   device_type: 'TQFN'
   #manufacturer: 'man'
   #part_number: 'mpn'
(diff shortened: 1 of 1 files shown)


poeschlr:
That part of the yaml file does not really control the footprint name. It is mainly for finding the correct entry in the yaml file (you could name it my_first_qfn_def and it would still produce the footprint with the same name as before) The footprint name is auto generated from the values given in the size definition.

To get the pull back suffix you would need to add it as the suffix definition (currently it is commented out) If this does not give you the desired result then you might either need to add a custom format string (also an option that is currently commented out) or adapt the code it self as it means the naming stuff is not powerful enough. (I have no time to test it right now. It might take till October till i find time to look into this in detail')

evanshultz:
@poeschlr 
Unfortunately the name generator isn't quite powerful enough.

Adding
```
suffix: 'Pullback'
```
does not put an underscore before this string (if the first char isn't an underscore probably one should be added) and this comes after the included `_ThermalVias` suffix. I would think the script should put this as the actual final string since the user has no control over it and it's convention to place the note about vias at the very, very end.

Adding
```
suffix: '_Pullback'
```
gives me the underscore but still has the same problem as above.

Adding
```
custom_name_format: 'TQFN-24-1EP_4x4mm_Pitch0.5mm_EP52.8x2.8mm_PullBack'
```
provides the expected footprint without vias but it doesn't give the version with vias. I don't see a rationale for this YAML option to suppress the via version.

I haven't looked into any of this, but I think a bit more power in the script would be worthwhile.

Furthermore, while I'm on the subject, there are now a ton of similar Package_* scripts. Maintaining them, like doing the above, is now a big pain. Perhaps modularization is a good idea? Put all the common stuff into a single, separate Python file from which all the modern IPC-compliant Package_* scripts draw?

poeschlr:
In the custom format simply include the thermal via key and it will be generated. It is just a format string. (It does not suppress anything. But on the other hand it only includes what you tell it to include)

poeschlr:
Additional info: if you add '{vias:s}' at the end of the custom name format you get the thermal vias stuff. See for example the footprint definitions in qfn_texas.yaml

evanshultz:
Nice. Thank you. Adding
```custom_name_format: 'TQFN-24-1EP_4x4mm_Pitch0.5mm_EP52.8x2.8mm_PullBack{vias:s}'```
does create both versions with the desired naming convention.

evanshultz:
OK. That should do it. Sorry for all the noise. Probably going to want to squash merge this one.

pointhi:
thx, I will merge now

PR #175, opened 2018-09-10
Hirose DF12C and DF12E

jneiva08:
Created script [Hirose DF12C DS](https://www.hirose.com/product/en/products/DF12/).

The most part of the script is working.
It generate all footprints of df12c ds. But when try to open with Kicad it don't recognizes.
Probably is because of naming.
I like to fix first, before create PR on kicad footprint .

The pad numbering isn't defined on datasheet.

Can someone help?

Thanks

### DIFF
--- a/scripts/Connector/Connector_Hirose/conn_hirose_df12c_ds_smd_top.py
+++ b/scripts/Connector/Connector_Hirose/conn_hirose_df12c_ds_smd_top.py
@@ -142,43 +142,45 @@ def generate_one_footprint(idx, pins, configuration):
     #    width=configuration['fab_line_width'], layer="F.Fab"))
 
     ######################## SilkS Layer ###########################
-    #poly_s_t= [
-    #    {'x': body_edge['left'] - configuration['silk_fab_offset'], 'y': body_edge['bottom'] + configuration['silk_fab_offset']},
-    #    {'x': body_edge['left'] - configuration['silk_fab_offset'], 'y': body_edge['top'] + 1 - configuration['silk_fab_offset']},
-    #    {'x': body_edge['left'] + 1 - configuration['silk_fab_offset'], 'y': body_edge['top'] - configuration['silk_fab_offset']},
-    #    {'x': body_edge['right'] - 1 + configuration['silk_fab_offset'], 'y': body_edge['top'] - configuration['silk_fab_offset']},
-    #    {'x': body_edge['right'] + configuration['silk_fab_offset'], 'y': body_edge['top'] + 1 - configuration['silk_fab_offset']},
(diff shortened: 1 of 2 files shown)


jneiva08:
The footprints weren't recognized because did have "(" ")". Fixed on last commit.

pointhi:
@jneiva08 if this is a bug in the footprint generator core this should be fixed instead. I will look into it

evanshultz:
This is a KiCad thing. See the conversation at https://github.com/KiCad/kicad-footprints/issues/932.

poeschlr:
I can not really find a reason as to why such chars should not be allowed (does not seem to be a reserved char in file systems. But some linux shells require it to be escaped -> this is a shell limitation not a file system limitation)
So we might want to raise a bugreport and ask for clarification. (I suspect this limitation will stay until v6)

jneiva08:
Exists some method to force %R to be in the center?

evanshultz:
@jneiva08 
I don't find a pin 1 mark either. Some similar parts do (https://www.molex.com/pdm_docs/sd/529910308_sd.pdf and http://suddendocs.samtec.com/prints/qsh-xxx-01-x-d-xxx-mkt.pdf and https://cdn.amphenol-icc.com/media/wysiwyg/files/drawing/10132798.pdf) but others don't (http://www.jst-mfg.com/product/pdf/eng/eJMC.pdf and https://www3.panasonic.biz/ac/e_download/control/connector/base-base/catalog/con_eng_p5k.pdf). I guess if you find a completely compatible connector with a pin 1 mark we could stick with that here for interchangeability? Otherwise, I suppose just pick whatever pin you want.

Regarding `%R`, can't it just be specifically placed? See https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/tools/footprint_text_fields.py for the fairly universal script doing this in many places in the library. (SMD RLC and connectors, for example.) Using a line-by-line approach you can see https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/Package_BGA/bga.py, starting at line 100. Does that help?

jneiva08:
@evanshultz I did find one gsm modem is MC55i-W ([datasheet](https://m2m.dk/wp-content/uploads/2012/11/mc55i-w_hid.pdf)) it use both connectors of 50 pins

Did confirm visually the connectors don't have any mark.

Later, I'll try the code to see if I can center '% R' in the middle of the footprin

poeschlr:
@pointhi the footprints have been merged. This is good to go. The refdes stuff is not important enough to hold this up and can be done later on (The benefit of scripts)

jneiva08:
Thanks

PR #176, opened 2018-09-14
Generate Bourns PTA single-gang slide potentiometers

ppelleti:
This script generates the slide potentiometers that were merged in KiCad/kicad-footprints#920.


pointhi:
thanks for your contribution

PR #183, opened 2018-09-20
Add EP center option to QFN-DFN generator

cp-aquila:
Some devices do have a slightly offset EP. For example this one: http://www.ti.com/lit/ml/mpds579b/mpds579b.pdf
This patch allows the user to set a different center for the Exposed-Pad.


pointhi:
thanks

PR #186, opened 2018-09-24
Script to generate through-hole PTC fuse footprints.

ppelleti:
This generates footprints for the Bourns MF-RG and MF-RHT series PTC fuses, which were merged in KiCad/kicad-footprints#579 and KiCad/kicad-footprints#580.


pointhi:
thanks for your contribution

Issue #154, opened 2018-07-23
Package KicadModTree


maximeborges:
It took me a few minutes to understand how to launch the scripts.
For reference, I copied the script in the root directory of the repo, and finally copied the KicadModTree/ to site-packages to be able to run any script anywhere.
I was searching for a way to properly install it, as any other module, but can't find anything.
Am I missing something or isn't there any way to correctly install this as a module at the moment ?

pointhi:
You are true, I'm missing a setup.py at the moment and calling the script is managed by some small hack (depending on the location of the script):

```python
import sys
import os

sys.path.append(os.path.join(sys.path[0], "..", "..")) # load parent path of  KicadModTree
```

I will look into it. This is a long overdue issue.

pointhi:
KicadModTree now avaliable on PyPI: https://pypi.org/project/KicadModTree/

PR #190, opened 2018-10-03
Fix import path for StandardBox

SeanDS:
I tried to run `BatteryHolder.py` and got the error:
```bash
Traceback (most recent call last):
  File "BatteryHolder.py", line 11, in <module>
    from tree.master.scripts.general.StandardBox import *
ModuleNotFoundError: No module named 'tree'
```

I think import paths must have changed elsewhere in the library because `tree` doesn't exist in the project nor as a dependency (as far as I can tell). I looked for other examples where a "general" script is used from another script to see how this is handled there, but I couldn't find any others so I fixed this by appending the "general" scripts directory to the Python path like you do with the `KicadModTree` path.

This fixes the issue for me on Python 3 (not tested Python 2), but I guess given that in #154 you packaged `KicadModTree` on PyPI and added a `setup.py` file, this path hacking should be removed entirely at some point.

### DIFF
--- a/scripts/Battery/BatteryHolder.py
+++ b/scripts/Battery/BatteryHolder.py
@@ -7,13 +7,16 @@ import re
 # load parent path of KicadModTree
 sys.path.append(os.path.join(sys.path[0], "..", ".."))
 
+# load scripts
+sys.path.append(os.path.join(sys.path[0], ".."))
+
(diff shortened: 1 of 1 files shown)


pointhi:
thanks for this fix

PR #191, opened 2018-10-04
Update potentiometer script

evanshultz:
I made these changes months ago and generated footprints at https://github.com/KiCad/kicad-footprints/pull/349. No changes since then.

I intended to provide an external interface to the script but just gave up and commented it out. It's not worth it with this script that doesn't orient the footprints the right way and has other limitations (for example items at https://github.com/KiCad/Potentiometers.pretty/pull/23). But at least these improvements can be merged and a newer script with more power and a YAML file to capture input data can be written later.

Here are the notes I took while I was updating the files:
```
- Separated SMD and THT generation scripts in accordance with library repo names
- Fixed Horizontal/Vertical confusion [https://github.com/KiCad/Potentiometers.pretty/issues/18]
- Updated footprint names to match KLC 3
- Force float division for Python 2.7 compatibility (`/2.0` instead of `/2`)
- Add F.Fab refdes [https://github.com/KiCad/Potentiometers.pretty/issues/20]
- Generated Bourns 3009P (all param info was there but no call to generate footprint function)
- Do not generate THT footprints with a hole if the datasheet doesn't show leads prepped appropriately
- Many fixes or updates to footprint names for clarity
- Do not add pin pitch in footprint name if MPN already defines this information
- Changed silk offset to 0.12mm (now totally outside of SMT parts)
- Set courtyard grid to 0.01mm
- Do not place silk under a part body
- Only show the screw drawings for a trimmer with screw if the screw height is >0
- Added code for a shaft hole on trimmer pots and remove screw drawing (function parameters existed but no supporting code)
- Added env var to 3D model path
- Removed pin number from mounting holes
- Remove `print(hbody,pinyoffset)` for Bourns 3006Y/W
- Remove pin number for through-board mounting holes
- Remove unusd `pn` variable
- Declare and assign values to slk_offset and txt_offset variables
- Add argument interface to footprint_scripts_potentiometers.py
- Greatly simplified generating footprint name, description, and tags
- Removed required "classname" and "footprint_name" parameters
- Changed "deco" parameter to "screwstyle" for vertical pots
- Changed "mount_below" parameter to "shaft_hole" function and default to False
- Changed name_additions behavior to accept a list of strings
- Changed "Double" in footprint name to "Dual"
- All footprints use proper 3D model parameters
- Add function headers to footprint_scripts_potentiometers.py
- Add additional commenting to all files (still underdocumented)

Beware: There is little to no error checking in the code. Also, there are a number of variable names that I find obscure and possibly more unused variables. Major refactoring could improve this code. And the way the name_additions works by splitting up a text string is weird. But I think this is OK for now.
```


pointhi:
thx

PR #194, opened 2018-10-05
Prevent bottom silk dot

evanshultz:
@poeschlr 
Please take a look at this. This code draws a bottom silk line with the same start and end points. It ends up placing a silk dot on the X axis under the lower F.SilkS horizontal line. There are no conditionals around this section of code so it runs for every single footprint using this script.

Why? I don't see why any gullwing parts need bottom silk. Am I missing the purpose of this?

I block commented to create the PR, but I can remove these few lines of code if it's not needed.


poeschlr:
You are right this part should not be in the script.

Just a heads up it currently generates stuff on the fab layer not silk layer (so your git commit message is not quite right)

I have no idea where it comes from. My guess would be it is a copy paste artifact.

evanshultz:
Unneeded code removed. You are right that I typed the wrong layer name.

This PR can be merged now.

PR #196, opened 2018-10-07
Possible for exclude pads, own size, spcific size

Misca1234:
I have made some modifications to the gullwing core scripts, so it will be able to make FP's like this

bild

with extra parameters in the .yaml file to control it

  ```
pad_size_x:
    minimum: 0.374
    maximum: 0.376
  pad_size_y:
    minimum: 0.49
    maximum: 0.51
  pad_distance: 0.85
  exclude_pin_list: [5]
```
exclude_pin_list:
The line exclude_pin_list: is inherit down to KicadModTree/nodes/specialized/PadArray.py
from tools/quad_dual_pad_border.py
where the pads whch number corresponds with the list in exclude_pin_list is newer added

pad_distance:
It is taken care of in ipc_gullwing_generator.py
If this is present in .yaml file then the pad distance from the x,y axis will be fixed to this distance

pad_size_x:
It is taken care of in ipc_gullwing_generator.py
If this is present the x size of the pad is constrained to the size given min and max values

pad_size_y:
It is taken care of in ipc_gullwing_generator.py
If this is present the y size of the pad is constrained to the size given min and max values

size_definitions/to_sot_smd.yam
is an example of a yaml file which produce the FP in the screen shoot

### CODE REVIEW

**Comment by evanshultz** [comment]
Being able to exclude a pin or pins is valuable and would be a nice addition.

However, the rest of the changes are simply wrong. I can only image you have some misunderstanding about the purpose of the script because the intent of the changes I read above are nonsensical.

IPC 7351 defines a specific algorithm to derive pad placement and pad size from the package dimensions. Being able to circumvent this algorithm by specifying pad placement and size means the footprints may not be (and likely will not be) IPC-compliant and _that is the whole purpose of this generator script_! I would reject any attempt to input the parameters that this script was written to create and vote to reject this PR on the basis that it violates the entire core purpose of the script.

@poeschlr 
You will definitely want to look at this and weight in.

**Comment by Misca1234** [comment]
The scripts can now be used for a much larger range of FP's which in turn means we will narrow down the number of scripts used to produce FP's

File: scripts/Packages/Package_Gullwing__QFP_SOIC_SO/ipc_gullwing_generator.py [hunk 1]
```diff
@@ -35,7 +35,7 @@ def __init__(self, configuration):
             except yaml.YAMLError as exc:
                 print(exc)
 
-    def calcPadDetails(self, device_dimensions, ipc_data, ipc_round_base):
+    def calcPadDetails(self, device_params, device_dimensions, ipc_data, ipc_round_base):
```
poeschlr: This is an unnecessary parameter here. (It holds original yaml data instead of parsed data as it should be. Meaning this function now again needs to take care of error handling.)
There must be a nicer way to achieve the same.

poeschlr:
The exclude pad list addition is something i had already planned so it should be ok to go.

Everything else is really strange. It basically removes the IPC part of the generators. I don't quite see the benefit of having this in the same generators. (After all the generator has ipc in its name. And a note about IPC compliance is added to the description of the footprints which can not be there if you do not use IPC rules to generate the footprint. -> so at the least the description generation would need to remove this note if the pads are not controlled via the ipc calculation.)

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/ipc_gullwing_generator.py
poeschlr: Maybe it would be better to not even call this function at all if the pad size is given.

Misca1234:
Well, we would have got closer to "one script make even more", I have removed all but the exclude pin

PR #197, opened 2018-10-07
Added exclude_pin_list

Misca1234:
Replacement of push 
https://github.com/pointhi/kicad-footprint-generator/pull/196
Only exclude_pin_list is included

### DIFF
--- a/KicadModTree/nodes/specialized/PadArray.py
+++ b/KicadModTree/nodes/specialized/PadArray.py
@@ -79,7 +79,7 @@ class PadArray(Node):
           pad number used for "pin 1" (default: 1)
         * *exclude_pin_list* (``int, Vector1D``) --
           which pin number should be skipped"
-          
+
     :Example:
 
     >>> from KicadModTree import *
@@ -103,7 +103,7 @@ class PadArray(Node):
         self.pincount = kwargs.get('pincount')
         if type(self.pincount) is not int or self.pincount <= 0:
             raise ValueError('{pc} is an invalid value for pincount'.format(pc=self.pincount))
-            
+
         self.exclude_pin_list = []
         if kwargs.get('exclude_pin_list'):
             self.exclude_pin_list = kwargs.get('exclude_pin_list')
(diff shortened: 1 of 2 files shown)


pointhi:
@Misca1234 please fix pep8 errors

Misca1234:
All ok, it was white spaces on empty lines

KicadModTree/nodes/specialized/PadArray.py
```diff
@@ -221,43 +231,46 @@ def _createPads(self, **kwargs):
...
-                    pads.append(
-                        ChamferedPad(
-                            number=number, at=current_pad_pos,
-                            corner_selection=kwargs.get('chamfer_corner_selection_last'),
-                            **current_pad_params
-                            ))
-                    continue
-            pads.append(Pad(number=number, at=current_pad_pos, **current_pad_params))
+            includePad = True
+            for exi in self.exclude_pin_list:
```
pointhi: equivalent should be:
```python
includePad = (i + self.initialPin) not in self.exclude_pin_list
```

Misca1234:
Change made

evanshultz:
@Misca1234 
This came into my mind for an unrelated reason but since it's closely related I'm posting it here. And furthermore, it's possible this PR will change due to the below. But maybe not.

There is a difference between excluding a pin (do not place a pin at a given location but use that pin number for the next pin location) and skipping that pin (not placing a pin at a location and moving on to the next pin number).

For example, as defined above, here is excluding the 5th pin in a generic 6 pin pad placement:
```
1    5
2
3    4
```

And here is skipping the same pin:
```
1    6
2
3    4
```

I cannot say this is the best terminology, but hopefully it's clear that there are two relevant paradigms that may be required for any footprint we may want to generate with this script. It may be prudent to at least consider if this PR would create any issues with the unimplemented paradigm. Certainly in the long run it would be worth supporting both paradigms.

pointhi:
@Misca1234 I will merge it now

PR #156, opened 2018-07-28
Added TDK-SLF Inductor Set

matyro:
See
https://github.com/KiCad/kicad-footprints/pull/352

Final footprint design still needs to be discused

PR #201, opened 2018-10-16
Added TDK SLF serie of inductors and updated StandardBox.py

Misca1234:
This push is replacing the push in 
https://github.com/pointhi/kicad-footprint-generator/pull/156

It is script based, where data is in a .yml file

This push also contain update and bug fixes for StandardBox.py
F:silk lines was drawn twice
The SMD tag was not properly set for the foot print
F.CrtYrd "goes around" right pads
smd pads are rounded rect
It is possible to customize the corners

Foot print push
https://github.com/KiCad/kicad-footprints/pull/1021

Foot print script push 
https://github.com/pointhi/kicad-footprint-generator/pull/201

bild

bild

### DIFF
--- a/scripts/Inductor_SMD/Inductor_SMD.py
+++ b/scripts/Inductor_SMD/Inductor_SMD.py
@@ -1,0 +1,62 @@
+#!/usr/bin/env python3
+
+import sys
+import os
+import re
+
+# load parent path of KicadModTree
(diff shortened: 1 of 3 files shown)


Misca1234:
@pointhi 
Could you merge this one, it includes a fix for standardbox.py that I need in other pushes

PR #205, opened 2018-10-19
Add Connector_Audio/Jack_3.5mm_Switronic_ST005_horizontal

asukiaaa:
Add a script file to make audio footprint.
https://github.com/KiCad/kicad-footprints/pull/1006

### DIFF
--- a/scripts/Connector/Connector_Audio/Jack_3.5mm_Switronic_ST-005-G_horizontal.py
+++ b/scripts/Connector/Connector_Audio/Jack_3.5mm_Switronic_ST-005-G_horizontal.py
@@ -1,0 +1,156 @@
+#!/usr/bin/env python3
+
+import sys
+import os
+
+# load parent path of KicadModTree
+sys.path.append(os.path.join(sys.path[0], "..", "..", ".."))
+
+from KicadModTree import *
+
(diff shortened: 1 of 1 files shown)


evanshultz:
@poeschlr 
Is this such a specialized connector as to not be worth scripting? Do you expect this script to be generally reusable for other parts outside of the family for which it was designed?

poeschlr:
@evanshultz To be honest i might not have invested the time to make a script for these (too little pay-off for too much effort). But i am not opposed to include such a script if it is already made.

@asukiaaa 
Your float vs integer thing (that you mentioned over on the footprint pull request) could stem from python2 vs python3. (python 3 defaults to float division while python 2 defaults to integer division. So i guess your latests runs where using python 2. A proper fix to make a script work for both python 2 and 3 would be to import division from the future library. But to be honest i would suggest not to invest time into python 2 compatibility at this point in time. Python 2 will be finally allowed to die at the end of this year. https://pythonclock.org/)

asukiaaa:
@poeschlr 
Thank you for the information about python.
I didn't know that.
I will mainly use python3.

poeschlr:
thanks

PR #208, opened 2018-10-20
Add BGA-624_25x25_21.0x21.0mm_Pitch0.8mm

tpambor:
fp


evanshultz:
I've committed https://github.com/pointhi/kicad-footprint-generator/pull/247, which allows for larger ball counts. This will conflict with one commit of this PR, but since I've implemented this in a programmatic way I dare to say my PR may be the best way forward? If so, and since there are other valuable commits in the PR, how should be reconcile this situation?

poeschlr:
The footprint is merged. This is ready to go.

pointhi:
thanks for your contribution

PR #209, opened 2018-10-21
add script for VSSOP-10_3x3mm_P0.5mm

penoud:
I have to cheat with the dimension in order to respect pins distance as the datasheet of the ADS1115 request it:
http://www.ti.com/lit/ds/symlink/ads1115.pdf


poeschlr:
Footprint is merged. This is good to go.

pointhi:
thanks for yours contribution

PR #212, opened 2018-10-25
Fix Molex PicoBlade THT part numbers and datasheet

dimhoff:
The Molex PicoBlade 53047-* and 53048-* connectors contain the pin count
in the part number. However the script for this footprint used 2 times
the pin count. Fixed this.

For the 53048-* connectors a wrong datasheet link was used.


evanshultz:
All changes look good to me. Thanks!

pointhi:
thanks for your contribution

PR #213, opened 2018-10-26
SMD_chip_package_rlc: Add 1008-scale package

florolf:
Add parameters for EIA 1008 size devices. Based off a sample of
inductors from various manufacturers (using Digikey stock as a rough
metric of popularity).

### DIFF
--- a/scripts/SMD_chip_package_rlc-etc/size_definitions/size_default_chip_devices.yaml
+++ b/scripts/SMD_chip_package_rlc-etc/size_definitions/size_default_chip_devices.yaml
@@ -24,15 +24,18 @@ SMD_0805:
     terminator_spacing_max: 1.55
     size_info: '(Body size source: https://docs.google.com/spreadsheets/d/1BsfQQcO9C6DZCsRaXUlFlo91Tg2WpOkGARC1WS5S8t0/edit?usp=sharing)'
 
-# SMD_1008:
-# I did not find any part that uses this package.
-#     code_imperial: "1008"
-#     code_metric: "2520"
-#     body_length: 2.5
-#     body_length_min:
-#     body_width: 2.0
-#     body_width_min:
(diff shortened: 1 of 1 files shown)


evanshultz:
Does this package size come in other components except for inductors?

The IPC 7351 scripts are built with a specific lead termination in mind, like the 3-side resistor (top, end, bottom) or the 5-side capacitor (add left and right sides to the earlier 3). Some of the inductors in the spreadsheet have different lead styles, like the Abracon and Panasonic. While it still may work, the pad size calculations will not be correct for other lead styles. Yet the Littelfuse, for example, has lead terminations like a chip capacitors and would probably be fine. I'm not opposed to adding this entry in the YAML file, but we may want to be more selective about the parts used so that we're only including parts which have lead terminations that are meant to be used with the script. @poeschlr ?

florolf:
> Does this package size come in other components except for inductors?

[Octopart](https://octopart.com/search?category_ids=4165&specs2.2.values=1008&start=0) suggests that while there are other components in that package, the vast majority are inductors. It might make sense to only generate it for inductors.

> Some of the inductors in the spreadsheet have different lead styles

Yeah, it definitely makes no sense to include those for the lead calculations. It might still make sense to include them for the calculation of the variation of the general package size, though.

evanshultz:
At the Octopart link you sent the resistors are a 2 element array and neither capacitor is a normal 5-side end termination. None of the EMI filter parts fit either. So yeah, just inductors I suppose. And only a subset of inductors of that size should be used with these footprints.

I wouldn't include parts with other leads styles (like the Panasonic and it's J-leads) at all since it's likely to contaminate the results. If the part isn't exactly compatible with the lead style required by the script I say leave it out.

pointhi:
please reference the corresponding kicad-footprint PR

PR #216, opened 2018-10-30
Add Atmel_TFBGA-120_8.0x8.0mm_Layout15x15_P0.5mm

cp-aquila:
http://ww1.microchip.com/downloads/en/DeviceDoc/SAM-D5xE5x-Family-Datasheet-01507B.pdf#page=2095


pointhi:
please reference the PR in kicad-footprint

cp-aquila:
Sorry for the long silence.
In the end, I did not use that footprint for my project. I can still try to get it merged, but I am not sure if its worth the effort. I have a symbol for the related CPU, but its not up to KLC standards yet. Also this BGA package is really hard to route/use, it needs via-in-pad with plugged vias to work properly.

So my suggestion is to just scrap this PR and close it.
What do you think?

pointhi:
@cp-aquila it's your choice how to continue

PR #193, opened 2018-10-05
Add TPS703xx PWP entry

evanshultz:
From http://www.ti.com/lit/ds/symlink/tps703.pdf. Footprints submitted at https://github.com/KiCad/kicad-footprints/pull/995.

PR #217, opened 2018-11-06
Add TPS703xx PWP entry

evanshultz:
Replacement for https://github.com/pointhi/kicad-footprint-generator/pull/193. From http://www.ti.com/lit/ds/symlink/tps703.pdf. Footprints merged at KiCad/kicad-footprints#995.


pointhi:
thanks for your contribution

PR #218, opened 2018-11-10
Add TI DSBGA (YZR) footprint

RockyTV:
http://www.ti.com/lit/ml/mxbg270/mxbg270.pdf

I'm not sure if the dimensions I used are correct. I've calculated the mean for the body size (max 1.55mm, min 1.489mm as per LMV7272 datasheet). I don't know where I can find the mask margin nor the pad diameter (I suppose it's the rightmost image in the datasheet, because the other one is land pattern recommendation).

Required by https://github.com/KiCad/kicad-symbols/pull/824

### DIFF
--- a/scripts/Package_BGA/bga.yml
+++ b/scripts/Package_BGA/bga.yml
@@ -41,10 +41,11 @@ Texas_DSBGA-8_1.5195x1.5195mm_Layout3x3_P0.5mm:
   pkg_width: 1.5195
   pkg_height: 1.5195
   pitch: 0.5
-  pad_diameter: 0.320
+  pad_diameter: 0.2625
   layout_x: 3
   layout_y: 3
   mask_margin: 0.05
+  row_skips: [[], [2], []]
 Texas_MicroStar_Junior_BGA-113_7.0x7.0mm_Layout12x12_P0.5mm:
   description: "Texas Instruments, BGA Microstar Junior, 7x7mm, 113 ball 12x12 grid, NSMD pad definition (http://www.ti.com/lit/ml/mpbg674/mpbg674.pdf, http://www.ti.com/lit/wp/ssyz015b/ssyz015b.pdf)"
   pkg_width: 7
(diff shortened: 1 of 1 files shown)


evanshultz:
Also include http://www.ti.com/lit/ds/symlink/lmv7272.pdf in the documentation since that will be required to know the package dimensions (D and E).

The pad diameter should be 0.2625 (middle of the range shown.

You can also skip the center pad using (check this because I didn't try it myself and it might be off-by-one): `row_skips: [[], [2], []]`.

pointhi:
please link the referenced PR for the footprint

evanshultz:
No footprint PR was submitted.

@RockyTV 
Please make the changes above and submit the footprint so we can wrap up your PRs. Thanks!

RockyTV:
I am extremely sorry (again!) for leaving the PR behind. Good news: using the script is easy and this is the generated package:



I don't know if the body size should be nominal, as specified in the IC's datasheet, or the mean value between min and max in the package datasheet.

evanshultz:
I only see min and max values in the comparator datasheet. You can average those as described at https://github.com/KiCad/kicad-symbols/pull/824#issuecomment-427592086. Where do you see other dimensions?

Please post the footprint in a PR at https://github.com/KiCad/kicad-footprints, and reference this PR.

Thank you!

RockyTV:
> I only see min and max values in the comparator datasheet. You can average those as described at [KiCad/kicad-symbols#824 (comment)](https://github.com/KiCad/kicad-symbols/pull/824#issuecomment-427592086). Where do you see other dimensions?

In the LMV727x datasheet, the nominal body size is 1.50mm x 1.50mm, but in a page that includes the package's dimensions, it states min and max values for the body size as well:



What I did was the average between those two, which resulted in 1.5195mm.

evanshultz:
Footprint is merged.

pointhi:
thanks for your contribution

PR #222, opened 2018-11-20
Make Conn_PinSocket script run again

matthijskooijman:
This fixes the Conn_PinSocket script and makes it run. It seems that one file (parameters.py) was forgotten in #82 by @terjeio. I found a version by him in the 3D model generator for these footprints, which worked with minimal changes.

I also made some changes for python3 support, now the script runs in Python2 and Python3 and produces identical results.

The footprints are also identical to the ones present in the official library, except that those have oval pads (with identical width and height) and this script now generates circle pads (due to bf378d72ff23364984763e78c2343a016b44d166).


pointhi:
thanks for your fix

PR #224, opened 2018-11-25
Script for Harwin M20-781xx45 connectors

jneiva08:
Script for Harwin M20-781xx45

Footprint PR [#1107](https://github.com/KiCad/kicad-footprints/pull/1107)

### DIFF
--- a/scripts/Connector/Connector_Harwin/conn_harwin_m20-781xx45_smd_top_dual_row.py
+++ b/scripts/Connector/Connector_Harwin/conn_harwin_m20-781xx45_smd_top_dual_row.py
@@ -1,0 +1,211 @@
+#!/usr/bin/env python3
+'''
+kicad-footprint-generator is free software: you can redistribute it and/or
+modify it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+kicad-footprint-generator is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+You should have received a copy of the GNU General Public License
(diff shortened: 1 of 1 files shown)


jneiva08:
@pointhi, how is the best way to remove the series name?

Now it is have Harwin_M20_M20-7810245, but SchrodingersGat recommended to remove the first M20.
To be fast i have removed `{series:s}_` in `fp_name_format_string` from conn_config_KLCv3.yaml

didn't commit the changes, i like to know how can proceed, thanks

evanshultz:
@SchrodingersGat @poeschlr 
These fall into the "Manufacturer Specific Connectors" section of http://kicad-pcb.org/libraries/klc/F3.6/, I believe, where `Series` comes before `MPN`. Neither field is optional. The first "M20" is for the `Series` and the second one was part of the `MPN` so it just looked weird but was following convention.

None of the other connectors in the Harwin repo have a datasheet or appear to be relatively modern so they don't make a good example, but let's look at the JST repo where I know many of them are recent. The have doubled text to follow KLC:
image

That does look a bit bizarre and does not add clarity to me.

This convention, however, works well for Molex (https://github.com/KiCad/kicad-footprints/tree/master/Connector_Molex.pretty) where you have Series names like "CLIK-Mate" and "Micro-Fit_3.0" which are quite distinct from the MPN.

The Samtec repo (https://github.com/KiCad/kicad-footprints/tree/master/Connector_Samtec.pretty) doesn't double text in the scripted LSTM footprint names and I think it's clear.

Should KLC state that the Series field is required only if it is not a substring of the MPN? This would require updating a number of footprint file names if we change existing ones, but we could start a new precedent (perhaps updating all for v6?) to avoid doubled text. I agree that removing one "M20" looks better and doesn't lose anything, but since we're noticing this inconsistency now let's sort it out.

pointhi:
thanks for your contribution

jneiva08:
Thanks for the merge.

Now as it is will generate as Harwin_M20_M20-7810245. When did make the question was to find the best option to have Harwin_M20-7810245.

PR #226, opened 2018-11-25
New foot prints for Valves

Misca1234:
Overhaul of the valve foot prints

All foot prints are script made (except CK6418)
All 3D models are script made

The models and foot print does not replace existing valve foot prints.
Removal of old foot prints will be done with a later push of thesymbols

`NOTE: this push include a fix in KicadModTree/nodes/base/Model.py and KicadModTree/KicadFileHandler.py which make it possible to use offset and at when setting the 3D  model cordinate`

Foot print push

Foot print script push

3D model push
https://github.com/KiCad/kicad-packages3D/pull/466

3D model scripts

bild

bild

bild

bild

bild

### DIFF
--- a/scripts/Valves/valve.py
+++ b/scripts/Valves/valve.py
@@ -1,0 +1,145 @@
+#!/usr/bin/env python3
+
+import math
+import os
+import sys
+
+# load parent path of KicadModTree
+sys.path.append(os.path.join(sys.path[0], "..", ".."))
+
+from KicadModTree import *
+
+def bga(args):
+    footprint_name = args["name"]
+    description = args["pkg_description"]
+    datasheet = args["pkg_datasheet"]
+    D = args["pkg_D"]
+    pkg_valve_offset = args["pkg_valve_offset"]
+    pkg_socket_offset = args["pkg_socket_offset"]
+    npth_pin = args["pkg_npth_pin"]
+    center_pin = args["pkg_center_pin"]
+    pin_type = args["pkg_pin_type"]
+    pin_number = args["pkg_pin_number"]
+    pin_arc = args["pkg_pin_arc"]
+    pin_diameter = args["pkg_pin_diameter"]
+    npth_pin = args["pkg_npth_pin"]
+    socket = args["pkg_socket"]
+    tube = args["pkg_tube"]
+
+    socket
+    
+    alpha_delta = 0 - ((pin_arc * math.pi) / 180.0)
+    h = pin_diameter / 2.0
+    origo_dx = (h * math.sin(alpha_delta))
+    origo_dy = (h * math.cos(alpha_delta))
+    
+    origo_x = 0 - origo_dx
+    origo_y = origo_dy
+    
+    s1 = [1.0, 1.0]
+    s2 = [1.0, 1.0]
+    
+    t1 = 0.15 * s1[0]
+    t2 = 0.15 * s2[0]
+
+    yRef = origo_y - (D / 2.0) - (s1[1] * 1.2)
+    yValue = origo_y + (D / 2.0) + (s1[1] * 1.2)
+
+    wFab = 0.10
+    wCrtYd = 0.05
+    wSilkS = 0.12
+    
+    description_tot = description + ', Made by script https://github.com/pointhi/kicad-footprint-generator/scripts/Valves, ' + datasheet
+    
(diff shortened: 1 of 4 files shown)


scripts/Valves/valve.py
```diff
@@ -0,0 +1,448 @@
...
+    xL2n = sadle_x * math.cos(math.radians(sadle_a)) - (0.0 * math.sin(math.radians(sadle_a)))
+    yL2n = sadle_x * math.sin(math.radians(sadle_a)) + (0.0 * math.cos(math.radians(sadle_a)))
+    #
+    #
+    #
+    x9n = x9 * math.cos(math.radians(sadle_a)) - (y9 * math.sin(math.radians(sadle_a)))
+    y9n = x9 * math.sin(math.radians(sadle_a)) + (y9 * math.cos(math.radians(sadle_a)))
+    #
+    x10n = x10 * math.cos(math.radians(sadle_a)) - (y10 * math.sin(math.radians(sadle_a)))
+    y10n = x10 * math.sin(math.radians(sadle_a)) + (y10 * math.cos(math.radians(sadle_a)))
```
poeschlr: all of these rotations would be a lot simpler if using an array and either a rotate function or maybe even better merge https://github.com/pointhi/kicad-footprint-generator/pull/351 into your branch as it defines a rotation function for Vector2D.

So the options are:
```python
# somewhere near the top of the file
def rotate(x, y, angle):
  xo = x * math.cos(math.radians(angle)) - (y * math.sin(math.radians(angle)))
  yo = x * math.sin(math.radians(angle)) + (y * math.cos(math.radians(angle)))
  return xo, yo

# coords_0 are Vector2D
coords = []
for v in coords_0:
  coords.append(Vector2D(rotate(v.x, v.y, sadle_a)))
```
---

Or if you merged my stuff then:
```python
coords = []
for v in coords_0:
  coords.append(Vector2D(v).rotate(sadle_a))
```
or if you do not need the orriginal coordinates then it could look like this:
```python
for i in range(len(coords)):
  coords[i].rotate(sadle_a)
```

scripts/Valves/valve.py
```diff
@@ -0,0 +1,448 @@
...
+    #
+    # reflect small circle in the large circles origo
+    #
+    #
+    # reflect large circle in the large circles origo
+    #
+    x10 = x7
+    y10 = 0.0 + (0.0 - y7)
+    x9 = x8
+    y9 = y3
```
poeschlr: To have an array for below you could do the following:
`coords=[None]*10`
and then instead of writing `x9 = x8`, `y9 = y3` you would write `coords[8] = Vector2D(coords[7].x, coords[2].y)`

(One does not simply start to count with x1,y1 it is x0,y0 :-) )

evanshultz:
Is `sadle` supposed to be `saddle`?

scripts/Valves/valve.py
```diff
@@ -0,0 +1,448 @@
...
+    npth_pin = args["pkg_npth_pin"]
+    center_pin = args["pkg_center_pin"]
+    pin_type = args["pkg_pin_type"]
+    pin_number = args["pkg_pin_number"]
+    pin_arc = args["pkg_pin_arc"]
+    pin_diameter = args["pkg_pin_diameter"]
+    socket = args["pkg_socket"]
+    tube = args["pkg_tube"]
+
+    socket
```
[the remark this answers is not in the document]
Misca1234: If I do not remember wrong it is becaouse it have 2 3D models, the socket is the "green one" and tube is the glas "dome" above

Misca1234:
"Is sadle supposed to be saddle?"

jee, it must be a typo

scripts/Valves/valve.py
Misca1234: "(One does not simply start to count with x1,y1 it is x0,y0 :-)"

ohhh, you should have been around the time when printers almost could not distingues between 0 and O :)

Misca1234:
@poeschlr 

You gave a lot of feedback of improvements of the readable and also a tip of merging
https://github.com/pointhi/kicad-footprint-generator/pull/351

If we all are alright I would like to avoid to change the scripts at the moment unless you want me to look into the 
https://github.com/KiCad/kicad-footprints/pull/1139#issuecomment-498418821 ?

PR #233, opened 2018-12-07
3M Textool ZIF sockets

ppelleti:
This is the script that generates the 3M Textool ZIF sockets in KiCad/kicad-footprints#495.


poeschlr:
The footprints have been merged. This should be good to go.

PR #236, opened 2018-12-08
Added Altech AK300 serie

Misca1234:
This the push of Altech AK300 serie
http://www.altechcorp.com/PDFS/PCBMETRC.PDF

All foot prints and 3D models are script made

Foot print push
https://github.com/KiCad/kicad-footprints/pull/1155

Foot print script push 
https://github.com/pointhi/kicad-footprint-generator/pull/236

3D model push
https://github.com/KiCad/kicad-packages3D/pull/478

3D model script push
https://github.com/easyw/kicad-3d-models-in-freecad/pull/242

Parameter file foot prints
```
AK300:
  description: "Altech AK300 serie connector"
  datasheet: "http://www.altechcorp.com/PDFS/PCBMETRC.PDF"
  tags: "Altech AK300 serie connector"
  manufacture: 'Altech' # Model name
  serie: 'AK300'        # Model name
  W: 12.5               # Package width
  H: 12.5               # Package height
  WD: 6.5               # > Y distance form pin center to package edge
  A1: 0.1               # package board seperation
  pin_number: [ 2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24]  # Which pin configuration
  PE: 2.5               # Distance from edge to pin
  PS: 5.0               # Distance between pins
  PD: [1.0, 0.8]        # Pin size, [1.0] diameter 1 mm, [1.0, 0.8] rectangle 1.0x0.8
  PL: 4.5               # Pin length
  PF: 'rect'            # Pin form 'round' or 'rect'
  SW: 2.7               # Blender width
  DD: 1.5               # Drill diameter
  PD: 3.0               # Pad diameter
  rotation: 0           # Rotation if required
  dest_dir_3D_prefix: 'Connector_Altech.3dshapes'  # Destination directory

```

Parameter file 3D model

```
    'AK300': Params(   # ModelName
        #
        #
        #
        manufacture = 'Altech', # Model name
        serie = 'AK300',        # Model name
        W  = 12.5,              # Package width
        H  = 12.5,              # Package height
        WD  = 6.5,              # > Y distance form pin center to package edge
        A1 = 0.1,               # package board seperation
        pin_number = [ 2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24],  # Which pin configuration
        PE = 2.5,               # Distance from edge to pin
        PS = 5.0,               # Distance between pins
        PD = [1.0, 0.8],        # Pin size, [1.0] diameter 1 mm, [1.0, 0.8] rectangle 1.0x0.8
        PL = 4.5,               # Pin length
        PF = 'rect',            # Pin form 'round' or 'rect'
        SW = 2.7,               # Blender width
        rotation = 0,           # Rotation if required
        body_color_key  = 'black body',         # Body color
        pin_color_key   = 'metal grey pins',    # Pin color
        dest_dir_prefix    = 'Connector_Altech.3dshapes'  # Destination directory
        ),

```

bild

bild

bild

bild

bild

bild


tolot27:
There is a small slider with a size of 0.6 mm to the right of the connector which is used to extend it to more "poles" (see picture). It looks like it is not included in the size calculation.
image.

Two other questions:
1. Are the variable names in the yaml somehow standardized?
2. Regarding to `WD # > Y distance form pin center to package edge`: Is it common to specify it from an edge rather than from the center of the part?

PS: I'd like to reuse this great script for Phoenix Contact PTDA.

tolot27:
And more questions? Which is the center of the StandardBox? In `pins.append` the y coordinate is zero. Does it mean it the pins are append in the center?

Misca1234:
> There is a small slider with a size of 0.6 mm to the right of the connector which is used to extend it 
> to more "poles" (see picture). It looks like it is not included in the size calculation.

No, it was not included, the 3D models is not super exactly as the real object, mostly lacking of accurate available drawings (some measurements are not included) and every more tiny detail also include more work.
Do you feel that the detail should be included ?



> And more questions? Where is the (rotation) center of the Standard Box? Is it calculated automatically
>  based on the bounding rectangle?
> In pins.append the y coordinate is zero. Does it mean it the pins are append in the center?

No, the standard box works like this
1. It only create rectangular foot prints, with unlimited pin, smd and npth pins, it does not handle other shapes, it's purpose is to make it easy to create most of the foot rpints, that is rectangle boxes with some pins.
2. You specify the length and width of the device
3. You specify where the pins are located (in real coordinates)
4. you specify where the upper left corner of the box is in relationship to pin 1

Misca1234:
> PS: I'd like to reuse this great script for Phoenix Contact PTDA.

No problem, however it have a bug where it does not setr the width of the courtyard line correctly, 
for this push I used the update version.
If you wait for this push to be merged you can use it without trouble
https://github.com/pointhi/kicad-footprint-generator/pull/201

Misca1234:
One more thing, you can place pads, pins and npth holes anywhere.
The script will draw the Fab line as it should be 
The silk and and the courtyard are drawn just out side , 0.13 and 0,25 mm distance but these lines will "crawl" around the Fab line and jump over pins/smd/npth holes so Standardbox.py will also work for those foot prints where a hole/pad is on the "edge" of the Fab line.

tolot27:
> > There is a small slider with a size of 0.6 mm to the right of the connector which is used to extend it
> > to more "poles" (see picture). It looks like it is not included in the size calculation.
> 
> No, it was not included, the 3D models is not super exactly as the real object, mostly lacking of accurate available drawings (some measurements are not included) and every more tiny detail also include more work.
> Do you feel that the detail should be included ?

No, it is not important. I did not focus on the 3D model. I just looked at the footprint. In general, the bounding box should cover the maximal size. In this case, only 0.6 mm have to be considered, which is not worth of the extra work.

> > And more questions? Where is the (rotation) center of the Standard Box? Is it calculated automatically
> > based on the bounding rectangle?
> > In pins.append the y coordinate is zero. Does it mean it the pins are append in the center?
> 
> No, the standard box works like this

Hence, the center is computed later by KiCad? In Eagle (where I came from) I had to take the center or coordinate origin into account. Otherwise, the part could not be rotated easily. But it looks like I don't have to take care of it in KiCad footprints.

Misca1234:
I see that Thomas have not merged the update of the Standardbox yet, I think he is on travling feet
If you want you can update the script yourself, there are some other fixes included in the push I am waiting for but for you the problem lies in when adding the F.CrtYrd lines, it uses the wrong line width, fairly easy fix

Misca1234:
@tolot27 
The merge of standardbox is merged now, please go ahead and use it

Misca1234:
Change file name ending and directory according to request in
https://github.com/KiCad/kicad-footprints/pull/1155

poeschlr:
@pointhi the footprints are merged, this is now good to go.

pointhi:
thanks for your contribution

PR #239, opened 2018-12-14
Added definition for QFN-16

SchrodingersGat:
Added a new QFN definition:

Example Part: [TSV524IQ4T](https://www.st.com/content/ccc/resource/technical/document/datasheet/4a/50/94/16/69/af/4b/58/DM00047334.pdf/files/DM00047334.pdf/jcr:content/translations/en.DM00047334.pdf)


poeschlr:
The footprint is merged. Meaning this is good to go.

PR #243, opened 2018-12-15
Keep out circle

Misca1234:
This little add on will create a "keep out circle" like this

bild

It is created like a normal circle

```

    :Keyword Arguments:
        * *center* (``Vector2D``) --
          center of the circle
        * *radius* (``float``) --
          radius of the circle
        * *layer* (``str``) --
          layer on which the circle is drawn (default: 'F.SilkS')
        * *width* (``float``) --
          width of the circle line (default: None, which means auto detection)
        * *step* (``float``) --
          the step betwene the crossing lines (default: 1 mm)
```

With the extra parameter 'step' which define the step along the x-axis for the intersection lines


Misca1234:
ping

PR #244, opened 2018-12-17
Add size definition for TI S-PWSON-N8

poeschlr:
normal qfn ipc definition is used as it gives the nearest result to the
suggestion in the datasheet. (exact fit not possible as the datasheet
suggestion uses an out of date standard.)

Datasheet link: http://www.ti.com/lit/ds/symlink/lp2951.pdf


poeschlr:
Included this into #250

PR #245, opened 2018-12-17
Adding size definitions for qfp footprints

poeschlr:
This PR will be used to add missing qfp size definitions to generate more of the qfp footprints that are already in the lib.
Will be split into multiple sections to make my life and the life of the reviewing librarian easier. (one commit here will mean one pull request on the official repo.)


evanshultz:
Since this is the one place that touched all footprint PRs, I'll ask here: Did you envision any batching for the review or shall this be done ad hoc?

poeschlr:
I made sure that there is always one commit for every pull request. 
Additionally i added the respective lines of the yaml definition files over at every pull request. (so it should be easy to review)

If we find mistakes then they will simply be fixed in the usual way. (adding a new commit for every fix necessary.)

This PR will add all LQFP footprints. (two are missing at the time of writing. Should be done within an hour or so)
If i start another series than it will be done in a separate pull request.

poeschlr:
Ok every LQFP footprint currently in the lib is now transferred to the script. (I think i need a break now.)

evanshultz:
All footprints are now merged so this is ready.

poeschlr:
Closing this as it is included in https://github.com/pointhi/kicad-footprint-generator/pull/259

PR #246, opened 2018-12-17
Add entries for 3 QFNs and a WLCSP

evanshultz:
A few footprints used in several PRs.

QFN16 and QFN24:
https://github.com/KiCad/kicad-symbols/pull/1064
http://www.cypress.com/file/46236/download
image

LFCSP:
https://github.com/KiCad/kicad-symbols/pull/1180
https://www.analog.com/media/en/technical-documentation/data-sheets/AD7682_7689.pdf
image

WLCSP:
https://github.com/KiCad/kicad-symbols/pull/1189#issuecomment-443025084
http://ww1.microchip.com/downloads/en/devicedoc/atmel-8235-8-bit-avr-microcontroller-attiny20_datasheet.pdf#page=208
image

This PR relies on an update to bga.py which I'll submit soon.

PR #247, opened 2018-12-18
BGA script update

evanshultz:
This update is required for https://github.com/pointhi/kicad-footprint-generator/pull/246.

My objective was to support unique X and Y pitches, but I went further. There are two commits:
1. Adding the basic functionality I needed.
1. Performing commonalization with the other Package_ scripts by Rene to a large extent.

Broken down by feature changes, the changes are:
- Support unique X and Y pitch
- Stream in YAML input files
- Use common global config files
- Build configurable and bigger list of row names

The last item above includes https://github.com/pointhi/kicad-footprint-generator/pull/238 and one commit of https://github.com/pointhi/kicad-footprint-generator/pull/208 (specifically https://github.com/pointhi/kicad-footprint-generator/pull/208/commits/2cd8f9ac0d43d2be3c31cb6e7f37fdc27b64e0e1) and one commit of https://github.com/pointhi/kicad-footprint-generator/pull/240 (specifically https://github.com/pointhi/kicad-footprint-generator/pull/240/commits/4bbbeed238ca87483c777ba387ddf7ae8871bb57), but it does it in a programmatic way that is easier to maintain. This could be split out into a separate helper function, but I didn't see that it applied to other footprints types so I kept it in the existing `bga.py` script.

There are a few remaining things that could further align the Package_* scripts, which are not done here. In somes cases this is simply because of outstanding PRs, in other cases I figured we should talk about it first (and it might be blocked by a PR anyway). Any of the below items should wait for a future PR.
1. Push this folder down a level to the Package folder.
1. Rename the external data files from YML to YAML extension.
1. Use a series configuration file. (IMO it's OK as it is but it could be more flexible.)
1. Allow ball size to be entered and derive padstack design from IPC rules.


evanshultz:
@Ratfink 
I know you're busy, but since you originated this script I'd be happy to hear any comments.

@poeschlr 
This was the only Package_* script not written by you. Is this an acceptable alignment to what I would expect is now our in-house "standard"? Is there something you'd like done if we're going to update this script to support external config files and stream YAML input data?

tpambor:
Sorry for the inconvenience I expected it to be kept open until this one landed. You should be able to resolve the conflict by merging master into your branch and resolve the conflict by keeping your version.

evanshultz:
@pointhi 
Should be good now. Sorry that I was a bit naive in resolving this, so you may want to squash merge.

pointhi:
@poeschlr can I merge this?

PR #249, opened 2018-12-20
Fix for rounded chamfered pad point calculation

poeschlr:
The calculation for the "inner" points ignored that there is an angle to
take care of. This fix implements a correction for 45 degree chamfered
pads plus an exception for non 45 degree ones (not implemented)


poeschlr:
Travis complains about a line length inside of a multi line string representing a generated footprint. Do you have any idea how to suppress this error?

pointhi:
try add ```` # NOQA: E501````

poeschlr:
This did the trick. Thanks.

pointhi:
thanks. Patch looks fine

PR #252, opened 2018-12-23
XP Power dc dc converters

ferdymercury:
https://github.com/KiCad/kicad-footprints/pull/1058

First attempt.


Misca1234:
You can do it even easier, if you look into 
kicad-footprint-generator/scripts/Battery/BatteryHolder.py
kicad-footprint-generator/scripts/Battery/BatteryHolder.yml
You can see how the battery holders are scripted, they are defined in a .yml file and uses the standardbox.py class to create, essential rectangular foot print, so you code will be like 4 lines

poeschlr:
Footprints have been merged. This should be good to go.

PR #254, opened 2018-12-26
Removed rounded corners on TDK inductor

Misca1234:
Removed the rounded corners of the TDK inductors

The corners are now 90 degree

bild

bild


Misca1234:
> Why remove the rounded corners?

It was asked for it in this push
https://github.com/KiCad/kicad-footprints/pull/1021

> I would prefer removing silk on the pin 1 side since the extra silk outside the part body could easily 
> get clipped off by nearby components. This is subjective, but something to consider.

Yes, that could be an improvement to be done

poeschlr:
thanks.

PR #256, opened 2018-12-30
Correct JST MPNs

evanshultz:
Fixes #911

This all started from https://github.com/KiCad/kicad-footprints/issues/1141 and I documented the first couple issues at https://github.com/KiCad/kicad-footprints/issues/911. I've now looked at all JST connector footprints and fixed only the errors in MPN and make other non-breaking updates (fixed keywords, etc.).

General single-row connectors:
- Changed "side entry" to "horizontal" and "top entry" to "vertical" in keywords

ACH (http://www.jst-mfg.com/product/pdf/eng/eACH.pdf):
- MPN started with "BBM" and should be "BM"

EH (http://www.jst-mfg.com/product/pdf/eng/eEH.pdf):
- When the pin count is <10 there is no leading zero in the MPN
- There is no 16 pin version

J2100 (http://www.jst-mfg.com/product/pdf/eng/eJFA-J2000.pdf, page 7):
- Keying was not in the footprint name; this is unclear (and affects 3D models) so could somebody check me on this?
- Changed pin numbers a1 through b5 to 1 though 10 (the "Line No" in the datasheet is not related to the pin number)

PH (http://www.jst-mfg.com/product/pdf/eng/ePH.pdf):
- Abstracts away the exact PN for normal vs low insertion force THT headers, but I left it alone

SFH (http://www.jst-mfg.com/product/pdf/eng/eSFH.pdf):
- Corrected MPN

XH (http://www.jst-mfg.com/product/pdf/eng/eXH.pdf):
- When the pin count is <10 there is no leading zero in the MPN
- Removed JST_XH_B11B-XH-AM_1x11_P2.50mm_Vertical (doesn't exist)

ZE (http://www.jst-mfg.com/product/pdf/eng/eZE.pdf):
- For SMT, the top entry was massing the final "T" and the side entry had a final "T" but shouldn't have

The following appear to be OK to be in footprint naming, so just keyword updates: AUH, GH, JWPF, LEA, NV, PH, SHL, SH, SUR, VH, XAG

I used Python 3.7.0.

@poeschlr
Would you be willing to correct the EH and XH 3D models? Those are the only ones at https://github.com/KiCad/kicad-packages3D/tree/master/Connector_JST.3dshapes that are affected and it looks like you submitted the existing files. Since you're not busy with anything else I hope that's OK. :)

### DIFF
--- a/scripts/Connector/Connector_JST/conn_jst_J2100_tht_side.py
+++ b/scripts/Connector/Connector_JST/conn_jst_J2100_tht_side.py
@@ -37,11 +37,12 @@ datasheet = 'http://www.jst-mfg.com/product/pdf/eng/eJFA-J2000.pdf'
 
 pitch = 2.50
 
-pin_range = [3,4,5,6,8,10] #number of pins in each row
+#pin_range = [3,4,5,6,8,10] #number of pins in each row
+pin_range = [6, 8, 10, 12, 16, 20]
 row_pitch = 2.50
 
 #FP name strings
-part_base = "S{n:02}B-J21DK-GG" #JST part number format string
+part_base = "S{n:02}B-J21DK-GG{s}R" #JST part number format string
 
 drill = 0.86 # 0.85 -0.03/+0.05 -> 0.86 +/-0.04
 mh_drill = 2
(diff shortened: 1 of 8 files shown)


evanshultz:
Footprints are pushed at https://github.com/KiCad/kicad-footprints/pull/1238.

PR #258, opened 2019-01-05
Integrate lga in dfn qfn generator

poeschlr:
This removes duplication as the lga and qfn/dfn generators are very similar. I only needed to add a few functionalities to the dfn/qfn generator to get the full feature set of the lga generator rebuild.


poeschlr:
@pointhi the footprints are merged. This should be good to go as well.

PR #265, opened 2019-01-11
Add DFN-8 with 6x5 body, 1.27 pitch, 3.4x4.3 ep

matthuszagh:
actually, this looks like it places the silkscreen pin 1 marker at the wrong pin. Possible bug?


matthuszagh:
Wrong footprint for the symbol I wanted. Don't see any reason to keep the PR.

PR #268, opened 2019-01-12
add size definition for Mini-Circuits XX112 SO package

lorem-ipsum:
started the sop.yaml. Matching footprint PR can be found [here](https://github.com/KiCad/kicad-footprints/pull/1062)


scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/sop.yaml
```diff
@@ -0,0 +1,49 @@
...
+    tolerance: 0.01
+  pitch: 1.27
+  num_pins_x: 0
+  num_pins_y: 4
+
+  EP_size_x:
+    nominal: 4.57
+    tolerance: 0.01
+  EP_size_y:
+    nominal: 4.45
```
poeschlr: Where does this come from? Parameter R is 4.57 in the datasheet. (Sadly there is no dimension given for the metal pad on the package it self. Only a suggested pad size.)

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/sop.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+  thermal_vias:
+    count: [2, 3]
+    drill: 0.51
+    # min_annular_ring: 0.15
+    paste_via_clearance: 0.1
+    EP_num_paste_pads: [2, 2]
+    # paste_between_vias: 1
+    # paste_rings_outside: 1
+    EP_paste_coverage: 0.65
```
poeschlr: You might want to increase this a bit as you have paste overlapping vias. (I would suggest 0.7 to counteract solder loss by wicking)

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/sop.yaml
```diff
@@ -0,0 +1,49 @@
+FileHeader:
+  library_Suffix: 'SO'
+  device_type: 'SOP'
+
+SOP-8-1EP_4.57x4.57mm_P1.27mm_EP4.57x4.45mm:
+  size_source: 'https://ww2.minicircuits.com/case_style/XX112.pdf'
+  body_size_x:
+    nominal: 4.57
+    tolerance: 0.01
```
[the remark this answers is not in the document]
poeschlr: I think the PL number takes the imperial number as the reference. (It would not make sense that the tolerance changes depending on which measurement system one uses.)

Meaning i am really convinced that 0.127 is the correct tolerance.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/sop.yaml
poeschlr: I am assuming that this dimension comes from the fact that the EP looks like ending at the same line as the pads. Meaning 3*1.27+0.64 = 4.45. Could you add a note in the yaml file about that?
(Something like 

``` yaml
nominal: 4.45     # not directly given in datasheet. EP edge coincides with pads. (3*1.27+0.64 = 4.45)
```

lorem-ipsum: ah, now I understand what you were pointing at exactly. Yes, that's there I got this number from. I also added your comment to make it clear.

lorem-ipsum:
Is the script ok now or is there anything left to do? I pulled the latest master branch, now it should be mergable again.

poeschlr:
@pointhi the footprint has been merged. This is good to go.

PR #271, opened 2019-01-18
add QFN-56-1EP_8x8mm_P0.5mm_EP6.1x6.1mm

penoud:
footprint PR https://github.com/KiCad/kicad-footprints/pull/1294

### DIFF
--- a/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/qfn/qfn-5x.yaml
+++ b/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/qfn/qfn-5x.yaml
@@ -226,3 +226,40 @@ QFN-56-1EP_8x8mm_P0.5mm_EP5.6x5.6mm:
   #pad_length_addition: 0.5
   #suffix: '_Pad{pad_x:.2f}x{pad_y:.2f}mm_HandSolder'
   #include_suffix_in_3dpath: 'False'
+QFN-56-1EP_8x8mm_P0.5mm_EP6.1x6.1mm:
+  device_type: 'QFN'
+  size_source: 'http://intantech.com/files/Intan_RHD2000_series_datasheet.pdf'
+  ipc_class: 'qfn' # 'qfn_pull_back'
+  #ipc_density: 'least' #overwrite global value for this device.
+  # custom_name_format:
+  body_size_x:
+    minimum: 8.0
+    maximum: 8.0
+  body_size_y:
+    minimum: 8.0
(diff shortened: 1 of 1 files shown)


poeschlr:
The datasheet only gives nominal values for all dimensions (I really dislike that to be honest.)
I am checking if there is a JEDEC standard that we can use as a replacement for this parts drawing. (Currently i am waiting for the automated mail that will give me access.)

poeschlr:
Sadly it seems i will not get an account at jedec (it might be that one would need a company mail address and state the same company in the application. I am not really prepared to do this.)

---

And another thing: I have noticed that body_height is not the correct name for what that parameter. I changed it to overall_height in my other parts already.

evanshultz:
All the dimensions are there, I think, so it's only the tolerances that are missing. Right?

JEDEC MO-220 variation VLLD-2 may describe this package. I just use their package drawings as needed, so I'm not an expert, but it looks correct to me. If I Google for this string in a PDF file I am able to come up with the spec. Take a look and see if it seems suitable.

If so, though, it still leaves the question open of what link we use for the package drawing. This is also an open question for the non-EP SO packages in your PR, Rene.

poeschlr:
To be honest i kind of forgot about this pull request. If there is a fitting JEDEC standard then we can use that of course. If not then we might as well merge as is (possibly add the manufacturer name to the footprint name to show users that it is not really generic) We can then still link to the original datasheet but mention that the JEDEC standard was used for getting the tolerances. (plus adding the jedec name)

evanshultz:
I checked again, and MO-220 (all variations) shows pin width as 0.18/0.25/0.3mm (min/nom/max).  The datasheet only shows 0.23mm, which I assume is nominal. It seems this package matches.

@penoud 
As it sounds like you don't have the JEDEC doc, I've updated the script and pushed changes. Can you please check that the package in the datasheet will fit with the changes I made, and then re-generate the footprint if it looks OK to you?

chschlue:
#422

evanshultz:
@chschlue 
#422 has been merged.

chschlue:
@evanshultz 
I know. I mentioned #422 on all open, affected PRs back then for reference.

evanshultz:
@chschlue 
Since it was relatively easy to catch PRs here with multiple comments, I flagged you so if they only need a simple re-generation by the contributor with this repo updated it should be quick to wrap up the footprint and script PRs. :)

myfreescalewebpage:
@penoud the footprint need to be generated again following merge of #422, can you do that ?

myfreescalewebpage:
No news of the author.

PR #274, opened 2019-01-24
add QFN definition for Mini-Ciruits DQ1225

lorem-ipsum:
It is very close to a existing QFN footprint, but the specified pitch is 0.51mm instead of 0.5mm (and the tolerances are different).[datasheet](https://ww2.minicircuits.com/case_style/DQ1225.pdf)


poeschlr:
@pointhi the footprint is merged. This is good to go.

PR #275, opened 2019-01-24
add size-definition for SOP-8 Mini-Circuits XX211

lorem-ipsum:
Add a new SOP-8 variant used by Mini-Circuits called XX211 by them. [datasheet](https://ww2.minicircuits.com/case_style/XX211.pdf)

### DIFF
--- a/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/sop.yaml
+++ b/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/sop.yaml
@@ -2,6 +2,31 @@ FileHeader:
   library_Suffix: 'SO'
   device_type: 'SOP'
 
+
+SOP-8_4.14x5.33mm_P1.27mm:
+  size_source: 'https://ww2.minicircuits.com/case_style/XX211.pdf'
+  body_size_x:
+    nominal: 4.14
+    tolerance: 0.762
+  body_size_y:
+    nominal: 5.33
+    tolerance: 0.762
+  body_height:
+    max: 1.96
+  overall_size_x:
+    nominal: 6.35
+    tolerance: 0.762
+  lead_width:
+    nominal: 0.43
+    tolerance: 0.127
+  lead_len:
+    nominal: 0.76
+    tolerance: 0.127
+  pitch: 1.27
+  num_pins_x: 0
+  num_pins_y: 4
+
+
 SOP-8_6.62x9.15mm_P2.54mm:
   size_source: 'http://www.ti.com/lit/ds/symlink/iso1050.pdf'
   body_size_x:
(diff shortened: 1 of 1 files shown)


scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/sop.yaml
```diff
@@ -2,6 +2,31 @@ FileHeader:
   library_Suffix: 'SO'
   device_type: 'SOP'
 
+
+SOP-8_4.14x5.33mm_P1.27mm:
+  size_source: 'https://ww2.minicircuits.com/case_style/XX211.pdf'
+  body_size_x:
+    nominal: 4.14
+    tolerance: 0.762
```
poeschlr: All dimensions are given with 3 significant digits (imperial dimensions) meaning the tolerance should be 0.015 inch (=0.381mm)

Additionally: I think body dimensions are all given as maximums in the table. (meaning you would need to subtract the tolerance from the maximum to reach the nominal size.)

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/sop.yaml
```diff
@@ -2,6 +2,31 @@ FileHeader:
...
+    max: 1.96
+  overall_size_x:
+    nominal: 6.35
+    tolerance: 0.762
+  lead_width:
+    nominal: 0.43
+    tolerance: 0.127
+  lead_len:
+    nominal: 0.76
+    tolerance: 0.127
```
poeschlr: only lead width is listed as having the special (reduced) tolerance. meaning the lead lenght has the normal 0.015 inch tolerance.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/sop.yaml
```diff
@@ -2,6 +2,31 @@ FileHeader:
...
+    nominal: 4.14
+    tolerance: 0.762
+  body_size_y:
+    nominal: 5.33
+    tolerance: 0.762
+  body_height:
+    max: 1.96
+  overall_size_x:
+    nominal: 6.35
+    tolerance: 0.762
```
poeschlr: Dimensions D and E are for max/min of the overall size. So this should be minimum: 5.59 and maximum: 6.35.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/sop.yaml
lorem-ipsum: This annoying Mini-Circuits datasheets...
Yes, you're probably right about using the 0.015 inch tolerance. But I think they are the nominal dimensions as the tolerance is specified as +- 0.0015 inches.

lorem-ipsum: ah, sorry. For the body you're right, there it is specified.

lorem-ipsum: changing this also changes the name of the footprint, as the nominal size changes, creating the rather ugly `SOP-8_3.759x4.949mm_P1.27mm.kicad_mod`. Should I keep it or change it somehow?

lorem-ipsum: I agree. This is also the main source of confusion in their data sheets. I will round it as you suggested and make a remark about it.

lorem-ipsum:
I made the rounding and updated the PRs depending on this.

PR #281, opened 2019-01-28
Addition of MSOP-16

JacobEFO:
NOTE: I made some changes to handle correct silkscreen for DFN packages in
3322523 which is later reverted in 7b53a9a.

This pull request implements the following two footprints from: https://github.com/KiCad/kicad-footprints/pull/499

The packages are based on:
http://www.analog.com/media/en/package-pcb-resources/package/pkg_pdf/ltc-legacy-msop/05081669_A_MS16.pdf
http://www.analog.com/media/en/package-pcb-resources/package/pkg_pdf/ltc-legacy-msop/05081667_F_MSE16.pdf

Additionally ipc_gullwing_generator.py is updated to handle exposed pads correctly. An error was made in a variable.

Footprint pull request: https://github.com/KiCad/kicad-footprints/pull/1349


scripts/Packages/Package_Gullwing__QFP_SOIC_SO/ipc_gullwing_generator.py
```diff
@@ -170,8 +170,8 @@ def __createFootprintVariant(self, device_params, header, dimensions, with_therm
                     }
             else:
                 EP_size = {
-                    'x':device_dimensions['EP_size_x'].nominal,
-                    'y':device_dimensions['EP_size_y'].nominal
+                    'x':dimensions['EP_size_x'].nominal,
+                    'y':dimensions['EP_size_y'].nominal
```
poeschlr: Wow that was a stupid mistake of mine.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/ipc_gullwing_generator.py
```diff
@@ -102,7 +102,7 @@ def deviceDimensions(device_size_data):
             'lead_len': TolerancedSize.fromYaml(device_size_data, base_name='lead_len')
         }
         dimensions['has_EP'] = False
-        if 'EP_size_x' in device_size_data:
+        if 'EP_size_x_min' in device_size_data and 'EP_size_x_max' in device_size_data or 'EP_size_x' in device_size_data:
```
poeschlr: This is ok for now but i think we will remove the old style sometime in the future.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/msop.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+# By: JEFO
+# Email: jacob.overgaard.andersen@gmail.com
+MSOP-16_3x4mm_P0.5mm:
+  size_source: 'http://www.analog.com/media/en/package-pcb-resources/package/pkg_pdf/ltc-legacy-msop/05081669_A_MS16.pdf'
+  body_size_x:
+    nominal: 3
+    tolerance: 0.102
+  body_size_y:
+    nominal: 4
```
poeschlr: given as 4.039 in the datasheet

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/msop.yaml
```diff
@@ -0,0 +1,67 @@
...
+# Email: jacob.overgaard.andersen@gmail.com
+# 
+# Note: Exposed pad is slightly larger to avoid weird numbers
+MSOP-16-1EP_3x4mm_P0.5mm_EP1.65x2.85mm:
+  size_source: 'https://www.analog.com/media/en/package-pcb-resources/package/pkg_pdf/ltc-legacy-msop/05081667_F_MSE16.pdf'
+  body_size_x:
+    nominal: 3
+    tolerance: 0.102
+  body_size_y:
+    nominal: 4
```
poeschlr: given as 4.039 in the datasheet

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/msop.yaml
```diff
@@ -0,0 +1,67 @@
...
+    tolerance: 0.152
+  lead_width:
+    minimum: 0.17
+    maximum: 0.27
+  lead_len:
+    nominal: 0.53
+    tolerance: 0.152
+
+  EP_size_x:
+    nominal: 1.65
```
poeschlr: given as 1.651 in the datasheet

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/msop.yaml
```diff
@@ -0,0 +1,67 @@
...
+    tolerance: 0.102
+  EP_size_y:
+    nominal: 2.85
```
poeschlr: given as 2.845 in the datasheet

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/msop.yaml
```diff
@@ -0,0 +1,67 @@
...
+    tolerance: 0.102
+  EP_num_paste_pads: [2, 2]
+
+  thermal_vias:
+    count: [2, 3]
+    drill: 0.25
+    paste_via_clearance: 0.1
+    EP_num_paste_pads: [2, 2]
+    paste_avoid_via: False
```
poeschlr: with this set to false i suggest to set paste coverage a bit higher (0.7 should be enough)

JacobEFO: Alright, excellent idea.

Any rule of thumb for what solder paste coverage to use in kicad libraries?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/msop.yaml
JacobEFO: Yes, you are in fact correct. This is was "modified" as per suggestion of @Shackmeister, since 4.039 would lead to new 3d-models needed, long naming etc.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/msop.yaml
poeschlr: according to application nodes by different manufacturers it seems like anything between 50 to 80 percent can give good results depending on your exact process, the stencil thickness and solder used.

We typically simply try to get close to the middle of that range (so 65%) and increase it slightly for footprints where paste overlaps vias. (because one will lose some solder paste in such cases.) I typically used 75% in the past. But now reduced it to 70% as 75 seems a bit excessive. (some application notes suggested a range of 50 to 70 percent coverage so 75 might indeed be too much.)

poeschlr: I have for similar parts rounded it to 4mm. For soic and similar parts i think the body size is not that important. A better long term solution would be to make the footprint name generation stuff more powerful. (default pyhton format strings don't really offer the options we need.)
That way we could have the real size given in the yaml file while still getting reasonable footprint names.

JacobEFO: I think that's a great idea. However, for now what should we decide?

I can either use the exact values from the datasheet and get a long name and a footprint that matches exactly, and then add a 3d-package for it, or we can keep it like this and have a reasonably well fitting footprint with a "decent" name.

poeschlr:
Did i forget to request the inclusion of the part height?

This parameter would help with our 3d model generation.

So include whatever is given in the datasheet (for this parameter it is even ok if only maximum is given. We will need to make the 3d model generators smart enough to understand this.):
```yaml
overall_height:
  minimum:
  nominal:
  maximum:
  tolerance:
```

JacobEFO:
> Did i forget to request the inclusion of the part height?
> 
> This parameter would help with our 3d model generation.
> 
> So include whatever is given in the datasheet (for this parameter it is even ok if only maximum is given. We will need to make the 3d model generators smart enough to understand this.):
> 
> ```yaml
> overall_height:
>   minimum:
>   nominal:
>   maximum:
>   tolerance:
> ```

Okay, I'll update the footprints soon to use the exact values stated in the datasheet and include 3d model height.

I must have missed you reply here ;)

poeschlr:
thanks

PR #283, opened 2019-01-31
Added DIP-8-16 foot print

Misca1234:
Replacement push of
https://github.com/KiCad/kicad-footprints/pull/914

MN3005
http://www.experimentalistsanonymous.com/diy/Datasheets/MN3005.pdf

Foot print push
https://github.com/KiCad/kicad-footprints/pull/1352

Foot print script push
https://github.com/pointhi/kicad-footprint-generator/pull/283

This push includes modification of generic script 
/tools/footprint_scripts_DIP.py
It is now possible to skip pins, start pin numbering on the right side with a pre defined value and add a a string to the beginning of the file name.

bild

bild

bild


poeschlr:
@pointhi the footprint has been merged. This should be good enough to be merged. (I fear the DIP script is one of the ones that will need rewriting sometime in the future.)

Misca1234:
I think there is two different variants of the scripts

PR #284, opened 2019-01-31
add definition for Analog CP-16-22 LFCSP-16

lorem-ipsum:
Add the size definitions for Analogs CP-16-22 LFCSP-16-1EP footprint.
[datasheet](https://www.analog.com/media/en/package-pcb-resources/package/pkg_pdf/lfcspcp/cp-16/CP_16_22.pdf)


poeschlr:
The datasheet uses min/nom/max dimensioning. Use the same in the size definition file.

evanshultz:
YAML parameters match the package drawing and are formatted the same now. Footprint merged.

PR #287, opened 2019-02-03
Add some SOIC footprints to the script definition

poeschlr:


scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/soic.yaml
```diff
@@ -164,6 +164,59 @@ SOIC-8-1EP_3.9x4.9mm_P1.27mm_EP2.41x3.81mm:
     # bottom_pad_size:
     paste_avoid_via: False
 
+SOIC-8-1EP_3.9x4.9mm_P1.27mm_EP2.41x3.1mm:
```
evanshultz: This one has the mask size in the footprint name. Is this automatic in the script is the mask size is specified?

evanshultz:
Since this is the script, and the only place where it has an effect, I'll mention this again here: how should the package height be captured for parts that don't go down to the board surface? Body height is being added to help make 3D models from the same YAML data, but the QFNs/DFNs went down to the PCB. Should there be body_height_bottom which captures the distance from the board to the bottom of the package?

poeschlr:
As mentioned in the other comment of mine: I might have chosen the wrong name for this property. I always used body height to mean overall height (pcb to top of body)

I will think of a better name and make a one PR to fix them all thingy.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/soic.yaml
poeschlr: See line number 222:
```yaml
  EP_mask_x: 2.71
```

poeschlr:
I fixed the body height parameter in all of my scripts. I now call it "overall_height" to make it very clear what it means.

PR #276, opened 2019-01-25
Added VQFN-28-1EP_4x4mm and VQFN32-1EP_5x5mm

dominformant:
With this commits I added the VQFN-28-1EP_4x4mm and VQFN32-1EP_5x5mm footprints. The according pull request for the footprints is: [KiCad/kicad-footprints/pull/1337](https://github.com/KiCad/kicad-footprints/pull/1337)

PR #288, opened 2019-02-03
Fix wrong pin pitch in VQFN-28-1EP_4x4mm_P0.45mm_EP2.4x2.4mm

dominformant:
With #276 I asked to merge two footprints, one of which unfortunately contained a mistake in the pin pitch. I'm really, really sorry for that and fixed it as soon as I recognized my mistake. Please merge the fix to the master, so that the error does not last to long.


dominformant:
Thank you very much for merging and please excuse my mistake in the the original pull request.

PR #290, opened 2019-02-06
Add Texas RGE0024C definition

chemicstry:
Based on the RGE0024H, but smaller EP

http://www.ti.com/lit/ds/symlink/pca9548a.pdf#page=31

### DIFF
--- a/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/qfn/qfn_texas.yaml
+++ b/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/qfn/qfn_texas.yaml
@@ -410,3 +410,55 @@ Texas_VQFN_RGE0024H:
     # grid:
     # bottom_pad_size:
     paste_avoid_via: False
+
+Texas_VQFN_RGE0024C:
+  device_type: 'QFN'
+  manufacturer: 'Texas'
+  #part_number: 'mpn'
+  size_source: 'http://www.ti.com/lit/ds/symlink/pca9548a.pdf#page=31'
+  ipc_class: 'qfn' # 'qfn_pull_back'
+  #ipc_density: 'least' #overwrite global value for this device.
+  custom_name_format: 'Texas_RGE00{pincount}C_EP{ep_size_x:g}x{ep_size_y:g}mm{vias:s}'
+  body_size_x_min: 3.9
+  body_size_x: 4.0
+  body_size_x_max: 4.1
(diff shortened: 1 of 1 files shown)


poeschlr:
Could you please use the newer style of defining dimensions:

```yaml
dimension_name:
  minimum: xx.x
  nominal: xx.x
  maximum: xx.x
  tolerance: xx.x # or [-xx.x, xx.x] for asymmetrical tolerances
```

use the same style as in the datasheet. If the datsheet uses nominal plus tolerance then use nominal plus tolerance. If it uses minimum, maximum and nominal then use the same. (If only min/max is given then leave out the nominal dimension.)

The only dimension that does not use this style is the pin pitch. (assumed to be given as a reference value with 0 tolerance.)

If a dimension is given as reference then use nominal plus tolerance with tolerance set to 0.

Could you also include what is given of the overall height (I suggest right after body_size_y)?
```yaml
overall_height:
  maximum: 1
```

You can remove the ep size limit parameters. (They are no longer required)

EP paste coverage of 80% seems excessive. It is right at the maximum suggested value i ever read in any document.
I typically use 65% for parts with no vias or where vias are avoided by paste and 70% where paste overlaps vias.

Note: I did not check any parameters in detail. (Do not really have time for that right now.)

chemicstry:
Updated as requested. I assume via drill size does not use that format?

80% EP paste coverage was given in the datasheet, should I reduce it to 70% anyway?

DanSGiesbrecht:
@poeschlr @pointhi, do these changes look good?

evanshultz:
@chemicstry 
Just two things to do:
1. Use 65% paste coverage. Our library is general purpose so using our standard values will be correct.
1. The datasheet page is 34, not 31.

chemicstry:
@evanshultz 
Fixed as requested. Should I update the [PR](https://github.com/KiCad/kicad-footprints/pull/1395) over at the kicad-footprints repo too? Or do these scripts get generated in an automated way?

evanshultz:
Looks good. Please update the footprint at https://github.com/KiCad/kicad-footprints/pull/1395 and then we can move back to the symbol. :)

evanshultz:
Hehe. Cross-posting. Yes, please.

evanshultz:
@pointhi 
Footprint has been merged.

evanshultz:
@pointhi 
Footprint has been merged.

evanshultz:
@pointhi 
Footprint has been merged.

poeschlr:
Thanks

PR #292, opened 2019-02-07
Add some more missing qfn footprints

poeschlr:
Inspired by the currently manually created uqfn footprints.


poeschlr:
@pointhi the footprints generated by this request have been merged. This should therefore be ready to go.

PR #295, opened 2019-02-15
size definition for WSON-8-1EP-3x2.5mm-1.2x1.5mmEP

lorem-ipsum:
This adds the size definition for a WSON type used by Texas Instruments.
[Mechanical specification](http://www.ti.com/lit/ml/mpds400/mpds400.pdf), used for [this regulator](http://www.ti.com/lit/ds/symlink/lp3982.pdf).


poeschlr:
use the `lead_to_edge` parameter to tell the script the pullback distance.

And please add the parameter  ```suffix: '_PullBack'``` to include this fact in the footprint name

lorem-ipsum:
I've added the two parameters and updated the corresponding footprint PR with the new output

poeschlr:

PR #296, opened 2019-02-21
Package SO: SO-4_4.4x3.9mm_P2.54mm scripetd version source

myfreescalewebpage:
Size definition from TLP3123, datasheet at https://toshiba.semicon-storage.com/info/docget.jsp?did=10047&prodName=TLP3123

PR on kicad footprint repo : https://github.com/KiCad/kicad-footprints/pull/1424

Joel

### DIFF
--- a/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/so.yaml
+++ b/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/so.yaml
@@ -30,8 +30,8 @@ SO-8_4.0x5.0mm_P1.27mm:
 SO-4_4.4x3.9mm_P2.54mm:
   size_source: 'https://toshiba.semicon-storage.com/info/docget.jsp?did=10047&prodName=TLP3123'
   body_size_x:
-    minimum: 4.15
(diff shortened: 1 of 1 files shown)


poeschlr:
The datasheet uses nominal plus tolerance but you use min/max.
Please enter dimensions in the same style as in the datasheet.

Example:
```yaml
  body_size_x:
    nominal: 4.4
    tolerance: 0.25
```

myfreescalewebpage:
@poeschlr fixed

poeschlr:
All measurements are given as nominal plus tolerance in the datasheet so please make all yaml specs fit that style of dimensioning. (Doing math is error prone this is why the yaml style must fit the datasheet.)

myfreescalewebpage:
oooops sorry for that, have made the previous fix too much quickly ! Should be correct now.

evanshultz:
The YAML file is correct but there is a big comment which perhaps should be removed? I've mentioned the scripts in general need documentation, but putting this in the middle of one component entry is not the right place.

myfreescalewebpage:
Done for this definition, probably this is something to cleanup every where if other are present, and written in a README file.

PR #300, opened 2019-02-25
fix typo bug in gullwing generator

lorem-ipsum:
This fixes a tiny typo bug in the gullwing generator. `device_dimensions` is called `dimensions` in this section. This error is occurs when trying to build the tqfp footprints.


pointhi:
should be fixed now

PR #301, opened 2019-02-25
Size definition for TQFP-100 with 5x5mm EP

lorem-ipsum:
[Mechanical drawing](https://www.analog.com/media/en/package-pcb-resources/package/pkg_pdf/tqfp_edsv/sv_100_4.pdf). It exists already in the library, but still with square pads and without thermal vias.

### DIFF
--- a/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/tqfp.yaml
+++ b/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/tqfp.yaml
@@ -93,6 +93,52 @@ TQFP_100:
   num_pins_x: 25
   num_pins_y: 25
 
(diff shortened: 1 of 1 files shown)


pointhi:
Please reference KiCad PR

lorem-ipsum:
The corresponding KiCad PR is [this one](https://github.com/KiCad/kicad-footprints/pull/1433)

poeschlr:
The fix you made in the generator has also been made by somebody else. This results in a merge conflict for this pull request.

please include the height of the component like is done for others. (parameter overall_height with in your case only maximum given. Or you could calculate the overall height from the body height and seating plane distance given which would allow for also minimum height to be included.)

lorem-ipsum:
I've added the overall height and rebased it against the current master (now without the fix)

poeschlr:
@pointhi this should also be good to go.

PR #307, opened 2019-03-04
LQFP-64 with 5mmx5mm EP

twam:
Footprint for https://github.com/KiCad/kicad-symbols/pull/1594


poeschlr:
The datasheet defines min and max values in addition to the nominal values. These must be included to get the correct footprint. (The tolerance is part of the pad calculation)

twam:
I added minimum/maximum values were given in the datasheet.

poeschlr:
You will need to update the footprint over on the footprint repo as well. (adding tolerance info will result in a different footprint. This is why it is required to be entered.)

twam:
Wanted to update this, after this was merged to avoid to many activity there. However rerun the script and updated https://github.com/KiCad/kicad-footprints/pull/1461.

poeschlr:
thanks

PR #315, opened 2019-03-19
Add entry for ST LGA-12 footprint

evanshultz:
https://www.st.com/resource/en/datasheet/lis2hh12.pdf

@poeschlr 
When I run the generator I get a footprint like this:
image

1. The courtyard isn't outside the body. Is this a known issue? Or did I do something wrong?
1. Would it be better to have silk all around the body except for the pin 1 corner?

I haven't looked at the code yet but wanted to ping you first.

@pointhi 
Please don't merge until the above questions are settled.

### DIFF
--- a/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/lga.yaml
+++ b/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/lga.yaml
@@ -1,3 +1,34 @@
+LGA-12_2x2mm_P0.35mm_LayoutBorder2x4y:
+  device_type: 'LGA'
+  library: Package_LGA
+  use_name_format: 'LGA'
+  ipc_density: 'least'
+  #manufacturer: 'man'
+  #part_number: 'mpn'
+  size_source: 'http://www.st.com/resource/en/datasheet/lis2hh12.pdf'
+  ipc_class: 'qfn_pull_back'
+  body_size_x_min: 1.85
+  body_size_x: 2.0
+  body_size_x_max: 2.15
+  body_size_y_min: 1.85
+  body_size_y: 2.0
+  body_size_y_max: 2.15
+  body_height_max: 1
+  lead_width_min: 0.25
+  lead_width_max: 0.25
+  lead_to_edge:
+    nominal: 0.1
+  lead_len_min: 0.275
+  lead_len_max: 0.275
+  pitch: 0.5
+  num_pins_x: 2
+  num_pins_y: 4
+  #chamfer_edge_pins: 0.25
+  #pin_count_grid:
+  #pad_length_addition: 0.5
+  #suffix: '_Pad{pad_x:.2f}x{pad_y:.2f}mm_HandSolder'
+  #include_suffix_in_3dpath: 'False'
+
 LGA-16_3x3mm_P0.5mm_LayoutBorder3x5y:
   device_type: 'LGA'
   library: Package_LGA


poeschlr:
This is a bug. I never considered that there could be components with pads completely inside the body outline. I will create a pull request with a fix shortly. You can then merge the fix into your branch.

poeschlr:
Fix found in https://github.com/pointhi/kicad-footprint-generator/pull/369

evanshultz:
@poeschlr 
Thanks!

Merging your commit produces a courtyard offset of 0.1mm on this footprint. Is that what you expect?
image

evanshultz:
Let me ask this differently as I understand the IPC density being requested specifies 0.1mm courtyard offset: As I copied this footprint entry from another LGA footprint entry that had a 'least' density, but none of the other entries do, is there a reason? Should I switch this one to the nominal density? Do you know why one LGA-14 entry is using 'least'? Should they both use the nominal density?

poeschlr:
Yes unless there is a good reason we currently use nominal density. I do not know why that other part uses least but could be because there was a problem with generating it on nominal (pads too large or something similar. We could double check the old pull requests for that part)

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/lga.yaml
```diff
@@ -1,3 +1,34 @@
+LGA-12_2x2mm_P0.35mm_LayoutBorder2x4y:
+  device_type: 'LGA'
+  library: Package_LGA
+  use_name_format: 'LGA'
+  ipc_density: 'least'
+  #manufacturer: 'man'
+  #part_number: 'mpn'
+  size_source: 'http://www.st.com/resource/en/datasheet/lis2hh12.pdf'
```
poeschlr: include page number

evanshultz:
The other footprint using IPC least density was submitted by you at https://github.com/KiCad/kicad-footprints/pull/524. I don't see any note why I should not be nominal. I found the script file was moved at https://github.com/pointhi/kicad-footprint-generator/pull/258 and already had this density set then, and I didn't look into the history any more.

https://www.st.com/resource/en/datasheet/lis2dh.pdf does not show a recommended layout so that doesn't help. I've therefore commented out the density setting on that footprint.

---

I added the page number to my footprint as you requested, commented out the IPC density setting, and changed to the newer dimension format. I also caught a typo in the footprint definition for lead pitch. The footprint now looks good to me:
image

To try and answer questions ahead of making the footprint PR, here are two questions:
1. Should this footprint replace `Package_LGA/LGA-12_2x2mm_P0.5mm`? It was added at https://github.com/KiCad/Housings_LGA.pretty/pull/2 for the same part (I assume because when you implemented the LGA generator you re-created all existing LGA footprints).
1. Would you like me to re-generate the other footprint that previously had the non-nominal IPC density and submit to the footprint repo also?

---

Would you like for me to update the dimension format on other entries in this YAML file here?

poeschlr:
am i blind? is there no footprint pull request fitting this?

evanshultz:
@poeschlr 
No. Please see my comments above where I was asking for the scope you'd like in the footprint PR.

evanshultz:
@poeschlr 
Let me try this again. I thought maybe I could get an answer to https://github.com/pointhi/kicad-footprint-generator/pull/315#issuecomment-536661352, or again with https://github.com/pointhi/kicad-footprint-generator/pull/315#issuecomment-539058220, but... no.

This PR is about a LGA-12 which the generator didn't handle properly. It now does, thanks to you. There is a footprint in the library for the same MPN which doesn't match the name of footprints the script generates. I will submit a new PR for the LGA-12 footprint, but should I also delete the old footprint for the same MPN in that PR?

In addition, I submitted the updated LGA-14 (now with nominal density) at https://github.com/KiCad/kicad-footprints/pull/1956.

poeschlr:
I see two options: make the generator use the old name for backwards compatibility or delete the old footprint and reconnect symbols if necessary.

PR #322, opened 2019-03-22
Harwin fix

evanshultz:
`m20-89xx.py`:
Apparently the framework has been updated and requires another parameter. Just use an empty string as the argument and the resulting footprints seem to be identical.

`conn_harwin_m20-781xx45_smd_top_dual_row.py`:
PadArray() doesn't support unnamed pads (empty strings) as self.initialPin must be an int. So change this to use Pad() as a workaround.


scripts/Connector/Connector_Harwin/conn_harwin_m20-781xx45_smd_top_dual_row.py
```diff
@@ -92,13 +92,19 @@ def generate_footprint(pins, configuration):
...
-        y_spacing=pitch, type=Pad.TYPE_NPTH, shape=Pad.SHAPE_CIRCLE,
-        size=peg_drill_tht, drill=peg_drill_tht, layers=Pad.LAYERS_NPTH))
+    # kicad_mod.append(PadArray(start=[-1.27, -B/2], initial="", pincount=pins,
+        # y_spacing=pitch, type=Pad.TYPE_NPTH, shape=Pad.SHAPE_CIRCLE,
+        # size=peg_drill_tht, drill=peg_drill_tht, layers=Pad.LAYERS_NPTH))
+
+    # kicad_mod.append(PadArray(start=[1.27, -B/2], initial="", pincount=pins,
+        # y_spacing=pitch, type=Pad.TYPE_NPTH, shape=Pad.SHAPE_CIRCLE,
+        # size=peg_drill_tht, drill=peg_drill_tht, layers=Pad.LAYERS_NPTH))
+
```
poeschlr: please do not keep outdated code around in comment form. This is what the git history is there for.

In general: might it be better to make pad array more powerful instead of reimplementing this here?

evanshultz:
@poeschlr 
Correct. This is just to fix the script.

I removed the unneeded code.

poeschlr:
I looked at the pad array again. It should be able to handle empty string as start number (with increment 0, that is then implicitly set)

evanshultz:
Did you try it or just look at the code? If you ran it, at least for me in Win10 with Python 3.7.2, it doesn't work. I already investigated this myself and wrote it up above. PadArray tries to add the empty string of the 'initial' parameter to the int of the 'increment' parameter. Can't add a string and int.

I pushed a commit that prints the footprint names as they're being built for convenience and also changed tabs to spaces where it didn't result in branch conflicts.

There is a way to make PadArray work, of course, but my goal is to fix the scripts so they run and I've done that in a very reasonable and understandable way using the framework as it stands today. I think it is a good enhancement that should be done, but I'm unclear if you're asking for more information above or objecting to merging this. Can you please merge or clarify your position?

poeschlr:
The exclude pad stuff does not work at all for any non numerical pin numbering. This needs to be completely rewritten. Meaning this is good to go. Documented the problem in https://github.com/pointhi/kicad-footprint-generator/issues/436

Thanks for the fix

PR #314, opened 2019-03-19
Add SOP16 from STC (in essence a SOIC-16)

fvollmer:
Add the SOP16 from STC. This footprint is a little bit strange, because it is pretty much a SOIC-16.

Datasheet: https://www.stcmicro.com/datasheet/STC15F2K60S2-en.pdf (page 156)

It appears like the datasheet can't be open at the moment (it states "Bandwidth Limit Exceeded"). You can use https://web.archive.org/web/20180713001956/https://www.stcmicro.com/datasheet/STC15F2K60S2-en.pdf in the meantime.

PR #324, opened 2019-03-24
Add support for imperial units to ipc toleranced size class

poeschlr:
https://github.com/pointhi/kicad-footprint-generator/pull/314 references a datasheet with imperial only dimensioned drawings. This adds a unit switch for every dimension of a component. (options are inch and mil, anything else will be assumed to be mm.)

This does not change the generation of existing footprints (tested on the texas instrument qfn footprints as i still had generated footprints from before the change. diff only showed a change in the timestamp which is what i expected if this works.)

The dimensions for the part in the linked footprint will be added as a testcase shortly

### DIFF
--- a/scripts/tools/ipc_pad_size_calculators.py
+++ b/scripts/tools/ipc_pad_size_calculators.py
@@ -5,7 +5,18 @@ def roundToBase(value, base):
     return round(value/base) * base
 
 class TolerancedSize():
-    def __init__(self, minimum=None, nominal=None, maximum=None, tolerance=None):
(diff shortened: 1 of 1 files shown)


pointhi:
looks good to me

poeschlr:
The merge might have been a bit premature. Not necessarily the part in the toleranced dimension class but the handling in the generator. Mainly because there is no way to define the unit for the pitch. I think the unit might really be better served to be set per size definition as i did with the plcc generator. (I also doubt there will ever be a drawing out there that mixes units.)

PR #326, opened 2019-03-26
Adding Analog_BGA-28_4.0x6.25mm configuration

eeyrjmr:


evanshultz:
Can you please update the description to be closer to the non-Lattice parts? More description would be nice.

Also, how did you come up with the mask margin? That leaves a 0.3mm minimum mask width between balls, which I think should be OK, but where did you get this value?

eeyrjmr:
I can add more info to the description

As to the mask mask... The aim was 0.1mm total on the diameter but unsure how the script works. The 0.05 on the radius comes from what is achievable for a Class3 board. So the value that is entered into the script... is it diameter or radius with respect to the balls?

eeyrjmr:
https://www.smta.org/chapters/files/Wisconsin_Solberg-_SMTA_Great_Lakes_Chpt_2011.pdf

Page 19
bga

if you recommend small I can update.

evanshultz:
Mask margin is radius. It's the margin, which is intended to mean the difference in size. Maybe the wording could be better?

Anyway, I'm OK with the mask margin you have I just wanted to know from where you got that value.

If you can update the description let's merge this. Thank you!

evanshultz:
Please remove `_NSMD` from the footprint name. This doesn't need to be specified as it's the default for our library.

In the description, there are a few double spaces and also if this footprint can be used for other applications besides uModule then remove that too. The package drawing says 'uModule' on it so maybe this package is specific?

eeyrjmr:
_NSMD removed. it was only added due to other such packages had it. Additional whitespaces removed.

So far Analog are only using this type of grid for their uModule series and they have multiple variants. I don't see why they shouldn't make use of that for other chips.

evanshultz:
@pointhi 
The footprint has been merged.

PR #328, opened 2019-03-29
QFP_SOIC_SO: add size definition for TSOP-6 in 1.5x2.9mm package

zonque:
The footprint PR is here: https://github.com/KiCad/kicad-footprints/pull/1511

### DIFF
--- a/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/tsop.yaml
+++ b/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/tsop.yaml
@@ -1,0 +1,27 @@
+FileHeader:
+  library_Suffix: 'SO'
+  device_type: 'TSOP'
+
+TSOP-6_1.50mmx2.90mm_P0.95mm:
+  size_source: 'https://www.nxp.com/docs/en/package-information/SOT457.pdf'
+  body_size_x:
+    minimum: 1.3
+    maximum: 1.7
+  body_size_y:
+    minimum: 2.7
+    maximum: 3.1
+  overall_size_x:
+    minimum: 2.5
+    maximum: 3.0
+  overall_height:
+    minimum: 0.9
+    maximum: 1.1
+  lead_width:
+    minimum: 0.25
+    maximum: 0.40
+  lead_len:
+    minimum: 0.2
+    maximum: 0.6
+  pitch: 0.95
+  num_pins_x: 0
+  num_pins_y: 3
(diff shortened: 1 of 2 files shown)


scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/tsop.yaml
```diff
@@ -0,0 +1,26 @@
+FileHeader:
+  library_Suffix: 'SO'
+  device_type: 'TSOP'
+
+TSOP-6_1.50mmx2.90mm_P0.95mm:
+  size_source: 'https://www.nxp.com/packages/SOT457'
```
poeschlr: Link directly to the pdf please.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/tsop.yaml
```diff
@@ -0,0 +1,26 @@
...
+    minimum: 2.5
+    maximum: 3.0
+  overall_height:
+    minimum: 0.9
+    maximum: 1.1
+  lead_width:
+    minimum: 0.25
+    maximum: 0.40
+  lead_len:
+    nominal: 0.58
```
poeschlr: only minimum and maximum are given in the datasheet (dimension LP)

For reference: 0.2 ... 0.6

zonque: Both issues address now. Thanks!

evanshultz:
IMO `SC-70-6` is a more common and palatable name for this package. NXP also references this standard in the package drawing linked here.

zonque:
> IMO `SC-70-6` is a more common and palatable name for this package. NXP also references this standard in the package drawing linked here.

Would you set `custom_name_format` for that, or how is the naming fixed in such cases? It *is* a TSOP package after all, so `device_type` should remain untouched?

evanshultz:
IMO I'd change the entry. True, it's a TSOP, but I think SC-74 is a more common name. And SC-74 is an industry-standard package instead of a generic package type like TSOP. That's my experience and gut feel, though.

zonque:
Okay. Here's what I did now.

* I renamed the `size_definitions/tsop.yaml` to `size_definitions/sc-74.yaml`
* `device_type` is set to `SC-74` so the resulting file will be called `SC-74-6_1.5x2.9mm_P0.95mm.kicad_mod`

The only thing I'm wondering is whether the library this ends up in (`Package_SO.pretty`) is correct, as there are other `SC-` packages which reside in `Package_TO_SOT_SMD`.

WDYT?

poeschlr:
This is a good question. I feel there really is not a right way to do it as there really is no real clean way to differentiate these two libraries. 

For me it however looks more like a normal SO part than a SOT part as it is symetrical (Sot is mostly non symetrical. SC is sadly used for many very different package types.)

poeschlr:
Thanks

PR #331, opened 2019-04-01
Texas R-PDSO-G6 DDC

jneiva08:
Added parameters for R-PDSO-G6 DDC, is an SOT-23-6 variation.

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1489

### DIFF
--- a/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/sot.yaml
+++ b/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/sot.yaml
@@ -1,0 +1,24 @@
+FileHeader:
+  library_Suffix: 'SOT'
+  device_type: 'SOT'
+
+R-PDSO-G6_DDC:
+  size_source: 'http://www.ti.com/lit/ds/symlink/lm5050-1.pdf'
+  body_size_x:
+    minimum: 1.45
+    maximum: 1.75
+  body_size_y:
+    minimum: 2.75
+    maximum: 3.05
+  overall_size_x:
+    minimum: 2.55
+    maximum: 3.05
+  lead_width:
+    minimum: 0.30
+    maximum: 0.50
+  lead_len:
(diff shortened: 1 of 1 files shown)


evanshultz:
Dimensions are good.

Since this is a generic footprint, though, let's not make it specific to TI. You can add the JEDEC info like is done at (see `SOIC-8_3.9x4.9mm_P1.27mm` at https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/soic.yaml). I verified the package drawing from TI matches the JEDEC doc except for `overall_size_x`, which is 2.6-3mm. If we reference the JEDEC doc we can probably just tighten up that one spec a bit to match.

evanshultz:
`Device_type` may not be quite right here, since SOT-363, SC-70, and various other related packages could be in this YAML file. Perhaps we need to do this per-entry?

@poeschlr 
Are you OK with all of this?

jneiva08:
Have changed library_Suffix to be in Package_TO_SOT_SMD
Changed name as suggested on symbol thread, but the tsot-23-6 already exists and the dimensions are a little different.

how shall proceed?

evanshultz:
We can overwrite the existing footprint with this new one. It shouldn't be a big change and it's a fix (in that we believe IPC is the best way to go for this library).

Can you please change the `Device_type` field in the script as mentioned above. Instead of for the whole file this should be on each entry in the file, I think. You?

poeschlr:
The script does not currently support overwriting the device type from the definition of a single package. I would not expect anybody to add this stuff (would not be mad if somebody did it either)

evanshultz:
@poeschlr 
Since this is a JEDEC-standard package, I would like to call it `TSOT-23-6` (as done in the script here) and overwrite the existing package. Essentially going from a hand-made footprint to an IPC-compliant one. Is that OK with you?

PR #334, opened 2019-04-02
Correct Micro-Fit 3.0 drill sizes to datasheet

evanshultz:
As reported at https://github.com/KiCad/kicad-footprints/issues/1518. THT pad drill from 1mm to 1.02mm are minor, but the mounting holes were 1mm and should have been 1.27mm.


poeschlr:
footprints have been merged

PR #335, opened 2019-04-03
Added HVSON8 footprint for NXP PCF8523 RTC

johnfwhitmore:
I've taken a copy of another footprint's yaml definitions as a base to start from. As a result there's a lot of comments as I've not got min and max values. I was hoping to create PR to check with CI but looks like that's not possible.

datasheet:https://www.nxp.com/docs/en/data-sheet/PCF8523.pdf

Screenshot from 2019-04-03 18-32-12

### DIFF
--- a/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/hvson8.yaml
+++ b/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/hvson8.yaml
@@ -7,23 +7,23 @@ HVSON-8-1EP_3.25x2.35mm_P0.8mm_E1.1x1.75mm:
   #ipc_density: 'least' #overwrite global value for this device.
   #body_size_x_min: 2.85
   body_size_x:
-  #  minimum: 5.9
-    nominal: 4.95
-  #  maximum: 6.1
+    minimum: 3.9
+    nominal: 4.0
+    maximum: 4.1
   body_size_y:
-  #  minimum: 4.9
-    nominal: 3.25
-  #  maximum: 5.1
+    minimum: 3.9
+    nominal: 4.0
+    maximum: 4.1
   # overall_height:
-  #  minimum: 0.70
-  #  nominal: 0.75
+    minimum: 1.00
+  #  nominal: 1.0
   #  maximum: 0.80
   lead_width:
-  #  minimum: 0.35
+    minimum: 0.30
     nominal: 0.35
-  #  maximum: 0.48
+    maximum: 0.40
   lead_len:
-  #  minimum: 0.55
+  #  minimum: 0.4
     nominal: 0.95
   #  maximum: 0.65
 
--- a/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/hvson8.yaml
+++ b/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/hvson8.yaml
@@ -1,40 +1,31 @@
-HVSON-8-1EP_3.25x2.35mm_P0.8mm_E1.1x1.75mm:
+HVSON-8-1EP_4x4mm_P0.8mm_E2.2x3.1mm:
   device_type: 'HVSON'
   library: Package_SON
-  #manufacturer: 'NXP'
-  size_source: 'https://www.nxp.com/docs/en/data-sheet/PCF8523.pdf (page 66)'
-  ipc_class: 'qfn_pull_back' #'qfn' | 'qfn_pull_back'
+  manufacturer: 'NXP'
+  size_source: 'https://www.nxp.com/docs/en/data-sheet/PCF8523.pdf (page 57)'
+  ipc_class: 'qfn' #'qfn' | 'qfn_pull_back'
   #ipc_density: 'least' #overwrite global value for this device.
-  #body_size_x_min: 2.85
   body_size_x:
     minimum: 3.9
-    nominal: 4.0
     maximum: 4.1
   body_size_y:
     minimum: 3.9
-    nominal: 4.0
     maximum: 4.1
-  # overall_height:
+  overall_height:
     minimum: 1.00
-  #  nominal: 1.0
-  #  maximum: 0.80
   lead_width:
     minimum: 0.30
-    nominal: 0.35
     maximum: 0.40
   lead_len:
-  #  minimum: 0.4
-    nominal: 0.95
-  #  maximum: 0.65
+    minimum: 0.4
+    maximum: 0.65
 
   EP_size_x:
-  #  minimum: 3.35
-    nominal: 1.1
-  #  maximum: 3.45
+    minimum: 2.05
+    maximum: 2.35
   EP_size_y:
-  #  minimum: 4.25
-    nominal: 1.75
-  #  maximum: 4.35
+    minimum: 2.95
+    maximum: 3.25
   # EP_paste_coverage: 0.65
   EP_num_paste_pads: [2, 2]
 
--- a/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/hvson8.yaml
+++ b/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/hvson8.yaml
@@ -1,4 +1,4 @@
-HVSON-8-1EP_4x4mm_P0.8mm_E2.2x3.1mm:
+HVSON-8-1EP_4x4mm_P0.8mm_EP2.2x3.1mm:
   device_type: 'HVSON'
   library: Package_SON
   manufacturer: 'NXP'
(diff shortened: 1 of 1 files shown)


johnfwhitmore:
There are automatic tests and they worked. But I struggle with sizes of devices, so I've probably not got this 100% as yet. Let me know the required changes and I'll amend.

evanshultz:
You can check out the same tests CI uses at https://github.com/KiCad/kicad-library-utils and run them locally.

Min and max values are given on page 57 of the datasheet. See the table below the package drawing. Enter that info and remove what you don't need instead of leaving comments (naturally comments are OK while we're discussing so I'm taking about in the final entry).

johnfwhitmore:
Scissors missed that page. Thanks I'll get back to you.

johnfwhitmore:
Just remembered why I ignore that page of information. That data on page 57 is for the physical device, I'm trying to enter the data for a footprint. So the pin length on the device has min of 0.4 and a max of 0.65 but the footprint only has a value of 0.95 which isn't between the min and the max.

Similarly the pad under the device should I use the data from the device or the data from the footprint. I've gone with footprint, as that's what I'm creating.

johnfwhitmore:
Screenshot from 2019-04-03 19-01-45

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/hvson8.yaml
```diff
@@ -0,0 +1,44 @@
+HVSON-8-1EP_3.25x2.35mm_P0.8mm_E1.1x1.75mm:
+  device_type: 'HVSON'
+  library: Package_SON
+  #manufacturer: 'NXP'
+  size_source: 'https://www.nxp.com/docs/en/data-sheet/PCF8523.pdf (page 66)'
+  ipc_class: 'qfn_pull_back' #'qfn' | 'qfn_pull_back'
```
evanshultz: Use `qfn` as this package doesn't have pull-back leads.

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/hvson8.yaml
```diff
@@ -0,0 +1,44 @@
+HVSON-8-1EP_3.25x2.35mm_P0.8mm_E1.1x1.75mm:
+  device_type: 'HVSON'
+  library: Package_SON
+  #manufacturer: 'NXP'
+  size_source: 'https://www.nxp.com/docs/en/data-sheet/PCF8523.pdf (page 66)'
```
evanshultz: Page 57

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/hvson8.yaml
```diff
@@ -0,0 +1,44 @@
+HVSON-8-1EP_3.25x2.35mm_P0.8mm_E1.1x1.75mm:
```
evanshultz: Dimensions are 4x4mm with an EP of 2.2x3.1mm.

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/hvson8.yaml
```diff
@@ -0,0 +1,44 @@
...
+  device_type: 'HVSON'
+  library: Package_SON
+  #manufacturer: 'NXP'
+  size_source: 'https://www.nxp.com/docs/en/data-sheet/PCF8523.pdf (page 66)'
+  ipc_class: 'qfn_pull_back' #'qfn' | 'qfn_pull_back'
+  #ipc_density: 'least' #overwrite global value for this device.
+  #body_size_x_min: 2.85
+  body_size_x:
+    minimum: 3.9
+    nominal: 4.0
```
evanshultz: Omit nominal if not given (the scripts calculates all info needed or lets you know what's missing).

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/hvson8.yaml
```diff
@@ -0,0 +1,44 @@
...
+    maximum: 4.1
+  body_size_y:
+    minimum: 3.9
+    nominal: 4.0
+    maximum: 4.1
+  # overall_height:
+    minimum: 1.00
+  #  nominal: 1.0
```
evanshultz: Remove these commented lines. Only max is OK for height.

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/hvson8.yaml
```diff
@@ -0,0 +1,44 @@
...
+  #  maximum: 0.80
+  lead_width:
+    minimum: 0.30
+    nominal: 0.35
```
evanshultz: Remove nominal.

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/hvson8.yaml
```diff
@@ -0,0 +1,44 @@
...
+    maximum: 0.40
+  lead_len:
```
evanshultz: Use the min and max values of L and remove the nominal one.

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/hvson8.yaml
```diff
@@ -0,0 +1,44 @@
...
+  #  minimum: 0.4
+    nominal: 0.95
+  #  maximum: 0.65
+
+  EP_size_x:
```
evanshultz: Use Eh and Dh values for the EP size, min and max with no nominal.

evanshultz:
The scripts generate a footprint from the package dimensions. So you enter the package info.

This is documented in IPC 7351 so search for that document if you want to know the methodology used to derive the footprint from the package design.

I've made some comments inline. While the scripts are powerful the documentation is missing so there's understandably a learning curve.

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/hvson8.yaml
johnfwhitmore: ok   :(

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/hvson8.yaml
johnfwhitmore: Yes sorry

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/hvson8.yaml
johnfwhitmore: OK  IPC 7351 right so there is a reason for this. I'll search for that document, thanks

johnfwhitmore:
#ipc_density: 'least' #overwrite global value for this device.

and

  # EP_paste_coverage: 0.65
  EP_num_paste_pads: [2, 2]

I've only two comments in the file now. I'll look into them in IPC7351

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/hvson8.yaml
```diff
@@ -0,0 +1,35 @@
+HVSON-8-1EP_4x4mm_P0.8mm_E2.2x3.1mm:
```
evanshultz: Missing the `P` in `EP`.

johnfwhitmore: Oops

evanshultz:
Those comments are fine to leave. They will need a thorough cleaning up later but there's too many of them to worry about individually at this time. Breakdown:
- We're using density B (nominal density) for all footprints now.
- Paste coverage is not in the IPC doc. It's a colloquial choice and the script uses a reasonable default value unless overridden.

johnfwhitmore:
OK thanks for that. I uncommented out the manufacturer. The nxp prefix doesn't look right to me. But I guess it's not a standard device and is NXP specific.

Screenshot from 2019-04-03 19-42-20

evanshultz:
It is standard. Page 57 states this package conforms to JEDEC MO-229.

johnfwhitmore:
So should I remove the manufacturer element?

evanshultz:
Yes please.

evanshultz:
Looks good. Thanks!

evanshultz:
@pointhi 
The footprint has been merged.

johnfwhitmore:
Thank you. I'll try the third device tomorrow and make it a cleaner addition

PR #336, opened 2019-04-05
Add ADF7021 footprint entry

evanshultz:
https://www.analog.com/media/en/package-pcb-resources/package/pkg_pdf/lfcspcp/cp_48_5.pdf


poeschlr:
@pointhi footprint is merged, this is good to go.

PR #337, opened 2019-04-10
add SOIC-8 for THS3491

aewallin:
SOIC-8 variant for THS3491. dimensions from datasheet page 48/53
http://www.ti.com/lit/ds/symlink/ths3491.pdf

kicad-footprints PR https://github.com/KiCad/kicad-footprints/pull/1541


aewallin:
this footprint was already included in the kicad libraries with another "Texas_xxx" name - so I am closing this as it is a duplicate.

PR #340, opened 2019-04-12
add Texas SO-20 PowerPAD

aewallin:
Used by e.g. OPA569:
http://www.ti.com/lit/ds/symlink/opa569.pdf

Further details on footprint at:
http://www.ti.com/lit/an/slma004b/slma004b.pdf


scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/so.yaml
```diff
@@ -47,3 +47,57 @@ SO-4_4.4x3.9mm_P2.54mm:
...
+    minimum: 1.35
+    maximum: 1.53
+  pitch: 1.27
+  num_pins_x: 0
+  num_pins_y: 10
+
+  # http://www.ti.com/lit/an/slma004b/slma004b.pdf
+  EP_size_x:
+    nominal: 6.045 # 0.238"
+    tolerance: 0
```
poeschlr: give the size of the exposed pad (package side) here and use EP_size_x_overwrite to define the size on the pcb (this way we can reuse the same info for the 3d model sometime in the future)

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/so.yaml
```diff
@@ -47,3 +47,57 @@ SO-4_4.4x3.9mm_P2.54mm:
...
+
+Texas_PowerPAD_SO-20-1EP_7.52x12.83mm_P1.27mm:
+  size_source: 'http://www.ti.com/lit/ds/symlink/opa569.pdf'
+  #custom_name_format: Texas_PowerPAD_SO-{pincount:d}-1EP_{size_x:.2g}x{size_y:.2g}mm_P{pitch:.2g}mm
+  body_size_x:
+    minimum: 7.45
+    maximum: 7.59
+  body_size_y:
+    minimum: 12.7
+    maximum: 12.95
```
poeschlr: add overall_height parameter (for use in 3d model generation)

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/so.yaml
```diff
@@ -47,3 +47,57 @@ SO-4_4.4x3.9mm_P2.54mm:
...
+  overall_size_x:
+    minimum: 10.16
+    maximum: 10.65
+  lead_width:
+    minimum: 0.35
+    maximum: 0.51
+  lead_len:
+    minimum: 1.35
+    maximum: 1.53
```
poeschlr: where do thes values come from? The datasheet gives 0.4 ... 1.27

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/so.yaml
```diff
@@ -47,3 +47,57 @@ SO-4_4.4x3.9mm_P2.54mm:
...
+  pitch: 1.27
+  num_pins_x: 0
+  num_pins_y: 10
+
+  # http://www.ti.com/lit/an/slma004b/slma004b.pdf
+  EP_size_x:
+    nominal: 6.045 # 0.238"
+    tolerance: 0
+  EP_mask_x: 2.159
```
poeschlr: The datasheet of your opa uses a pad size of 2.72..3.81 which is much larger. Is it possible that the second datasheet is not for exactly this one package but shows the layout of another part as an example?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/so.yaml
```diff
@@ -47,3 +47,57 @@ SO-4_4.4x3.9mm_P2.54mm:
   pitch: 2.54
   num_pins_x: 0
   num_pins_y: 2
+
+Texas_PowerPAD_SO-20-1EP_7.52x12.83mm_P1.27mm:
+  size_source: 'http://www.ti.com/lit/ds/symlink/opa569.pdf'
```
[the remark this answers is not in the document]
aewallin: I added a repeated size_source line - hope that works.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/so.yaml
aewallin: now added both EP_size and _overwrite lines.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/so.yaml
aewallin: corrected with datasheet numbers

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/so.yaml
aewallin: I don't know of another source for the PowerPAD recommended land-pattern - so I'm assuming it is ok.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/so.yaml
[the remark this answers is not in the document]
poeschlr: simply add the second source to the same parameter separated with a comma (inside the same string. meaning `size_source: 'doc1, doc2'`

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/so.yaml
poeschlr: I can not imagine that this would be ok. The pad on the package is larger than the area free of solder mask. This will create massive problems with reflow soldering. I suggest you set the mask cutout to the same value as the nominal package pad and slightly increase the via grid spacing to still have them spaced equally in this area.

aewallin:
On page 9 of this document there is a Note describing the recommended mask-opening as the max size of the EP - I have now changed this.
http://www.ti.com/lit/an/slma002h/slma002h.pdf

poeschlr:
You fixed the grid for the thermal vias but massively increased the uncovered area. This looks very strange now.
Screenshot from 2019-06-03 20-23-50

aewallin:
@poeschlr are there still changes requested from you?
I tried to re-base this PR on latest master.

aewallin:
This now follows FIG16 from the OPA569 datasheet.
so-20_vias

poeschlr:
Could you explain why your pull request touches this many things? And has this many commits?
Maybe a good idea would be to start fresh from master and cherrypick only the commits that are really needed.

aewallin:
closing. see #474 instead.

PR #342, opened 2019-04-22
replacement of all yaml.load with yaml.safe_load. CVE-2017-18342

eeyrjmr:
Bulk change via **find . -type f | xargs sed -i  's/yaml.load/yaml.safe_load/g'** to mitigate  CVE-2017-18342


pointhi:
Thanks for correcting these.

PR #343, opened 2019-04-24
adds xilinx artix 7 based bga footprints and functionality

herostrat:
All dimensions and information: 
https://www.xilinx.com/support/documentation/user_guides/ug475_7Series_Pkg_Pinout.pdf

Additional Information and discussion here: https://github.com/KiCad/kicad-footprints/pull/1557


scripts/Package_BGA/bga_xilinix.yml
```diff
@@ -0,0 +1,148 @@
+# Artix-7 FPGA Packages
+# Notes:
+#  - Often names are similiar, e.g. FB, FBG and FBV. I choose FBG as it is the common one.
+#    The addition stems from different solder types for the BGA balls
+#  - Xilinx_CPG236 and Xilinx_CPG236 are the same, different name comes from bank differences
```
evanshultz: The two packages listed here are the same but should be different.

herostrat: You are absolutely correct.
The second name should be Xilinx_CPG238.

The change is already pushed.
Thanks for the fast feedback!

scripts/Package_BGA/bga_xilinix.yml
```diff
@@ -0,0 +1,148 @@
...
+  additional_tags: "CP236 CPG236"
+  pkg_width: 10
+  pkg_height: 10
+  pitch: 0.5
+  pad_diameter: 0.275
+  mask_margin: 0.05
+  layout_x: 19
+  layout_y: 19
+  row_skips: [[], [], [], [[4,17]], [[4,17]], [[4,17]], [[4,7],[14,17]], [[4,7],[14,17]], [[4,7],[14,17]], [[4,7],[9,12],[14,17]], [[4,7],[14,17]], [[4,7],[14,17]], [[4,7],[14,17]], [[4,17]], [[4,17]], [[4,17]], [], [], []]
+Xilinx_CPG238:
```
evanshultz: Footprints with only a single package matching them don't use the `additional_tags` key; should all entries here use that for consistency even if there's only a single package mentioned?

scripts/Package_BGA/bga_xilinix.yml
```diff
@@ -0,0 +1,148 @@
...
+  pad_diameter: 0.5
+  mask_margin: 0.05
+  layout_x: 22
+  layout_y: 22
+Xilinx_RB484:
+  description: "Artix-7 BGA, 22x22 grid, 23x23mm package, 1mm pitch; https://www.xilinx.com/support/documentation/user_guides/ug475_7Series_Pkg_Pinout.pdf, NSMD pad definition Appendix A"
+  pkg_width: 23
+  pkg_height: 23
+  pitch: 1
+  pad_diameter: 0.5
```
evanshultz: Should be 0.53mm.

evanshultz:
Normally I'd prefer to keep script changes and YAML input definitions changes separate, but the script change is minor so let's do it.

I will merge the footprints like this, but note that the 3D models will be wrong because the difference in height for packages using the same footprint will not be reflected as the footprints are mapped to a single 3D model. I think we should rename some parameters in the YAML file to use 'X' and 'Y' instead of 'width' and 'height', along with adding the package height, but I can go back and do that for all BGA entries once you've finished touching the files here and there won't be branch conflicts. I don't want to make any more hassle for you to finish this work.

Another minor note is that the script ought to be able to tack on pitch, body size, etc. in the description from the input parameters instead of needing that typed in twice. Another thing I'll fix once you've wrapped up adding all these footprints.

Some thoughts:
1. Please add the page number to the link to UG475 for clarity.
1. If the packages are identical, like CSG324 and CSG325, what about merging them? For example a single footprint called `Xilinx_CSG324_CSG325`? Naturally, this is at odds with having unique footprints for every unique 3D model.

scripts/Package_BGA/bga_xilinix.yml
herostrat: I think you are right, I will include available options into the tags.
Therefore it is consistent over the hole range.

scripts/Package_BGA/bga_xilinix.yml
herostrat: Wow, good catch.
Will be updated.
I will be more careful in the future!

herostrat:
> Normally I'd prefer to keep script changes and YAML input definitions changes separate, but the script change is minor so let's do it.

I will take this into account in the future!

> I will merge the footprints like this, but note that the 3D models will be wrong because the difference in height for packages using the same footprint will not be reflected as the footprints are mapped to a single 3D model. I think we should rename some parameters in the YAML file to use 'X' and 'Y' instead of 'width' and 'height', along with adding the package height, but I can go back and do that for all BGA entries once you've finished touching the files here and there won't be branch conflicts. I don't want to make any more hassle for you to finish this work.

I devided the footprints into units with the same 3D model.
The height is not that easy to define for all packages, as some are not quadratic but have interesting shapes, i.e.:
grafik
But I agree with the change of keywords, and as a basic reference the maximum height should suffice.
I can make the required changes if you like before I continue with my list in https://github.com/KiCad/kicad-footprints/issues/1560 

> Another minor note is that the script ought to be able to tack on pitch, body size, etc. in the description from the input parameters instead of needing that typed in twice. Another thing I'll fix once you've wrapped up adding all these footprints.

Good idea, that should minimize errors in copying information.

> Some thoughts:
> 
>     1. Please add the page number to the link to UG475 for clarity.

Done.

>     2. If the packages are identical, like CSG324 and CSG325, what about merging them? For example a single footprint called `Xilinx_CSG324_CSG325`? Naturally, this is at odds with having unique footprints for every unique 3D model.

I would agree that merging would be possible, but because there are so few and the work to divide them afterwards, if 3d models would be added in the future, is very high I would think that letting them be divided is better and easier to understand.

evanshultz:
Let's leave the keywords alone for now. Thanks for offering.

Everything else sounds good and hopefully we can get these footprints all added soon since it looks like symbols may be coming quickly as well!

scripts/Package_BGA/bga.py
```diff
@@ -177,7 +182,7 @@ def crtYdRound(x):
     f.append(Model(filename="{}Package_{}.3dshapes/{}.wrl".format(
                   config['3d_model_prefix'], package_type, fp_id)))
 
-    f.setTags("{} {} {}".format(package_type, balls, pitch_string))
+    f.setTags("{} {} {} {}".format(package_type, balls, pitch_string, additionalTag))
```
evanshultz: This will put a space regardless of if the 4th argument is present or not.  Let's remove the space and format as I mentioned above.

scripts/Package_BGA/bga.py
```diff
@@ -25,6 +25,11 @@ def generateFootprint(config, fp_params, fp_id):
     layoutX = fp_params["layout_x"]
     layoutY = fp_params["layout_y"]
     
+    if "additional_tags" in fp_params:
+        additionalTag = fp_params["additional_tags"]
+    else:
+        additionalTag = []
```
evanshultz: Since this is a string just use an empty string. It's not a list.

scripts/Package_BGA/bga.py
```diff
@@ -25,6 +25,11 @@ def generateFootprint(config, fp_params, fp_id):
     layoutX = fp_params["layout_x"]
     layoutY = fp_params["layout_y"]
     
+    if "additional_tags" in fp_params:
+        additionalTag = fp_params["additional_tags"]
```
evanshultz: Put a space in front of this so there isn't a space added in the .format() statement for each conditions.

scripts/Package_BGA/bga_xilinx.yml
```diff
@@ -0,0 +1,152 @@
...
+#  - Often names are similiar, e.g. FB, FBG and FBV. I choose FBG as it is the common one.
+#    The addition stems from different solder types for the BGA balls
+#  - Xilinx_CPG236 and Xilinx_CPG238 are the same, different name comes from bank differences
+#  - Xilinx_CSG324 and Xilinx_CSG325 are the same, different name comes from bank differences
+#  - Xilinx_RS484 and Xilinx_SBG484 have same landing pattern but different heights/3D packages
+#  - Xilinx_RB484, Xilinx_FBG484 and Xilinx_FGG484 have same landing pattern but different heights/3D packages
+#  - Xilinx_RB676, Xilinx_FBG676 and Xilinx_FGG676 have same landing pattern but different heights/3D packages
+
+Xilinx_CPG236:
+  description: "Artix-7 BGA, 19x19 grid, 10x10mm package, 0.5mm pitch; https://www.xilinx.com/support/documentation/user_guides/ug475_7Series_Pkg_Pinout.pdf p. 266, NSMD pad definition Appendix A"
```
evanshultz: Please format the page section as `#page=266` so it will open to the PDF page directly.

herostrat:
I think that's it :)
I fixed the points and updated the footprint PR as well.

evanshultz:
@pointhi 
Footprints have been merged.

PR #346, opened 2019-05-01
adds xilinx virtex bga packages

herostrat:
adds the parameters for the virtex packages.

Important notes:

Virtex-7 FPGA Packages
Notes: 
 - FFG1157 and FFG1158 have same footprint and same 3D package and are therefore in one footprint under FFG1157
 - FFG1926, FFG1927, FFG1928 and FFG1930 have same footprint and same 3D package and are therefore in one footprint under FFG1926
 - FLG1925, FLG1926, FLG1928 and FLG1930 have same footprint and same 3D package and are therefore in one footprint under FLG1925
 - RF1157 and
 RF1158 have same footprint and same 3D package and are therefore in one footprint under RF1157

### DIFF
--- a/scripts/Package_BGA/bga_xilinx.yml
+++ b/scripts/Package_BGA/bga_xilinx.yml
@@ -150,3 +150,108 @@ Xilinx_FFG1156:
   mask_margin: 0.05
   layout_x: 34
   layout_y: 34
+  
+# Virtex-7 FPGA Packages
+# Notes: 
+#  - FFG1157 and FFG1158 have same footprint and same 3D package and are therefore in one footprint under FFG1157
+#  - FFG1926, FFG1927, FFG1928 and FFG1930 have same footprint and same 3D package and are therefore in one footprint under FFG1926
+#  - FLG1925, FLG1926, FLG1928 and FLG1930 have same footprint and same 3D package and are therefore in one footprint under FLG1925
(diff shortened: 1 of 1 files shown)


evanshultz:
@herostrat 
All of these look good, but I realized something I meant to communicate earlier but I think I had forgotten or done it too poorly. To make it easier for users to find the package as named by Xilinx, I would think having all relevant package names in the footprint would be best. So rename `Xilinx_RF1157` to `Xilinx_RF1157_RF1158`. That way the user can easily see all packages that apply to the footprint (with matching 3D models). What do you think?

herostrat:
@evanshultz 
Tbh I wanted to ask the same question, as the naming "collisions" will be even more visible in the remaining packages.
I will make adjustments accordingly.

One additional question:
There are some packages that are named the same in Artix and Kintex. The 3D model is the same as far as I can tell, too.
For them I will just update the description and add Kintex. Is this sensible?

evanshultz:
Yes, I think that's fine. The package doesn't need to be related to the family. In fact, I don't think the description needs to specify the family but only the package info (who knows that Xilinx will do in the future?).

evanshultz:
@herostrat 
Sorry for pushing, but are you planning to get to this soon? I'm ready to review and merge. :)

PR #350, opened 2019-05-08
adds definition for Xilinx Kintex-7 BGA FPGAs

herostrat:
Details: https://github.com/KiCad/kicad-footprints/issues/1560
Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1584

@evanshultz the next batch

Notes: I added Kintex and their page in the doc to the three footprints already added during the Artix batch.
I choose to keep the family in the description, as the footprint names do not seem to mean a lot to Xilinx. It could be that the same name for a different family can have a different 3d model in the future.

### DIFF
--- a/scripts/Package_BGA/bga_xilinx.yml
+++ b/scripts/Package_BGA/bga_xilinx.yml
@@ -81,7 +81,7 @@ Xilinx_FTG256:
   layout_x: 16
   layout_y: 16
 Xilinx_FBG484:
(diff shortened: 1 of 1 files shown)


evanshultz:
Hmm. Annoying. The packages look different in the drawing but we can't be sure if they're the same or not. Even though they share a name. I guess this is the best compromise. Weak sauce on Xilinx's part...

Everything looks good here. Thanks!

@pointhi 
Footprints have been merged.

evanshultz:
@herostrat 
I assume you still want to make a new script PR for each set of footprints.

@pointhi 
Could you please merge this so we can move on with the next batch? Thanks!

herostrat:
@evanshultz  yes, I have already prepared the footprints for the Zynq-7000 versions (which should be the last afaik) and had waited until this PR got merged.

PR #351, opened 2019-05-09
Ring pad support inspired by wuerth electronics smd mounting hardware

poeschlr:
Note this is work in progress. For ringed pads to work i will also need to implement:
- [x] polar coordinate support for Vector2d including rotating it around a given point
- [x] arcs with more options to define them
  - center, start, angle (original)
  - center, mid, angle
  - center, start, end, long_way (long way parameter needed as every two points plus center combination has two valid arcs. Note that the end point can be off the arc which will result in a warning and the radius from center to end is ignored only the angle is used.)
  - start, mid, end (prepared but not implemented)
- [x] split geometry stuff from kicad types (In part to avoid circular references)
  - [x] translate and rotate support for geometric primitives
  - [x] copy function and init from type(self)
  - [ ] cut with lines
     - missing for circle
- [ ] polar array  (If i get time)

I noticed that the serializer class can not really deal with inheritance. But i do not think i will have time to look into this.


poeschlr:
@pointhi the general idea of this is now done. There is still a bit of documentation and error checking to add but i think it should be far enough for you to take a look if you are happy with the general idea behind these changes.

poeschlr:
@pointhi do you have any feedback? Or simply no time?

pointhi:
I did a rough check and think it should be ok. I would suggest to not rely on Multiple Inheritance to much, because it makes code harder to read and more error prone

KicadModTree/util/geometric_util.py
```diff
@@ -0,0 +1,600 @@
...
+        rad_s, ang_s = self.start_pos.to_polar(origin=self.center_pos)
+        rad_p, ang_p = Vector2D(point).to_polar(origin=self.center_pos)
+
+        ang_p_s = (ang_p - ang_s) % 360
+        if self.angle < 0:
+            ang_p_s -= 360
+        return (rad_p, ang_p_s)
+
+    def _compareAngles(self, a1, a2, tolerance=1e-7):
+        r""" compare which of the two angles given in the local coordinate system
```
cpresser: compare what?

KicadModTree/util/geometric_util.py
```diff
@@ -0,0 +1,600 @@
...
+            origin=self.center_pos, use_degrees=True)
+
+    def _toLocalCoordinates(self, point):
+        rad_s, ang_s = self.start_pos.to_polar(origin=self.center_pos)
+        rad_p, ang_p = Vector2D(point).to_polar(origin=self.center_pos)
+
+        ang_p_s = (ang_p - ang_s) % 360
+        if self.angle < 0:
+            ang_p_s -= 360
+        return (rad_p, ang_p_s)
```
cpresser: rename to `_toLocalPolarCoordinates`

KicadModTree/nodes/base/Circle.py
```diff
@@ -41,14 +42,43 @@ class Circle(Node):
...
+        :params:
+            * *distance_vector* (``Vector2D``)
+                2D vector defining by how much and in what direction to translate.
+        """
+
+        self.center_pos += distance_vector
+        return self
+
+    def cut(self, *other):
+        raise NotImplemented("cut for circles not yet implemented")
```
cpresser: even if its not implemented, you could call the cut function of the base class.  Once its implemented you only need to change one code location.

PR #353, opened 2019-05-13
Add TI DWV dimensions

evanshultz:
http://www.ti.com/lit/ml/mpds382b/mpds382b.pdf


poeschlr:

PR #359, opened 2019-05-17
Added TQFP-48-1EP_7x7mm_P0.5mm_EP3.62x3.62mm

jstjst:
Footprint Datasheet: http://www.ti.com/lit/ml/mpqf051a/mpqf051a.pdf
Footprint is used in http://www.ti.com/lit/ds/symlink/tusb4020bi.pdf

### DIFF
--- a/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/tqfp.yaml
+++ b/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/tqfp.yaml
@@ -53,6 +53,57 @@ TQFP-48-1EP_7x7mm_P0.5mm_EP5.0x5.0mm:
     # bottom_pad_size:
     # paste_avoid_via: True
 
+TQFP-48-1EP_7x7mm_P0.5mm_EP3.62x3.62mm:
+  size_source: 'http://www.ti.com/lit/ml/mpqf051a/mpqf051a.pdf'
+  body_size_x:
+    nominal: 7
+    tolerance: 0
+  body_size_y:
+    nominal: 7
+    tolerance: 0
+  overall_size_x:
+    nominal: 9
+    tolerance: 0
+  overall_size_y:
+    nominal: 9
+    tolerance: 0
+  lead_width:
+    minimum: 0.17
+    maximum: 0.27
+  lead_len:
+    minimum: 0.45
+    maximum: 0.75
+  pitch: 0.5
+  num_pins_x: 12
+  num_pins_y: 12
+
+  EP_size_x:
+    minimum: 3.52
+    nominal: 3.62
+    maximum: 5.72
+  EP_size_y:
+    minimum: 3.52
+    nominal: 3.62
+    maximum: 5.72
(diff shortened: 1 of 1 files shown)


pointhi:
Please reference the footprint PR in KiCad

poeschlr:
It does not seem like there is a footprint open over on the footprint repo that would fit this contribution.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/tqfp.yaml
```diff
@@ -53,6 +53,57 @@ TQFP-48-1EP_7x7mm_P0.5mm_EP5.0x5.0mm:
     # bottom_pad_size:
     # paste_avoid_via: True
 
+TQFP-48-1EP_7x7mm_P0.5mm_EP3.62x3.62mm:
+  size_source: 'http://www.ti.com/lit/ml/mpqf051a/mpqf051a.pdf'
+  body_size_x:
+    nominal: 7
```
evanshultz: Use dimensioning from the datasheet. This should be 6.8mm min and 7.2mm max.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/tqfp.yaml
```diff
@@ -53,6 +53,57 @@ TQFP-48-1EP_7x7mm_P0.5mm_EP5.0x5.0mm:
...
+    maximum: 0.27
+  lead_len:
+    minimum: 0.45
+    maximum: 0.75
+  pitch: 0.5
+  num_pins_x: 12
+  num_pins_y: 12
+
+  EP_size_x:
+    minimum: 3.52
```
evanshultz: Only the nominal size is given for the EP, so get rid of min and max.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/tqfp.yaml
```diff
@@ -53,6 +53,57 @@ TQFP-48-1EP_7x7mm_P0.5mm_EP5.0x5.0mm:
...
+    nominal: 3.62
+    maximum: 5.72
+  EP_size_y:
+    minimum: 3.52
+    nominal: 3.62
+    maximum: 5.72
+  # EP_paste_coverage: 0.65
+  EP_num_paste_pads: [2, 2]
+  #heel_reduction: 0.1 #for relatively large EP pads (increase clearance)
```
evanshultz: Get rid of this comment line and the two lines below.

evanshultz:
@jstjst 
1. You need to use the part datasheet or add that document since the EP size isn't shown in the package drawing you currently link. Add `#page=421 to the end of the datasheet link.
1. I also added comments above.

evanshultz:
@jstjst 
Are you planning to come back to this PR soon?

chschlue:
#422

evanshultz:
@jstjst 
Are you available to make the fixes above?

PR #361, opened 2019-05-22
Add WQFN-24

ObKo:
This PR adds WQFN-24 with 0.5mm pitch, 2.6x2.6mm exposed pad and 0.8mm height - http://www.ti.com/lit/ds/symlink/lm26480.pdf#page=39

wqfn

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1608


poeschlr:

PR #362, opened 2019-05-23
Add footprint generator usage instructions

yankee14:
I have no experience with yaml or python, so I was stumbling through learning how to use this library. I would like to submit this PR in hope that it might save someone else some time!


pointhi:
Thx, looks good.

Issue #352, opened 2019-05-11
should we exclude the test stuff from code climate?


poeschlr:
Right now the code climate stuff mainly complains about the testcases. Makes it hard to see where one could invest time to make something more maintainable. My suggestion would therefore be to add the test directory to its ignore list.

poeschlr:
@pointhi i interpret you adding enhancement as a "I will consider it" and opened a pull request https://github.com/pointhi/kicad-footprint-generator/pull/366

If you decide against it simply close both the pull request and issue as invalid.

PR #366, opened 2019-06-02
Exclude tests from codeclimate analysis

poeschlr:
As proposed in https://github.com/pointhi/kicad-footprint-generator/issues/352


poeschlr:
Ok this removes > 70 false positives. If we want we can change it from excluding the test directory completely and do it for specific rules (example keep the line length check but remove everything else. But i think travis checks this anyways so no need to also use code climate here.)

pointhi:
thanks

PR #369, opened 2019-06-06
Fix no lead courtyard handling for pull back versions

poeschlr:
Fixes the problems discovered in https://github.com/pointhi/kicad-footprint-generator/pull/315

footprint pull request: https://github.com/KiCad/kicad-footprints/pull/1630


poeschlr:
@pointhi the footprints have been merged. I however do not think i should merge my own changes without somebody else looking over them so could you take a quick look?

PR #373, opened 2019-06-10
added DFN-14_1.35x3.5mm_P0.5mm

Shackmeister:
datasheet https://m.littelfuse.com/~/media/electronics/datasheets/tvs_diode_arrays/littelfuse_tvs_diode_array_sp3012_datasheet.pdf.pdf#page=7


evanshultz:
All dimensions look good to me, but a question: should this be `uDFN-14...` to include the micro symbol Littelfuse uses? It might avoid a footprint name conflict later and perhaps is the best descriptor of the package.

evanshultz:
@Shackmeister 
Can you take a look at my comment above so we can get this, and the footprint, merged?

Shackmeister:
*removes spiderweb surrounding his git account*

evanshultz:
The unique dimensions are probably satisfactory. Will merge.

PR #374, opened 2019-06-10
Add SOP-24_7.5x15.4mm_P1.27mm

fauxpark:
Required for IS31FL3218.

Datasheet: http://www.issi.com/WW/pdf/31FL3218.pdf#page=14

fp


evanshultz:
Footprint is merged.

PR #376, opened 2019-06-11
Hirose df63

### LINKED ISSUE #375: Hirose DF63 1Pin connector missing NPTH
The `Hirose_DF63-1P-3.96DSA_1x01_P3.96mm_Horizontal` footprint does not include the NPT hole as the connectors with more pins do. This is erroneous. It should include it, as the [datasheet](https://www.hirose.com/product/document?clcode=CL0680-0529-2-00&productname=DF63-1618PCF&series=DF63&documenttype=Catalog&lang=en&documentid=D31622_en) (page 3), but also the following picture, show.
grafik

Additionally, all the `DF63` connectors currently in the library are named `Horizontal`, although they're actually the vertical parts. (their 'angled' counterparts would be the horizontal ones).

Would a patch be accepted, that adds the NPTH unconditionally (removing this [line](https://github.com/pointhi/kicad-footprint-generator/blob/4b4fcc8c5a8ee948747c915815fae72e64c40e91/scripts/Connector/Connector_Hirose/conn_hirose_df63_tht_top.py#L150))?

jhalmen:
fixes #375

### DIFF
--- a/scripts/Connector/Connector_Hirose/conn_hirose_df63_tht_top.py
+++ b/scripts/Connector/Connector_Hirose/conn_hirose_df63_tht_top.py
@@ -52,7 +52,7 @@ datasheet = 'https://www.hirose.com/product/en/products/DF63/'
 #pins_per_row per row
 pins_per_row_range = [1,2,3,4,5,6]
 
-#Molex part number
+#Hirose part number
 #n = number of circuits per row
 part_code = "DF63-{n:d}P-3.96DSA"
 
@@ -94,7 +94,7 @@ def generate_one_footprint(pins, configuration):
     footprint_name = footprint_name.replace("__",'_')
 
     kicad_mod = Footprint(footprint_name)
-    kicad_mod.setDescription("Molex {:s}, {:s}, {:d} Pins per row ({:s}), generated with kicad-footprint-generator".format(series_long, mpn, pins_per_row, datasheet))
+    kicad_mod.setDescription("Hirose {:s}, {:s}, {:d} Pins per row ({:s}), generated with kicad-footprint-generator".format(series_long, mpn, pins_per_row, datasheet))
     kicad_mod.setTags(configuration['keyword_fp_string'].format(series=series,
         orientation=orientation_str, man=manufacturer,
         entry=configuration['entry_direction'][orientation]))
--- a/scripts/Connector/Connector_Hirose/conn_hirose_df63_tht_top.py
+++ b/scripts/Connector/Connector_Hirose/conn_hirose_df63_tht_top.py
@@ -151,7 +151,7 @@ def generate_one_footprint(pins, form_type, configuration):
     )
 
     #mounting hole
-    if form_type == 'M':
+    if form_type in ['M', '']:
         peg_location = [-1.5, 3.25]
     else:
         peg_location = [B+1.5, 3.25]
@@ -257,4 +257,8 @@ if __name__ == "__main__":
 
     for pins_per_row in pins_per_row_range:
         for form_type in types:
+            if form_type is 'R' and pins_per_row is 6:
+                continue
+            if form_type is 'M' and pins_per_row in [5,6]:
+                form_type = ''
             generate_one_footprint(pins_per_row, form_type, configuration)
--- a/scripts/Connector/Connector_Hirose/conn_hirose_df63_tht_top.py
+++ b/scripts/Connector/Connector_Hirose/conn_hirose_df63_tht_top.py
@@ -96,7 +96,7 @@ def generate_one_footprint(pins, form_type, configuration):
     footprint_name = footprint_name.replace("__",'_')
 
     kicad_mod = Footprint(footprint_name)
-    kicad_mod.setDescription("Hirose {:s}, {:s}, {:d} Pins per row ({:s}), generated with kicad-footprint-generator".format(series_long, mpn, pins_per_row, datasheet))
+    kicad_mod.setDescription("Hirose {:s}, {:s}, {:d} Pins per row ({:s}), generated with kicad-footprint-generator".format(series_long, mpn, pins, datasheet))
     kicad_mod.setTags(configuration['keyword_fp_string'].format(series=series,
         orientation=orientation_str, man=manufacturer,
         entry=configuration['entry_direction'][orientation]))
--- a/scripts/Connector/Connector_Hirose/conn_hirose_df63_tht_top.py
+++ b/scripts/Connector/Connector_Hirose/conn_hirose_df63_tht_top.py
@@ -45,7 +45,7 @@ from footprint_text_fields import addTextFields
 series = ""
 series_long = 'DF63 through hole'
 manufacturer = 'Hirose'
-orientation = 'H'
+orientation = 'V'
 number_of_rows = 1
 datasheet = 'https://www.hirose.com/product/en/products/DF63/'
 
--- a/scripts/Connector/Connector_Hirose/conn_hirose_df63_tht_top.py
+++ b/scripts/Connector/Connector_Hirose/conn_hirose_df63_tht_top.py
@@ -147,8 +147,7 @@ def generate_one_footprint(pins, configuration):
     )
 
     #mounting hole
-    if pins > 1:
-        kicad_mod.append(Pad(at=[-1.5,3.25],type=Pad.TYPE_NPTH,layers=Pad.LAYERS_NPTH,shape=Pad.SHAPE_CIRCLE,size=mount_size,drill=mount_size))
+    kicad_mod.append(Pad(at=[-1.5,3.25],type=Pad.TYPE_NPTH,layers=Pad.LAYERS_NPTH,shape=Pad.SHAPE_CIRCLE,size=mount_size,drill=mount_size))
 
     #connector outline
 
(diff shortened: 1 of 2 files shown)


evanshultz:
Picking up the conversation from the earlier thread...

Line 55 also mentions the wrong manufacturer. Are you able to fix that and then regenerate footprints? The footprint name will stay the same but then we have the fixed 1-pin footprint ready to go. And ultimately this is all about having correct footprints in the library.

Lines 46 and 57 could have 'M' added to indicate the correct MPN, I think. I didn't try this so maybe more work is needed? The `series` variable may need to be corrected to fix the MPN. I'm happy you're willing to contribute, but let me know if you're not up for fixing the MPN because we don't want you to take such initiative to mention the issue and help out to fix it but then feel frustrated trying to do more.

If we can get this generating the correct M version it should be relatively easy to make it generate the R version too. Not having a suffix on the script name would make we assume all DF63 variants can be built. But that's a later step than just correcting the existing script. (I mention it as a note for myself, if nothing else.)

jhalmen:
The M is version is a new series that's still being introduced (see [this document](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=8&ved=2ahUKEwjllJXG5uTiAhXux6YKHUClCRYQFjAHegQIBRAC&url=https%3A%2F%2Fwww.hot-electronic.de%2Fmedia%2Fgetfile%2F79981ac0-3ffd-410b-93cb-2b8ebec00383%2F&usg=AOvVaw1lCD1LoPfOnSIgak2X76KN)), the footprints are the same as the normal DF63 series. So I'm unsure of the importance of adding the `M` to the series name.

As it stands now, the footprints for this series are generated correctly. I'm happy to work on it, but i don't have very much time currently. If you don't get to it, I will take a look at generating the R version too, tho.

evanshultz:
If DF63 and DF63M are the same, then, all is well. Making M and R versions doesn't make the existing footprints wrong (outside of orientation). Thanks for that extra info.

If you can submit a PR for the corrected footprints then I think can be wrapped up for now. While it's a breaking change to the library, the existing "horizontal" footprints are clearly wrong and should be "vertical", so we can remove them all and replace with the new ones you make. Does that sound OK to you?

jhalmen:
I looked over it again today, and also realized that the single pin connector was not wide enough, fixed that.
I also had a go at generating the R version of the footprint, and checked that it is correctly generated by comparing it to a 3D step file of the 2 Pin connector that can be downloaded from Hirose.
This only leaves the Horizontal footprints now.

Also a point of discussion should be the 3d models that get added to the footprints by the script. They seem not to exist on my installation. Should that part just be removed maybe?
Or are the 3d-models just behind, and they should be added instead?

If you're ok with the changes in the script, i'll gladly open the PR for the new footprints!

jhalmen:
I decided to go with the 'M' type description after all, because the type without designation is getting deprecated.

scripts/Connector/Connector_Hirose/conn_hirose_df63_tht_top.py
```diff
@@ -149,7 +149,11 @@ def generate_one_footprint(pins, configuration):
     )
 
     #mounting hole
-    kicad_mod.append(Pad(at=[-1.5,3.25],type=Pad.TYPE_NPTH,layers=Pad.LAYERS_NPTH,shape=Pad.SHAPE_CIRCLE,size=mount_size,drill=mount_size))
+    if form_type != 'R':
```
evanshultz: Since the M version is being specifically implemented, can you change this to `==`? That is more clear to read.

evanshultz:
Where is the location of the R version's peg location shown? I didn't check out any 3D models but I don't see it in the datasheet.

jhalmen:
In the series datasheet it actually doesn't specify clearly where the  peg is, it just says "guide post: right side" (vs. the M version "left side").
But it is stated in the drawings of the specific footprints. example: [DF63R-1P](https://www.hirose.com/product/document?clcode=CL0680-0517-3-00&productname=DF63R-1P-3.96DSA&series=DF63&documenttype=2DDrawing&lang=en&documentid=0000956173), [DF63R-2P](https://www.hirose.com/product/document?clcode=CL0680-0518-6-00&productname=DF63R-2P-3.96DSA&series=DF63&documenttype=2DDrawing&lang=en&documentid=0000956176) vs the [DF63M series](https://www.hirose.com/product/document?clcode=CL0680-0566-0-00&productname=DF63M-1P-3.96DSA&series=DF63&documenttype=2DDrawing&lang=en&documentid=0001262566)

evanshultz:
Thanks! This script looks great now.

evanshultz:
Not quite finished with this. The R version only goes up to 5 pins, not 6.

jhalmen:
now that you say that, i also realised that apparently the M version doesn't exist for all pin versions either.
the 5 and 6 pin footprints don't get the M designator.
i guess it would be good to fix that too?

evanshultz:
OK, I read the docs linked above and let me write this down. You wrote it above but bullets are easier for my little dinosaur brain to process:
- DF63 comes in 1-6 pins and is being phased out
- DF63M is the new replacement for the above and is the same but only available in 1-4 pins
- DF63R is a new part with reversed peg placement and is available in 1-5 pins

The script doesn't generate all these now, leading to possible maintenance issues. DF63M and R are built correctly, but the no-suffix footprints only come in 5-6 pins. This is only from looking at your code and not running it. Can you make the code generate all 3 types in all pin counts so that if there are other issues during KiCad's v5 release cycle we're covered already? Note that you don't have to do this, but it would be nice to be able to manage things nicely for the next year or two until v6 is released, and then remove the no-suffix footprints and be able to easily disable them from generation in this script.

evanshultz:
I guess this is a better summary so I'll repeat what I wrote at https://github.com/KiCad/kicad-footprints/pull/1649#issuecomment-504494869. It seems, at least from Digikey, that the no suffix 5 and 6 pin version will continue and DF63M only replaces no-suffix for 1-4 pins. I updated the comment above. Is that correct?

jhalmen:
the edited summary is correct, except that the R version is afaik not new. there was just no footprint for it in kicad. (as there are still no footprints for the 7.92mm pitch parts).
the current state of the script generates all vertical 3.96mm pitch parts that are not being phased out.

scripts/Connector/Connector_Hirose/conn_hirose_df63_tht_top.py
```diff
@@ -94,7 +96,7 @@ def generate_one_footprint(pins, configuration):
     footprint_name = footprint_name.replace("__",'_')
 
     kicad_mod = Footprint(footprint_name)
-    kicad_mod.setDescription("Molex {:s}, {:s}, {:d} Pins per row ({:s}), generated with kicad-footprint-generator".format(series_long, mpn, pins_per_row, datasheet))
+    kicad_mod.setDescription("Hirose {:s}, {:s}, {:d} Pins per row ({:s}), generated with kicad-footprint-generator".format(series_long, mpn, pins_per_row, datasheet))
```
evanshultz: `pins_per_row` can't be printed as a range and leads to a double space. I think `pins` is correct?

evanshultz:
I meant that the R version was a new offering from Hirose, not new to KiCad, but that wasn't clear. Seems I've finally got my head wrapped around it now. Thanks for the explanation.

I made one note about the script above. It doesn't substantially affect the footprints so I'll merge them as they are if you don't want to update them. But the script should be tidied up before merging.

scripts/Connector/Connector_Hirose/conn_hirose_df63_tht_top.py
jhalmen: `pins` is correct, but i don't see the double space you're mentioning in the generated footprints.
i changed it, because it's correct to use `pins`, but the generated footprints differ only in the edit time

evanshultz:
Odd. Maybe I was wrong. Here is the GH diff:
image

If there's still a double space, what is causing it?

jhalmen:
oh, now i see what you mean.
after a bit of digging, it's because the `series` variable is empty.
just a few lines before we do a `footprint_name.replace("__",'_')` because of the same reason, but we don't do the same for the footprint keywords (which are defined as `keyword_fp_string: 'connector {man:s} {series:s} {entry:s}'`)

i pushed a quick fix

jhalmen:
would this suffice?

evanshultz:
Got it. Thanks for chasing this down.

From https://github.com/KiCad/kicad-footprints/issues/1141, the lead librarian asked to keep the series name even if it's doubled. So then `series` should get a value of `DF63` and these ugly hacks can go away. Sorry that you got caught with an old script and this minor issue popped up.

Also, if there are more relevant docs perhaps add those links? Since this script does several different footprints with odd names it wouldn't hurt to have more info in the script instead of just in the comments here.

jhalmen:
I tried to cherry-pick that commit, not sure if i did everything right tho.

Is this what you had in mind?

evanshultz:
Thanks! The diff looks good. Can you push new footprints?

jhalmen:
pushed

evanshultz:
Thanks so much for cleaning this all up!

I think the DF63 footprints are good for now and I've made https://github.com/KiCad/kicad-footprints/issues/1663 to capture these (and other) footprints that need cleaned up.

@pointhi @poeschlr 
Footprints are in. This is ready to merge.

jhalmen:
thank you for being so thorough!

antoniovazquezblanco:
Ping @pointhi 

Thanks!

PR #380, opened 2019-06-15
Add readme to gullwing generator

poeschlr:
Work in progress


scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,100 @@
+# IPC gullwing generator
+
+This generator uses IPC-7351B equations and by default fillet definitions to derive a footprint for gullwing stile packages.
```
evanshultz: stile -> style

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,100 @@
...
+
+Every further entry in the script is assumed to be a package size definition. The top level parameter will give the internal parameter set name. This one must be unique in this file and should be representative of the footprint as it will be used in error messages. (It also makes later maintenance easier if it is easy to determine which footprint was genererated by which parameter set.)
+
+``` yaml
+internal_name:
+  # parameter list
+```
+
+The following parameters are supported for every package
+- Size source to be added to footprint documentation field (`size_source`) {url}
```
evanshultz: IMO it would be worth considering if the parameter name should go first and then the description; right now they're backwards

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,100 @@
...
+- footprint name control
+  - Manufacturer if specific to one manufacturer (`manufacturer`) {string}
+  - Part number if specific to one part (`part_number`) {string}
+  - Custom naming (`custom_name_format`) {format string}
+- Body size (`body_size_x`, `body_size_y`, `overall_height`) {dimension}
+- Lead dimensions:
+  - Over all size representing lead tip to lead tip dimension (`overall_size_x`, `overall_size_y`) {dimension}
```
evanshultz: 'Overall' is one word

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
evanshultz: 'url' here doesn't correspond to a data type, but 'string' and'int' and 'float' are used below and does.  'dimension' is used below and doesn't. Perhaps this needs to include data type and also a description, like 'string, url' or 'int/float, dimension'?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,100 @@
...
+  - Lead with (`lead_width`) {dimension}
```
evanshultz: missing 'd'

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,100 @@
...
+internal_name:
+  # parameter list
+```
+
+The following parameters are supported for every package
+- Size source to be added to footprint documentation field (`size_source`) {url}
+- footprint name control
+  - Manufacturer if specific to one manufacturer (`manufacturer`) {string}
+  - Part number if specific to one part (`part_number`) {string}
+  - Custom naming (`custom_name_format`) {format string}
```
evanshultz: For many of the parameters an example would be helpful, but on this one especially it would be useful to get a feel for when and how this is used because there are a few specific cases where it solves a particular problem. In fact, having an actual entry might be useful to point out what are required and what are optional (and when/where/why optional parameters are used).

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,100 @@
...
+- Body size (`body_size_x`, `body_size_y`, `overall_height`) {dimension}
+- Lead dimensions:
+  - Over all size representing lead tip to lead tip dimension (`overall_size_x`, `overall_size_y`) {dimension}
+  - Lead with (`lead_width`) {dimension}
+  - Lead length representing tip till bend (`lead_len`)
```
evanshultz: 'till' has an extra 'l'. As there are two lead bends perhaps a better wording would be 'distance lead is roughly parallel to the PCB, from the lead bend closest to the PCB surface to the lead tip'?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,100 @@
...
+- lead pitch currently equal for all sides (`pitch`) {float}
```
evanshultz: Maybe enclose 'currently equal for all sides' in parentheses?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,100 @@
...
+- pin count, 0 means no pins in that direction (`num_pins_x`, `num_pins_y`) {int}
+- Exposed pad:
+  - Size of package exposed pad or slug (`EP_size_x`, `EP_size_y`) {dimension}
+  - Optional the size of the footprint pad (`EP_size_x_overwrite`, `EP_size_y_overwrite`) {float}
+  - Optional the size of the mask cutout (`EP_mask_x`, `EP_mask_y`) {float}
+  - Paste can be split into regular grid with (`EP_num_paste_pads`) {[int (x), int (y)]}
+    - Paste coverage multiplicator (`EP_paste_coverage`) {float (0..1)}
```
evanshultz: Should 'multiplicator' be 'percentage'?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,100 @@
...
+  - Thermal vias (`thermal_vias`) with following sup parameters
```
evanshultz: Remove 's' in 'sup'.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,100 @@
...
+### String based
+
+The parameter can be given as a string of the format. Whitespace is ignored.
+
+```yaml
+parameter_name: 1.2 # nominal only (reference dimension marked as such in datasheet)
+parameter_name: 1.1 .. 1.2 .. 1.3 # min .. nominal .. max
+parameter_name: 1.1 .. 1.3 # min .. max
+parameter_name: 1.2 +/-0.1 # nominal plus symmetrical tolerance
+parameter_name: 1.2 +0.1 -0.05 # nominal plus asymmetrical tolerance
```
evanshultz: Is the plus required for the positive tolerance? (Just trying to think of potential loopholes or questions...)

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,100 @@
+# IPC gullwing generator
```
evanshultz: Thanks for taking this on!!!

- I don't see anywhere in here where it is specified that the datasheet dimensioning is to be used.
- I wonder if an image of an applicable package would be useful?
- Since many of the sections will apply to many of the generators, is there a way to have "master" sections of this doc which can be stitched together, with package-specific sections, to make a complete readme? Having the same content in a dozen files seems like a potential maintenance nightmare.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: I don't know what you mean here.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: percentage would imply a value between 0 and 100. This value is the result of dividing two values (The overall area divided by the paste area) without multiplying senselessly with 100.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: Introducing a ":" between the name and the description might be a more universal fix.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: The measurement in question is the one projected to the pcb plane. Meaning the bend angle tolerance is already part of its tolerance.

I think i will include a picture in the final version.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: url can be seen as an abstract datatype. (It is expected to be a string representing an url. Just because i do not check this for validity does not mean it is not expected to be an url)

poeschlr: Both have their merits. I prefer this style as it allows combining multiple parameters.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: Well i reduced it down to two generators so there are not really that many readmes that can share data in any useful way. (And the no lead style packages are way too different to merit the extra work for the little gain.)

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
evanshultz: Apparently I was looking word-by-word. What does 'sup' mean? Should 'following sup' change to 'the following'?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
evanshultz: Then perhaps mention it's possible use as a header for the entire YAML file?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
evanshultz: Then the word should be 'multiplier'.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -23,47 +23,60 @@ FileHeader:
...
+  - given as count value (`EP_num_paste_pads`) {[int (x), int (y)]}
+  - More specific as between vias and outside ring (`paste_between_vias`, `paste_rings_outside`)
 
 
 ## Dimension parameter format
 
 ### String based
 
-The parameter can be given as a string of the format. Whitespace is ignored.
+The parameter can be given as a string of one of the following formats (whitespace characters are ignored).
```
evanshultz: I think 'in one of' instead of 'of one of'.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -46,11 +46,14 @@ internal_package_name:
...
 - Size of package exposed pad or slug (`EP_size_x`, `EP_size_y`) {dimension}
-- Optional the size of the footprint pad, equal to nominal EP_size (`EP_size_x_overwrite`, `EP_size_y_overwrite`) {float}
+- Optional the size of the footprint pad (`EP_size_x_overwrite`, `EP_size_y_overwrite`) {float}
+   - Pad size is equal to nominal package pad size if not given.
 - Optional the size of the mask cutout (`EP_mask_x`, `EP_mask_y`) {float}
-- Paste can be split into regular grid with (`EP_num_paste_pads`) {[int (x), int (y)]}
-  - Paste coverage multiplicator (`EP_paste_coverage`) {float (0..1), default=0.65}
+   - Use to create solder mask defined pads (in combination with `EP_size_x_overwrite`)
+- Paste is typically split into a regular grid with (`EP_num_paste_pads`) {[int (x), int (y)]}
+  - The paste coverage multiplicator determines how much of the exposed copper area is covered by paste. (`EP_paste_coverage`) {float (0..1), default=0.65}
```
evanshultz: 'multiplicator' -> 'multiplier'

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,121 @@
+# IPC gullwing generator
+
+This generator uses IPC-7351B equations and by default fillet definitions to derive a footprint for gullwing stile packages.
```
evanshultz: 'stile' -> 'style'

evanshultz: What does 'by default' add here? Can it be removed? The capability to override IPC7351B fillet parameters would be best in a following sentence. Also, should this mention that the pad style defaults to rounded rectangle, which comes from preliminary IPC7251C documents.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,121 @@
...
+Examples of supported packages include: QFN, SOIC, SO
```
evanshultz: I think this needs a period at the end. Also, consider putting 'and' before 'SO'.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
evanshultz: Because you're mixing the content of the field and the field type it's messy.

The URL is a string with a specific format and so the `size_source` field takes a string type.

In addition, generally it will be package documentation and not footprint documentation (special footprint info may be included).

Also, adding a colon as you said to separate the field and info about it would be nice.

Maybe 'Size source: link to datasheet and optionally other package documentation (`size_source`) {string}'? Things could be reformatted to avoid duplication or near-duplication of info if the following format is adopted: '`size_source`: link to datasheet and optionally other package documentation {string}'.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,121 @@
...
+Every further entry in the script is assumed to be a package size definition. The top level parameter will give the internal parameter set name. This one must be unique in this file and should be representative of the footprint as it will be used in error messages. (It also makes later maintenance easier if it is easy to determine which footprint was genererated by which parameter set.)
+
+``` yaml
+internal_package_name:
+  # parameter list (See below)
+```
+
+### Documentation and naming parameters
+- Size source to be added to footprint documentation field (`size_source`) {url}
+- footprint name control
```
evanshultz: Capital 'F'?

evanshultz: And maybe just 'Footprint name' without the 'control'? I'm not understanding what that word adds.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Package dimensions
+![dimension example](./documentation/dimension_system.svg)
+- Body size (`body_size_x`, `body_size_y`, `overall_height`) {dimension}
+- Lead dimensions:
+  - Over all size representing lead tip to lead tip dimension (`overall_size_x`, `overall_size_y`) {dimension}
+  - Lead with (`lead_width`) {dimension}
+  - Lead length representing tip till bend (`lead_len`)
+- lead pitch, currently equal for all sides (`pitch`) {float}
+- pin count, 0 means no pins in that direction (`num_pins_x`, `num_pins_y`) {int}
```
evanshultz: Perhaps an example here, referencing the SOIC-looking image above to show when a zero is used?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Exposed pad handling:
+![exposed pad example](../documentation/ep_handling.svg)
+- Size of package exposed pad or slug (`EP_size_x`, `EP_size_y`) {dimension}
+- Optional the size of the footprint pad (`EP_size_x_overwrite`, `EP_size_y_overwrite`) {float}
+   - Pad size is equal to nominal package pad size if not given.
+- Optional the size of the mask cutout (`EP_mask_x`, `EP_mask_y`) {float}
+   - Use to create solder mask defined pads (in combination with `EP_size_x_overwrite`)
+- Paste is typically split into a regular grid with (`EP_num_paste_pads`) {[int (x), int (y)]}
+  - The paste coverage multiplicator determines how much of the exposed copper area is covered by paste. (`EP_paste_coverage`) {float (0..1), default=0.65}
```
evanshultz: Isn't this field optional, since a default value is shown? Maybe there's a better way to indicate optional parameters?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Thermal vias
+A package with exposed pad can generate a version with thermal vias. This will be generated in addition to the normal version.
+
+![exposed pad example](../documentation/thermal_vias.svg)
```
evanshultz: The image adds a lot of clarify but I have a few questions:
1. The image doesn't show the meaning of each item by color. Blue is copper. Pink is paste. Is the orange-ish color mask?
1. In the first row, what is creating the orange-ish round rect? Is it mask as I guessed above? What setting?
1. Similarly, the bottom two images show different padstack but the info below them in the same.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,121 @@
...
+- Final hole size (`drill`) {float}
+- Optional grid (`grid`) {[float (x), float (y)]}
+  - Auto generated if not given (outermost pad will touch pad edge.)
+- Optional paste coverage overwrite (`EP_paste_coverage`) {float (0..1)}
+  - Increase compared to non thermal via version to combat solder loss through wicking.
+- Should paste avoid vias? (`paste_avoid_via`) {bool}
+  - Can lead to math exceptions (reduce paste coverage)
+  - Clearance between via hole and paste controlled with (`paste_via_clearance`) {float}
+- Number paste pads
+  - given as count value (`EP_num_paste_pads`) {[int (x), int (y)]}
```
evanshultz: Isn't this optional, and by default a grid one less in X and Y than the via grid? If not, would that be a nice enhancement (as a general question, not a request of you).

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,121 @@
...
+    # thermal via version parameters
+```
+- Number of vias generated in the regular grid (`count`) {[int (x), int (y)]}
+- Final hole size (`drill`) {float}
+- Optional grid (`grid`) {[float (x), float (y)]}
+  - Auto generated if not given (outermost pad will touch pad edge.)
+- Optional paste coverage overwrite (`EP_paste_coverage`) {float (0..1)}
+  - Increase compared to non thermal via version to combat solder loss through wicking.
+- Should paste avoid vias? (`paste_avoid_via`) {bool}
+  - Can lead to math exceptions (reduce paste coverage)
```
evanshultz: Maybe add 'to fix' just before the closing parenthesis because one may not know this is offering a solution.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Dict based
+
+```yaml
+parameter_name: # nominal only
+  nominal: 1.2
+  tolerance: 0 # optional to make it clear that this is the case
+parameter_name: # minimum maximum and nominal
+  minimum: 1.1
+  nominal: 1.2 # is optional but should be included if given in the datasheet
```
evanshultz: I think 'is' in the comment can be removed.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,121 @@
...
+parameter_name: 1.2 +/-0.1 # nominal plus symmetrical tolerance
+parameter_name: 1.2 +0.1 -0.05 # nominal plus asymmetrical tolerance
+```
+
+### Dict based
+
+```yaml
+parameter_name: # nominal only
+  nominal: 1.2
+  tolerance: 0 # optional to make it clear that this is the case
```
evanshultz: It may be optional but what is the desired format of the YAML file?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,121 @@
...
+  - Can lead to math exceptions (reduce paste coverage)
+  - Clearance between via hole and paste controlled with (`paste_via_clearance`) {float}
+- Number paste pads
+  - given as count value (`EP_num_paste_pads`) {[int (x), int (y)]}
+  - More specific as between vias and outside ring (`paste_between_vias`, `paste_rings_outside`)
+
+
+## Dimension parameter format
+
+### String based
```
evanshultz: I think a hyphen is missing and should be `String-based`. Same as for the dict format below.

evanshultz:
The images add an incredible amount of clarity. Beautiful. I've made more comments above but this will be fantastic and very helpful.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: Any one that is not marked as deprecated. (So both the string and dict based one. The string based one simply is much less work to enter. Which is why i will use it should i add stuff in the future.)

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: No the number of paste pads must be given in either of the two formats (The grid for paste pads can not be specified.)

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: Clarified image

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: done

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: made it even clearer than your suggestion

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: removed this and added a much clearer description up front

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: done

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: changed to footprint name generation control and added more details

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: done

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: changed

poeschlr: fixed

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: fixed

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: fixed

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: I looked into it: github readme file format does not support includes.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: yes

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: Restructuring removed this

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: fixed

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: The custom naming format is now better documented. It is an advanced parameter that requires understanding of the python format string format. (the documentation should document how to use the script not how to use it for contributing to the official kicad library. I added notes regarding how it is used for library contributions where appropriate)

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: I document what the intention of the file format is. What it can be misused for is something that the user will need to check for themselfes (again the documentation is about the generator not how it is used for the official library.)

poeschlr: I am not sure how your suggestion fits with the source code line this is attached to.

poeschlr:
@evanshultz i think i have addressed everything you suggested. Is this now good enough for a first iteration of the documentation?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,151 @@
...
+
+This generator uses IPC-7351B equations and fillet definitions to derive a footprint for gullwing style packages. (fillet definitions can be changed by pointing to a different definition file.) The script generates footprints in zero orientation version A (pin 1 at top left corner).
+Pads are generated using rounded rectangle pads as suggested in preliminary releases of IPC-7351C.
+
+Examples of supported packages include: QFN, SOIC, SO.
+
+## Running the script
+
+The script requires python 3.4 or newer. Run it with:
+`python3 ipc_gullwing_generator.py size_definitions/eqfp.yaml` (replace eqfp.yaml with the size definition file that contains your part.)
```
evanshultz: Can you enclose the file name within the parentheses in code to match the text before the parentheses?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,151 @@
...
+### Documentation and naming parameters
+- Size source to be added to footprint documentation field (`size_source`) {url}
+- Footprint name generation control
+  - Manufacturer name. Will be added as a prefix if given. (`manufacturer`) {string}
+  - Part number. Will be added as a prefix if given. (`part_number`) {string}
+  - suffix: A custom suffix (added after pin pitch in default naming format). Can include parameters pad_x and pad_y.
+  - Custom naming (`custom_name_format`) {python format string}
+    - The full default format string is '{man:s}_{mpn:s}_{pkg:s}-{pincount:d}-1EP_{size_x:g}x{size_y:g}mm_P{pitch:g}mm{suffix:s}_EP{ep_size_x:g}x{ep_size_y:g}mm_Mask{mask_size_x:g}x{mask_size_y:g}mm{suffix2:s}{vias:s}' (The same parameters can be used in your custom format. Exposed pad parameters are not available for components without exposed pad.)
+
+_Note: Contributions intended for the official library shall not include the manufacturer or part number unless the footprint is specific to that manufacturer or part. Similarly avoid custom naming for official library contributions unless required to achieve the requested name (Example TI specific naming)._
```
evanshultz: Should be `TI-specific`, with a hyphen.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,151 @@
...
+internal_package_name:
+  # parameter list (See below)
+```
+
+### Documentation and naming parameters
+- Size source to be added to footprint documentation field (`size_source`) {url}
+- Footprint name generation control
+  - Manufacturer name. Will be added as a prefix if given. (`manufacturer`) {string}
+  - Part number. Will be added as a prefix if given. (`part_number`) {string}
+  - suffix: A custom suffix (added after pin pitch in default naming format). Can include parameters pad_x and pad_y.
```
evanshultz: Should "suffix" be capitalized like the other parameters? And should 'pad_x' and 'pad_y' be shown as code?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
evanshultz: See above. Not changed.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
evanshultz: Also not addressed.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
evanshultz: The typo with the extra 'l' was not fixed.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,151 @@
...
+![dimension example](./documentation/dimension_system.svg)
+- Body size (`body_size_x`, `body_size_y`, `overall_height`) {dimension}
+- Lead dimensions:
+  - Over all size representing lead tip to lead tip dimension (`overall_size_x`, `overall_size_y`) {dimension}
+  - Lead with (`lead_width`) {dimension}
+  - Lead length representing tip till bend (`lead_len`)
+- lead pitch, currently equal for all sides (`pitch`) {float}
+- pin count (`num_pins_x`, `num_pins_y`) {int}
+  - num_pins_x=0 is used for generating SOIC like packages.
+  - num_pins_y=0 is used to generat SOIC like package footprints but with inverted pin numbering (Mirrored numbering scheme. Some manufactures use this style in their datasheets. Make sure you are not looking at the bottom view before using this. Not supported for QFP and similar.)
```
evanshultz: Typo of 'generate'. (Missing the 'e'.)

evanshultz:
@poeschlr 
I think I caught some minor typos when re-reading just now.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: What typo do you mean exactly?

poeschlr:
I marked the stuff as resolved that is done to clean up here a bit (got quite hard to see what still needs to be done or answered)

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
evanshultz: The general informality of the statement; perhaps change `lead_len` description to `Distance between the lead bend and lead tip`?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
evanshultz: Did I miss anywhere that states to get the dimensional information from a datasheet or package drawing? You recall a recent PR where the contributor tried to use the footprint dimensions.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,151 @@
+# IPC gullwing generator
+
+This generator uses IPC-7351B equations and fillet definitions to derive a footprint for gullwing style packages. (fillet definitions can be changed by pointing to a different definition file.) The script generates footprints in zero orientation version A (pin 1 at top left corner).
```
evanshultz: Missing a capital 'F' in the sentence inside the parens.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,151 @@
...
+Pads are generated using rounded rectangle pads as suggested in preliminary releases of IPC-7351C.
+
+Examples of supported packages include: QFN, SOIC, SO.
```
evanshultz: QFN is an example of a gullwing package? Is this a typo? Maybe examples could include `SOIC, TSSOP, HTSSOP, etc.` to capture more package types that are supported?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Running the script
+
+The script requires python 3.4 or newer. Run it with:
+`python3 ipc_gullwing_generator.py size_definitions/eqfp.yaml` (replace `eqfp.yaml` with the size definition file that contains your part.)
+
+### Optional script parameters
+
+* --global_config: the config file defining how the footprint will look like. (KLC)' (default='../../tools/global_config_files/config_KLCv3.0.yaml')
```
evanshultz: The first two descriptions start with 'the' but the last three don't. I suggest removing the 'the' after the colon.

Also, what is the apostrophe doing after `(KLC)`?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,151 @@
...
+* --series_config: the config file defining series parameters (footprint naming). (default='../package_config_KLCv3.yaml')
+
+* --density: Density level (L,N,M) (default='N')
```
evanshultz: Should this mention it's the IPC density level and not some other generic density parameter? The parameter below, for example, explicitly mentions the relevance to IPC even though the parameter name also includes this word.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,151 @@
...
+```
+
+### Documentation and naming parameters
+- Size source to be added to footprint documentation field (`size_source`) {url}
+- Footprint name generation control
+  - Manufacturer name. Will be added as a prefix if given. (`manufacturer`) {string}
+  - Part number. Will be added as a prefix if given. (`part_number`) {string}
+  - Suffix: A custom suffix (added after pin pitch in default naming format). Can include parameters `pad_x` and `pad_y`.
+  - Custom naming (`custom_name_format`) {python format string}
+    - The full default format string is '{man:s}_{mpn:s}_{pkg:s}-{pincount:d}-1EP_{size_x:g}x{size_y:g}mm_P{pitch:g}mm{suffix:s}_EP{ep_size_x:g}x{ep_size_y:g}mm_Mask{mask_size_x:g}x{mask_size_y:g}mm{suffix2:s}{vias:s}' (The same parameters can be used in your custom format. Exposed pad parameters are not available for components without exposed pad.)
```
evanshultz: Should this string be enclosed in code marker since the underscores make the mpn section appear in italiac and mess up the intended rendering of the format?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,151 @@
...
+_Note: Contributions intended for the official library shall not include the manufacturer or part number unless the footprint is specific to that manufacturer or part. Similarly avoid custom naming for official library contributions unless required to achieve the requested name (Example TI-specific naming)._
+
+### Package dimensions
+![dimension example](./documentation/dimension_system.svg)
+- Body size (`body_size_x`, `body_size_y`, `overall_height`) {dimension}
+- Lead dimensions:
+  - Overall size representing lead tip to lead tip dimension (`overall_size_x`, `overall_size_y`) {dimension}
+  - Lead width (`lead_width`) {dimension}
+  - Lead length representing tip till bend (`lead_len`)
+- lead pitch, currently equal for all sides (`pitch`) {float}
```
evanshultz: This and the next bullet start with a lowercase letter but the two above it user uppercase. Capitalize them?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,151 @@
...
+- Footprint name generation control
+  - Manufacturer name. Will be added as a prefix if given. (`manufacturer`) {string}
+  - Part number. Will be added as a prefix if given. (`part_number`) {string}
+  - Suffix: A custom suffix (added after pin pitch in default naming format). Can include parameters `pad_x` and `pad_y`.
+  - Custom naming (`custom_name_format`) {python format string}
+    - The full default format string is '{man:s}_{mpn:s}_{pkg:s}-{pincount:d}-1EP_{size_x:g}x{size_y:g}mm_P{pitch:g}mm{suffix:s}_EP{ep_size_x:g}x{ep_size_y:g}mm_Mask{mask_size_x:g}x{mask_size_y:g}mm{suffix2:s}{vias:s}' (The same parameters can be used in your custom format. Exposed pad parameters are not available for components without exposed pad.)
+
+_Note: Contributions intended for the official library shall not include the manufacturer or part number unless the footprint is specific to that manufacturer or part. Similarly avoid custom naming for official library contributions unless required to achieve the requested name (Example TI-specific naming)._
+
+### Package dimensions
```
evanshultz: Perhaps the into to this section should mention the format for dimensions is given below? Especially since the format is quite specific and can take several forms.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
evanshultz: Also, this line does not end in the data type in curved brackets.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,151 @@
...
+![dimension example](./documentation/dimension_system.svg)
```
evanshultz: The image shows `lead_length` while the parameter name is `lead_len`.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,151 @@
...
+- pin count (`num_pins_x`, `num_pins_y`) {int}
+  - num_pins_x=0 is used for generating SOIC like packages.
+  - num_pins_y=0 is used to generate SOIC like package footprints but with inverted pin numbering (Mirrored numbering scheme. Some manufactures use this style in their datasheets. Make sure you are not looking at the bottom view before using this. Not supported for QFP and similar.)
+
+### Exposed pad handling:
+![exposed pad example](../documentation/ep_handling.svg)
+- Size of package exposed pad or slug (`EP_size_x`, `EP_size_y`) {dimension}
+- Optional the size of the footprint pad (`EP_size_x_overwrite`, `EP_size_y_overwrite`) {float}
+   - Pad size is equal to nominal package pad size if not given.
+   - Use this to create a soldermask defined pad.
```
evanshultz: 'soldermask-defined' should have a hyphen.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,151 @@
...
+  - Lead length representing tip till bend (`lead_len`)
+- lead pitch, currently equal for all sides (`pitch`) {float}
+- pin count (`num_pins_x`, `num_pins_y`) {int}
+  - num_pins_x=0 is used for generating SOIC like packages.
+  - num_pins_y=0 is used to generate SOIC like package footprints but with inverted pin numbering (Mirrored numbering scheme. Some manufactures use this style in their datasheets. Make sure you are not looking at the bottom view before using this. Not supported for QFP and similar.)
+
+### Exposed pad handling:
+![exposed pad example](../documentation/ep_handling.svg)
+- Size of package exposed pad or slug (`EP_size_x`, `EP_size_y`) {dimension}
+- Optional the size of the footprint pad (`EP_size_x_overwrite`, `EP_size_y_overwrite`) {float}
```
evanshultz: I still don't like starting this with 'Optional the'. It breaks the format above and just doesn't read well even though the meaning is probably understandable. Maybe change the text before the parens to `Size of the footprint pad [optional]`? Also below.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,151 @@
...
+![exposed pad example](../documentation/thermal_vias.svg)
+
+``` yaml
+  thermal_vias:
+    # thermal via version parameters
+```
+- Number of vias generated in the regular grid (`count`) {[int (x), int (y)]}
+- Final hole size (`drill`) {float}
+- Optional grid (`grid`) {[float (x), float (y)]}
+  - Auto generated if not given (outermost pad will touch pad edge.)
```
evanshultz: Period should be after the closing paren.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,151 @@
...
+- Optional paste coverage overwrite (`EP_paste_coverage`) {float (0..1)}
+  - Used to combat solder wicking (increase compared to non thermal vias version).
```
evanshultz: I had to read this a couple of times to grasp it what I think it's saying. Maybe re-word to `Increase in paste coverage relative to the non-thermal via footprint to combat solder wicking`?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,151 @@
...
+- Should paste avoid vias? (`paste_avoid_via`) {bool}
```
evanshultz: Since this is the only question I'd change the wording to make it a statement and avoid the question mark.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,151 @@
...
+  - Clearance between via hole and paste controlled with (`paste_via_clearance`) {float}
```
evanshultz: I think you can remove the phrase `controlled with `.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,151 @@
...
+  - Can lead to math exceptions. Possible fixes:
+     - Reduce paste coverage (make sure you still have enough paste),
```
evanshultz: Remove trailing comma.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,151 @@
...
+     - Play with the via grid and number of paste pads (having an outer ring of paste pads often helps. This is only possible if there is space on the outside)
```
evanshultz: Perhaps change `on the outside` to `for paste pads outside the outermost via grid`.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,151 @@
...
+     - If no fix is satisfactory then select avoid vias as false and increase paste coverage to combat solder loss.
+- Number paste pads
+  - given as count value (`EP_num_paste_pads`) {[int (x), int (y)]}
+  - More specific as between vias and outside ring (`paste_between_vias`, `paste_rings_outside`)
```
evanshultz: The description sounds more clear to me as `Specify paste pad locations`. Also, what is the data type? (I know bool, but it's not written here.)

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,151 @@
...
+- Optional paste coverage overwrite (`EP_paste_coverage`) {float (0..1)}
+  - Used to combat solder wicking (increase compared to non thermal vias version).
+- Should paste avoid vias? (`paste_avoid_via`) {bool}
+  - Clearance between via hole and paste controlled with (`paste_via_clearance`) {float}
+  - Can lead to math exceptions. Possible fixes:
+     - Reduce paste coverage (make sure you still have enough paste),
+     - Play with the via grid and number of paste pads (having an outer ring of paste pads often helps. This is only possible if there is space on the outside)
+     - If no fix is satisfactory then select avoid vias as false and increase paste coverage to combat solder loss.
+- Number paste pads
+  - given as count value (`EP_num_paste_pads`) {[int (x), int (y)]}
```
evanshultz: This description could simply be `Quantity of paste pads`.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,151 @@
...
+  - More specific as between vias and outside ring (`paste_between_vias`, `paste_rings_outside`)
+
+
+## Dimension parameter format
+Dimensions in datasheets are either given with minimum and maximum value (optionally including nominal) or with the nominal value plus a tolerance. Some values of the datasheet are given as reference value without tolerance. (Tolerances in this measurement are already included in other dimensions.) The script reflects this by offering the same options.
+
+Always include the nominal dimension if the tolerance is asymmetrical as the resulting footprint will differ if it is not included.
+
+_Note: Contributions that are intended for the official kicad library must use the same dimensioning format as the datasheet. (If min, nom and max are given then all 3 must be entered into the yaml file even if the tolerance is symmetrical. Similarly use nominal plus tolerance format if the datasheet is dimensioned this way.)_
```
evanshultz: Capitalize `KiCad`.

evanshultz: And `YAML`.

evanshultz:
I've re-read and made several semantic comments. I know you want to get this merged but I suspect if polish isn't proposed now it has a worse chance of being included later.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
tolot27: Is it really necessary to put the sentence regarding to fillet definitions in brackets?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: typo, should have been qfp

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: copied directly from python so the ' is there because it is a string in python.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: good point (both here and in python)

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: fixed

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: added

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: both fixed

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: fixed

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: done

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: fixed

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: fixed

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: fixed

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: done

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: done

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: Made it clearer, did not use your suggestion as i feel your suggestion makes it seem as if this parameter is a multiplier of the normal parameter.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: fixed

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: done

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: done

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: clarified it

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: The fact that you misunderstood this setting shows i needed to find a better wording, hopefully done.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: fixed

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: done

poeschlr:
helvetica is a sans-serif font but non free meaning not available in inkscape. See for example https://superuser.com/questions/94992/is-it-possible-to-install-the-helvetica-font-on-linux

This is why i instead simply set the font to sans-serif which lets the browser decide which one to use.

tolot27:
> This is why i instead simply set the font to sans-serif which lets the browser decide which one to use.

Changing it from `font-family:Sans` to `font-family:sans-serif` did the trick. Now, browsers renders them correctly as a serif-less font and it is much more readable now. Thx.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
evanshultz: Were you going to make any change? Such as changing this parameter's description to `IPC 7351 density level`?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
```diff
@@ -0,0 +1,153 @@
...
+
+_Note: Contributions intended for the official library shall not include the manufacturer or part number unless the footprint is specific to that manufacturer or part. Similarly avoid custom naming for official library contributions unless required to achieve the requested name (Example TI-specific naming)._
+
+### Package dimensions
+![dimension example](./documentation/dimension_system.svg)
+- Body size (`body_size_x`, `body_size_y`, `overall_height`) {dimension}
+- Lead dimensions:
+  - Overall size representing lead tip to lead tip dimension (`overall_size_x`, `overall_size_y`) {dimension}
+  - Lead width (`lead_width`) {dimension}
+  - Lead length Distance between the lead bend and lead tip (`lead_len`) {dimension}
```
evanshultz: The capital 'D' should be lowercase or else add a colon before it.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
evanshultz: Now the sentence inside the parens is captalized, but it's not a stand-alone sentence. Add a period to close the preceding sentence and move the period for the paren sentence inside the paren or else don't capitalize the phrase inside the parents. For clarity, the first solution looks like: `This generator uses IPC-7351B equations and fillet definitions to derive a footprint for gullwing style packages from the dimensions of the package. (The suggested footprint is only used to determine the size of the exposed pad.)`

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
evanshultz: This looks good, but also make the same change on line 74.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
evanshultz: Got it. I did misunderstand and your new text is clear. Thanks. It does need `non via` changed to `non-via` (with a hyphen) and `losed by` changed to `lost due to`.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
evanshultz: This sounds nice. Just change `setup` to `set up`.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
evanshultz: Indeed I did. This is nice. Should 'paste avoid via' be `paste_avoid_via` to show it's referring specifically to the parameter?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: done

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md
poeschlr: I am not so sure about that. After all it should be clear to anyone that additional info is given below and it would kind of destroy the flow here.

evanshultz:
OK. No further changes requested by me.

But one question: The legacy and dict-based dimension formats allow specifying just the max, but I don't see that possibility in the string-based dimension format. Is that possible? If not, I suppose you would like to include that capability and documentation in a separate PR but I'm asking now in case it's already there. As you know, only max is often used for package height.

poeschlr:
The string based does not allow to have max only. I do not even have an idea how it could be done.
Also max only is only allowed for the overall height parameter which is not used at all right now (only prepared for some possible future where we use this data set as the input for the 3d model generation)

poeschlr:
@pointhi might it be a good idea to give evan merge rights like you gave to me? (This project is too important to be left to a single person)

pointhi:
@poeschlr I'm fine with adding @evanshultz as maintainer

pointhi:
@evanshultz I sent an invite

PR #381, opened 2019-06-15
Add script for Microchip TFBGA, 14 x 14 196 pin, 11 x 11mm package, 0.75 mm pitch, 0.32 mm ball diameter

yankee14:
[Associated symbol PR](https://github.com/KiCad/kicad-symbols/pull/1866)
[Associated footprint PR](https://github.com/KiCad/kicad-footprints/pull/1604)
[Associated abandoned PR that I screwed up because I don't know how to use git, apparently](https://github.com/pointhi/kicad-footprint-generator/pull/364)

Screencap showing pad diameter:
TFBGA-196_pad_diameter

Screencap showing mask diameter. KiCad's measurement tool rounded off the decimals; it is 0.275 mm:
TFBGA-196_mask_diameter

@evanshultz https://github.com/pointhi/kicad-footprint-generator/pull/364#discussion_r292211562
> Add "#page=2956" onto the end of the link to just right to this package's page

Done!

@evanshultz https://github.com/pointhi/kicad-footprint-generator/pull/364#discussion_r292211787
> The ball diameter does not need to be specified, I don't think. What was the reason to add it?

Microchip's ball diameter/pad diameter pairing diverges from IPC recommendations. If you think it is not necessary, I will remove the text. See IPC chart in [this engineering stackexchange post](https://electronics.stackexchange.com/a/165241).

@evanshultz https://github.com/pointhi/kicad-footprint-generator/pull/364#issuecomment-500614143
> Also, would you please add the package height of 1.2mm

Done!

@evanshultz https://github.com/pointhi/kicad-footprint-generator/pull/364#issuecomment-500614143
> I made a couple comments inline.

I think those comments are lost because I screwed up that PR. I'm sorry.

### DIFF
--- a/scripts/Packages/Package_BGA/bga.yaml
+++ b/scripts/Packages/Package_BGA/bga.yaml
@@ -128,7 +128,7 @@ TFBGA-196_11x11mm_Layout14x14_P0.75mm_Ball0.32mm_Pad0.35mm_SMD:
   overall_height: 1.20
   pitch: 0.75
   pad_diameter: 0.35
-  mask_margin: 0.0375
+  mask_margin: -0.0375
   paste_margin: 0.000001
   layout_x: 14
   layout_y: 14
--- a/scripts/Packages/Package_BGA/bga.yaml
+++ b/scripts/Packages/Package_BGA/bga.yaml
@@ -120,7 +120,7 @@ TFBGA-121_10x10mm_Layout11x11_P0.8mm:
   paste_margin: 0.000001
   layout_x: 11
   layout_y: 11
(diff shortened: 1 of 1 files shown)


evanshultz:
Wrong page number. We don't need the ball diameter here.

yankee14:
The other TFBGA definitions also omit the Pad and SMD fields from the name. Should I remove those?

evanshultz:
The pad size, yes, I suppose can be removed. (I'm sure we will run into problems later because of under-specification, but let's stick with our current paradigm so fixing it has only one process for the whole library whenever that happens.)

The `_SMD` needs to stay because this footprint has solder mask defined pads and this is atypical in our library (and against best practices as documented online and in my experience).

PR #383, opened 2019-06-16
Updated dfn.yaml script with TDFN-6 footprints

TiZed:
Added two TDFN-6 footprints, one with thermal vias and another without.


poeschlr:
Thanks

PR #384, opened 2019-06-18
TE 826576

evanshultz:
https://www.te.com/commerce/DocumentDelivery/DDEController?Action=srchrtrv&DocNm=826576&DocType=Customer+Drawing&DocLang=English

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1652


poeschlr:
Note to self: y size is currently wrong (datasheet self-contradictory) see footprint pull request.

evanshultz:
@poeschlr 
Script updated. I tried to capture the issue and reason to use the datasheet table values and not the equation in the script for the future.

poeschlr:
thanks

PR #385, opened 2019-06-20
add LFCSP-20-1EP-4x4mm_P0.5mm_EP2.6x2.6mm

lorem-ipsum:
This adds the size definition for another LFCSP footprint from Analog. In their naming it is CP-20-8 ([Datasheet](https://www.analog.com/media/en/package-pcb-resources/package/pkg_pdf/lfcspcp/cp-20/CP_20_8.pdf))

### DIFF
--- a/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/lfcsp.yaml
+++ b/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/lfcsp.yaml
@@ -180,7 +180,7 @@ LFCSP-20-1EP_4x4mm_P0.5mm_EP2.6x2.6mm:
     EP_num_paste_pads: [2, 2]
     #paste_between_vias: 1
     #paste_rings_outside: 1
-    EP_paste_coverage: 0.75
+    EP_paste_coverage: 0.70
     #grid: [1, 1]
     # bottom_pad_size:
     paste_avoid_via: False
--- a/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/lfcsp.yaml
+++ b/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/lfcsp.yaml
@@ -10,11 +10,11 @@ LFCSP-16-1EP_3x3mm_P0.5mm_EP1.6x1.6mm:
   body_size_x:
     minimum: 2.9
     nominal: 3
-    minimum: 3.1
+    maximum: 3.1
   body_size_y:
     minimum: 2.9
     nominal: 3
-    minimum: 3.1
+    maximum: 3.1
   body_height:
     minimum: 0.7
     nominal: 0.75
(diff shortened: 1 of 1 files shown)


scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/lfcsp.yaml
```diff
@@ -128,6 +128,73 @@ LFCSP-16-1EP_3x3mm_P0.5mm_EP1.7x1.7mm:
...
+  #manufacturer: 'man'
+  #part_number: 'mpn'
+  size_source: 'https://www.analog.com/media/en/package-pcb-resources/package/pkg_pdf/lfcspcp/cp-20/CP_20_8.pdf'
+  ipc_class: 'qfn' # 'qfn_pull_back'
+  #ipc_density: 'least' #overwrite global value for this device.
+  # custom_name_format:
+  body_size_x:
+    minimum: 3.9
+    nominal: 4
+    minimum: 4.1
```
evanshultz: Should be `maximum`, and also the one below

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/lfcsp.yaml
```diff
@@ -128,6 +128,73 @@ LFCSP-16-1EP_3x3mm_P0.5mm_EP1.7x1.7mm:
...
+
+  thermal_vias:
+    count: [3, 3]
+    drill: 0.3
+    # min_annular_ring: 0.15
+    paste_via_clearance: 0.1
+    EP_num_paste_pads: [2, 2]
+    #paste_between_vias: 1
+    #paste_rings_outside: 1
+    EP_paste_coverage: 0.75
```
evanshultz: This is a bit high. Usually we will target ~70%. If the paste isn't over the vias, this can be less. (75% is probably OK but if you're changing other items this should be fine to be a bit lower or just comment out and accept the default.)

evanshultz:
Comments inline above.

Since the YAML file is for LFCSP packages, this really should have the following at the top of the file and not in each entry:
```
 device_type: 'LFCSP'
 library: Package_CSP
```

But it's OK and this can be done later.

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/lfcsp.yaml
lorem-ipsum: The error was already present in the entries above (which I copied as a start). Fixed it in both

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/lfcsp.yaml
lorem-ipsum: reduced it as suggested. The number was also taken from the entries above.

lorem-ipsum:
I did add the definition to the begining of the file, but removed it again as the generator aparently does not handle it (yet). Had a quick look whether there is a simple solution, but couldn't find any.

evanshultz:
Maybe it's a different generator that handles this. I could be confused. Thank you for correcting the other footprint definitions. Anyway, I think this is good to go. Thanks!

@pointhi @poeschlr 
This is ready to go.

poeschlr:
The no lead and gullwing generators have slightly different yaml handling. I know i should abstract this but the list of things i need to do is getting longer instead of shorter so it is something everyone needs to either live with or fix themselves.

evanshultz:
@poeschlr 
Got it. Ready to merge then.

Issue #134, opened 2018-06-18
IDC connector footprints


evanshultz:
I have a few questions about the IDC header footprints:
1. Is there a script to generate them? An example footprint is `IDC-Header_2x13_P2.54mm_Vertical_Lock`.
1. The vertical footprints have large THT pads which aren't in the [referenced datasheet](https://www.reboul.fr/storage/00003af6.pdf).
1. The vertical footprints don't show the latch in the open position, which means any required keepout for any tall components or other elements near the PCB can't be easily found. I've seen this shown in silk on every PCB using this type of header I've ever seen.
![image](https://user-images.githubusercontent.com/1936989/41556572-0e9014fa-72f0-11e8-998e-2b546db54e53.png)
1. The horizontal header is potentially even worse because no components can lay on the PCB (or near the edge if the connector doesn't hang totally off) and the open position of the latches aren't shown.
1. The horizontal is called `Lock` but it doesn't include latches in the drawing but only the shrouded header that can optionally support latches!
![image](https://user-images.githubusercontent.com/1936989/41556615-296a6e7e-72f0-11e8-9c69-bbcfded8a800.png)
1. Latches come in various sizes, so the vertical and horizontal footprints may need a couple different footprints to cover short and long latches. Even if they vary a little bit, certainly a reasonable generic footprint could be created.
1. Using `Lock` seems like a poor word choice. Locking can be achieved with a ramp or other means for a similar connector, but latches are more specific to this type of lock. In other words, latches seem to me to be a subcategory of locks and we could be more specific in our footprint names.
1. Would it be best to reference a standard for these connectors? DIN 41651, I believe, is the standard for these connectors (but I don't have it to review).

evanshultz:
Anybody?

evanshultz:
Nobody? On any item above?

evanshultz:
@Shackmeister @terjeio 
Do either of you know anything about the footprints I've mentioned above?

matthijskooijman:
The scripts at https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Multicomp might also be relevant. These generate footprints for the multicomp MC9A12 and MC9A22, which are 2.54mm vertical and horizontal shrouded pin headers. The footprints are different from the ones in the Connector_IDC library, so this is probably not the same script. I'm not sure whether the script is a good one to base upon in terms of use of the framework (but since it is three years old, it might not).

evanshultz:
[...]

PR #387, opened 2019-06-21
Add support for IDC headers

evanshultz:
- Also added some missing variables
- Only those with mounting holes are included as current clipping tools are inadequate for non-symmetrical footprints (about pin 1 placement) with lines that could intersect a non-rectangular object

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1586


evanshultz:
Added some other headers which existed before but the generator didn't make. I used a very simple method to avoid clipping pads with silk since the existing 'clipper' functions can't handle a polyline that is intersected by multiple objects. Screenshots of these footprints, along with all others, are at the footprint PR above.

evanshultz:
@poeschlr 
Here is the script PR for the IDC footprints that are now merged. https://github.com/pointhi/kicad-footprint-generator/issues/134 is an issue about it.

poeschlr:
thanks

PR #389, opened 2019-06-25
JST VH script update

evanshultz:
only VH vertical no-shroud script updated, and not others
-On page 4 of the datasheet http://www.jst-mfg.com/product/pdf/eng/eVH.pdf you can see there are many flavors of pin-omitted headers
-B2P3-VH is popular as an AC connector where earth ground is not required so I wanted to generate it
-Support for all post-omitted versions is provided and I've checked the 5 examples in the datasheet

Also:
-Improved maintenance in conn_jst_vh_tht_top.py with variable reuse
-Removed comment which doesn't seem to be correct (datasheet update??) in conn_jst_VH_tht_top-shrouded.py, and had a typo anyway.

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1662


scripts/Connector/Connector_JST/conn_jst_vh_tht_top.py
```diff
@@ -149,11 +156,28 @@ def generate_one_footprint(pins, series_params, configuration):
     else:
         optional_pad_params['tht_pad1_shape'] = Pad.SHAPE_ROUNDRECT
 
-    kicad_mod.append(PadArray(
-        pincount=pins, x_spacing=pitch,
-        type=Pad.TYPE_THT, shape=shape,
-        size=pad_size, drill=drill, layers=Pad.LAYERS_THT,
-        **optional_pad_params))
+    if post_omitted:
```
poeschlr: Why not use the exclude pin parameter of the pad array?

https://github.com/pointhi/kicad-footprint-generator/blob/f84ae65fdbc720440412c36a69a591330ae5bc50/KicadModTree/nodes/specialized/PadArray.py#L107

evanshultz:
Simply because I didn't even recall this method existing. Thanks for mentioning it. Am I required to change this script or can it be merged as is?

evanshultz:
@poeschlr 
Are you requiring any change or just asking a question?

poeschlr:
I would really prefer if you could rewrite this such that it uses the exclude pin feature already present in the array class.

evanshultz:
@poeschlr 
Commit pushed. Is this OK? The resulting footprints do not change so there is no commit at that PR.

poeschlr:
You might want to check again. The shape of pad 1 changes for me (should now be rounded rectangle instead of circle -> would fix my request on the footprint pull request)

This is how B2P3-VH looks to me when running the generator:
Screenshot from 2019-09-27 22-11-00

evanshultz:
Ah. OK. I missed that because the new output of the script looked good.

Using the exclude_pin_list results in non-contiguous pins, like you can see above (only pins 1 and 3 with 2 missing). Is that OK with you? Previously the new footprint I contributed had pins 1 and 2. Do you have a preference? Using only contiguous pins seems like the best way to me since the user picks a symbol that has the number of pins in the physical part.

poeschlr:
Yea you are probably right that using the numbering as it is now might require quite an insane number of symbols. The downside will be that the symbol pin number might not agree with what is stamped onto the housing (as i guess you still buy the normal 3 pin housing and just leave pin 2 without crimp insert)

evanshultz:
@poeschlr 
Right, you get the existing plastic housing with the metal pins simply not present.

So undo the last change I made but make pin 1 a round rect? Will that be good?

evanshultz:
@poeschlr 
I undid the last thing I did, going back to a for loop to generate pads if some pins are omitted, and I added the logic to make pin 1 a rect or roundrect (which I should have done from the beginning but forgot about). Is this OK now?

poeschlr:
thanks

PR #390, opened 2019-06-26
Add TE 928492 script

evanshultz:
https://github.com/KiCad/kicad-footprints/files/3331008/ENG_CD_928492_B3.pdf
image

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1664

### DIFF
--- a/scripts/Connector/Connector_TE-Connectivity/conn_te_928492.py
+++ b/scripts/Connector/Connector_TE-Connectivity/conn_te_928492.py
@@ -1,0 +1,159 @@
+#!/usr/bin/env python3
+
+# KicadModTree is free software: you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# KicadModTree is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with kicad-footprint-generator. If not, see < http://www.gnu.org/licenses/ >.
+#
+# (C) 2016 by Thomas Pointhuber, <thomas.pointhuber@gmx.at>
+
+"""
+
+Datasheet not found available online
+See https://github.com/KiCad/kicad-footprints/files/3331008/ENG_CD_928492_B3.pdf
+
+Generates only the gray color part numbers
+Footprints are the same, but different colors have unique keying
+Thus 3D models would be unique
+
+"""
+
+import sys
+import os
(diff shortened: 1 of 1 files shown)


scripts/Connector/Connector_TE-Connectivity/conn_te_928492.py
```diff
@@ -0,0 +1,159 @@
...
+    drill = [1.2, 1.1] # pin row closest to open end is 1.1mm and back row is 1.2mm
+    pad_size = [2.4, 2.4]
+    
+    body_overhang = 7.4 # body width increase over total pin width
+    body_left = -body_overhang / 2
+    body_right = (pincount - 1) * pitch + body_overhang / 2
+    body_bottom = pitch + 8.5
+    body_top = body_bottom - 17.4
+
+    # double digit pincount parts have leading "1" otherwise "0" for gray color (other numbers mean different colors)
```
herostrat: It is hard to tell (at least for me, but the 2-928492-1 seems to be an exception.
The difference in naming seems to also stem from polarity indentation.

grafik

Tbf I do not think that's a big issue.

scripts/Connector/Connector_TE-Connectivity/conn_te_928492.py
```diff
@@ -0,0 +1,159 @@
...
+
+manufacturer = "TE-Connectivity"
+partnumber = "928492"
+datasheet = "https://github.com/KiCad/kicad-footprints/files/3331008/ENG_CD_928492_B3.pdf"
+
+pincounts = [2,4,6,7,8,9,11,12]
+
+def generate_one_footprint(pincount, configuration):
+    pitch = 5
+    drill = [1.2, 1.1] # pin row closest to open end is 1.1mm and back row is 1.2mm
```
herostrat: Just to be (absolutely) safe, you corrected the DF11 footprints wrt to tolerances.
For both it says +-0.05mm, resulting in 1.225 and 1.125.
Is this too exact?

scripts/Connector/Connector_TE-Connectivity/conn_te_928492.py
evanshultz: Hmmm, the sentence is awkward and easily confused. It captures the correct info if read as I intended. I improved the wording which hopefully addresses this comment.

scripts/Connector/Connector_TE-Connectivity/conn_te_928492.py
evanshultz: @herostrat 
We should center our pads in the tolerance range. For an equal plus and minus tolerance, the nominal value is in the middle and thus the size to use. We only deviate from the nominal value if the tolerance range is unequal. I think this is correct so I feel I'm missing your point. Can you help me understand?

herostrat: ah ok, I read that wrong.
Sorry for the false positive.

PR #392, opened 2019-07-03
Add Hirose DF11

antoniovazquezblanco:


scripts/Connector/Connector_Hirose/conn_hirose_df11_tht_top.py
```diff
@@ -0,0 +1,239 @@
...
+
+#pins_per_row per row
+pins_per_row_range = range(2,17)
+
+#Molex part number
+#n = number of circuits per row
+part_code = "DF11-{n}DP-2DSA"
+
+pitch = 2
+drill = 0.8
```
evanshultz: For 0.5mm square pins, that are ~0.7mm diagonal, IPC 2221 would say use 0.2mm oversize for the drill or 0.9mm. But if the datasheet tells us something different we go with that. In this case, the datasheet shows 0.8mm -0/+0.1mm which is a nominal 0.85mm. So please use as a the drill size.

scripts/Connector/Connector_Hirose/conn_hirose_df11_tht_top.py
```diff
@@ -0,0 +1,239 @@
...
+series_long = 'DF11 through hole'
+manufacturer = 'Hirose'
+orientation = 'V'
+number_of_rows = 2
+datasheet = 'https://www.hirose.com/product/document?clcode=&productname=&series=DF11&documenttype=Catalog&lang=en&documentid=D31688_en'
+
+#pins_per_row per row
+pins_per_row_range = range(2,17)
+
+#Molex part number
```
evanshultz: Wrong manufacturer name.

scripts/Connector/Connector_Hirose/conn_hirose_df11_tht_top.py
```diff
@@ -0,0 +1,239 @@
...
+    kicad_mod.setDescription("Molex {:s}, {:s}, {:d} Pins per row ({:s}), generated with kicad-footprint-generator".format(series_long, mpn, pins_per_row, datasheet))
+    kicad_mod.setTags(configuration['keyword_fp_string'].format(series=series,
+        orientation=orientation_str, man=manufacturer,
+        entry=configuration['entry_direction'][orientation]))
+
+    A = (pins - 1) * pitch
+    B = A + 4
+
+    # create pads
+    #createNumberedPadsTHT(kicad_mod, pincount, pitch, drill, {'x':x_dia, 'y':y_dia})
```
evanshultz: Get rid of this line

scripts/Connector/Connector_Hirose/conn_hirose_df11_tht_top.py
```diff
@@ -0,0 +1,239 @@
...
+    orientation_str = configuration['orientation_options'][orientation]
+    footprint_name = configuration['fp_name_format_string'].format(man=manufacturer,
+        series=series,
+        mpn=mpn, num_rows=number_of_rows, pins_per_row=pins, mounting_pad = "",
+        pitch=pitch, orientation=orientation_str)
+
+    footprint_name = footprint_name.replace("__",'_')
+
+    kicad_mod = Footprint(footprint_name)
+    kicad_mod.setDescription("Molex {:s}, {:s}, {:d} Pins per row ({:s}), generated with kicad-footprint-generator".format(series_long, mpn, pins_per_row, datasheet))
```
evanshultz: Wrong manufacturer name. Better to make this the variable from line 20 since I've caught this in other scripts too.

scripts/Connector/Connector_Hirose/conn_hirose_df11_tht_top.py
```diff
@@ -0,0 +1,239 @@
...
+    # create pads
+    #createNumberedPadsTHT(kicad_mod, pincount, pitch, drill, {'x':x_dia, 'y':y_dia})
+
+    optional_pad_params = {}
+    if configuration['kicad4_compatible']:
+        optional_pad_params['tht_pad1_shape'] = Pad.SHAPE_RECT
+    else:
+        optional_pad_params['tht_pad1_shape'] = Pad.SHAPE_ROUNDRECT
+
+    kicad_mod.append(PadArray(start=[0, 0], pincount=pins, increment=2, x_spacing=pitch,
```
evanshultz: You can do this in a single loop with zip(), like `for initial, initial_y in zip([1, 2], [0, pitch]):`. You can remove lines 83-86 and put this before line 88, and then change the argument `initial` to `initial` instead of `2`. Easier to maintain.

antoniovazquezblanco:
Done!

scripts/Connector/Connector_Hirose/conn_hirose_df13_tht_side.py
```diff
@@ -159,13 +159,11 @@ def generate_one_footprint(pins, configuration):
...
         kicad_mod.append(PolygoneLine(
             polygone=[
-                {'x': px,'y': py},
                 {'x': px-w,'y': py},
                 {'x': px-w,'y': py-l+0.25*w},
                 {'x': px,'y': py-l},
                 {'x': px+w,'y': py-l+0.25*w},
-                {'x': px+w,'y': py},
-                {'x': px,'y': py}],
+                {'x': px+w,'y': py}],
```
evanshultz: Is this correct? Doesn't the polygon need to have the same start and end point? That's what I've seen. Did you try this?

evanshultz:
@pointhi 
Please review the PadArray change in https://github.com/pointhi/kicad-footprint-generator/pull/392/commits/1c6fdb53d206656894fc330d147edb212919f43b.

evanshultz:
@antoniovazquezblanco 
I assume this all works. I didn't try it myself, but from the diff it looks correct or I trust you generated new footprints and confirmed all was well?

scripts/Connector/Connector_Hirose/conn_hirose_df13_tht_side.py
antoniovazquezblanco: This generates a polyline with no errors. I've updated the generated footprints in the other PR in case you want to have a look at the result. Thanks!

antoniovazquezblanco:
> I assume this all works. I didn't try it myself, but from the diff it looks correct or I trust you generated new footprints and confirmed all was well?

I have generated the footprints and updated them in the other PR

scripts/Connector/Connector_Hirose/conn_hirose_df11_tht_top.py
```diff
@@ -0,0 +1,234 @@
...
+import argparse
+import yaml
+from helpers import *
+from KicadModTree import *
+
+sys.path.append(os.path.join(sys.path[0], "..", "..", "tools"))  # load parent path of tools
+from footprint_text_fields import addTextFields
+
+series = 'DF11'
+series_long = 'DF11 through hole'
```
evanshultz: It's a little annoying that `DF11` is in two places here. The script could be smarter.

evanshultz:
@pointhi @poeschlr 
Footprints are merged.

Would you please review the changes to PadArray.py?

KicadModTree/nodes/specialized/PadArray.py
```diff
@@ -231,10 +231,9 @@ def _createPads(self, **kwargs):
             delta_pos = Vector2D(0, 0)
 
         for i, number in enumerate(pad_numbers):
-            includePad = (i + self.initialPin) not in self.exclude_pin_list
-            for exi in self.exclude_pin_list:
-                if (i + self.initialPin) == exi:
-                    includePad = False
+            includePad = True
+            if type(self.initialPin) == 'int':
+                includePad = (self.initialPin + i) not in self.exclude_pin_list
```
poeschlr: To be honest i would do something like the following:

```python
if number != "":
    include = number not in self.exclude_pin_list
else:
    include = i not in self.exclude_pin_list
```

As far as i can tell this feature is currently only used by the vh connectors and the ipc ic package scripts. Both should still work with my suggested modification (We need to check if `SOIC-4_4.55x3.7mm_P2.54mm` is still generated correctly after the change)

antoniovazquezblanco: Perfect! I will wait for https://github.com/pointhi/kicad-footprint-generator/issues/436 to be merged, then I will fix the conflicts in my PR to use your changes after everything has been tested so that there are not regressions.

Thanks!

poeschlr: 436 is an issue not a pull request

antoniovazquezblanco: I've pushed another version that I think is more complete than your proposal.

In your proposal if an str based list is passed to padarray exclusion would not be tested. The current solution checks for that.

poeschlr:
thanks

antoniovazquezblanco:
Thank you!

PR #396, opened 2019-07-18
Capacitor_THT: Update and regenerate models

antoniovazquezblanco:
I have found that the capacitors in the repo are outdated compared with the latest scripts due to the fact that https://github.com/KiCad/kicad-footprints/pull/550 was not merged.

Updated the script to split capacitors in a slightly better way.


scripts/Capacitors_THT/make_Capacitors_THT.py
```diff
@@ -181,7 +181,7 @@
         makeResistorRadial(seriesname=seriesname, rm=rm, w=w, h=d, ddrill=ddrill, R_POW=R_POW,
                                     type=type, w2=w2, x_3d=[0, 0, 0], s_3d=[1, 1, 1], has3d=1,
                                     specialfpname="", add_description=add_description, name_additions=name_additions,
-                                    specialtags=specialtags, classname="C", lib_name="${KISYS3DMOD}/Capacitor_THT_Rectangular", script3d=scr, height3d=h3d)
+                                    specialtags=specialtags, classname="C", lib_name="Capacitor_Rect_THT", script3d=scr, height3d=h3d)
```
poeschlr: I think this should be Capacitor_THT.

scripts/Capacitors_THT/make_Capacitors_THT.py
```diff
@@ -208,7 +208,7 @@
         makeResistorRadial(seriesname=seriesname, rm=rm, rm2=rm2, w=w, h=d, ddrill=ddrill, R_POW=R_POW,
                                     type=type, w2=w2, x_3d=[0, 0, 0], s_3d=[1, 1, 1], has3d=1,
                                     specialfpname="", add_description=add_description, name_additions=name_additions,
-                                    specialtags=specialtags, classname="C", lib_name="${KISYS3DMOD}/Capacitor_THT_Rectangular", script3d=scr, height3d=10)
+                                    specialtags=specialtags, classname="C", lib_name="Capacitor_Rect_THT", script3d=scr, height3d=10)
```
poeschlr: Also Capacitor_THT

scripts/Capacitors_THT/make_Capacitors_THT.py
antoniovazquezblanco: Given that there are so many rectangular capacitors, I was voting on splitting them into a separate library. Do you think that is a bad idea? http://kicad-pcb.org/libraries/klc/G1.2/

antoniovazquezblanco: https://github.com/KiCad/kicad-footprints/pull/1715

antoniovazquezblanco:
I've made the requested changes

antoniovazquezblanco:
Capacitors will no longer be split, but models need to be regenerated and updated

PR #402, opened 2019-07-26
Add FCI SFV-xxR FPC connector script

cp-aquila:
Add  script to generate footprints for the Amphenol-FCI SFV connectors series:
https://cdn.amphenol-icc.com/media/wysiwyg/files/drawing/sfv12b_hlf.pdf

This is based upon the script "conn_ffc_te_84952-84953.py" and only slightly modified
- Numerical values of pin and tab and body position + size
- Removed some Silkscreen
- Pin1 marker is smaller
- Two variants 
  - **-1ST**: Bottom contacts
  - **-2ST**: Top contacts
- I could not find the Pick'n'Place origin/offset in the datasheet, so I just put in 2mm which seems fine when looking at the connector. I would like to change that if someone spots the correct value.

Here is an example of the 10Pin part which I will be using for my board.
grafik

Currently this script build the -1ST and -2ST variant, but they are physically identical.
There is also a -3ST and -4ST variant ([Datasheet](https://www.mouser.de/datasheet/2/18/10117319-1361747.pdf)) that also seem to have the same footprint.
The script could also generate them, but that seems excessive. Even the two variants are the same, the description is different.

### DIFF
--- a/scripts/Connector/Connector_Amphenol/conn_ffc_fci-SFVxxR-xST.py
+++ b/scripts/Connector/Connector_Amphenol/conn_ffc_fci-SFVxxR-xST.py
@@ -1,0 +1,218 @@
+#!/usr/bin/env python3
+
+# KicadModTree is free software: you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# KicadModTree is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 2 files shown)


poeschlr:
Is there a footprint pull request?

cp-aquila:
> 
> 
> Is there a footprint pull request?

Sorry, I did forget that: https://github.com/pointhi/kicad-footprint-generator/pull/402#issuecomment-536167007

Is there a feature like 'alias' for footprints? The two variants are the same, just the description differs.

poeschlr:
No there is nothing similar to aliases and i doubt it will ever be introduced.
In this case the 3d model might look different depending on how detailed it is made so we would need differing footprints anyway.

And i might even go so far as to suggest including a suffix that indicates the contact direction (we might even want to add that suffix to other such connectors as part of the v6 development)

PR #403, opened 2019-07-27
Fix Wago 236 naming, drill diameters, differentiate between 1 and 2 pin

chschlue:
Fix Wago 236 naming, drill diameters, differentiate between 1 and 2 pin

Script got part numbers wrong before, ~courtyards inadequate for single pole blocks,~ drill diameter too small, created nonexistent parts.

FP PR https://github.com/KiCad/kicad-footprints/pull/1752
FP PR https://github.com/KiCad/kicad-footprints/pull/1753

### DIFF
--- a/scripts/TerminalBlock_WAGO/make_TerminalBlock_WAGO.py
+++ b/scripts/TerminalBlock_WAGO/make_TerminalBlock_WAGO.py
@@ -121,15 +121,12 @@ if __name__ == '__main__':
                                   nibbleSize=nibbleSize, nibblePos=nibblePos, fabref_offset=fabref_offset,
                                   tags_additional=[], lib_name="${KISYS3DMOD}/"+classname, classname=classname, classname_description=classname_description, webpage=webpage, script_generated_note=script_generated_note)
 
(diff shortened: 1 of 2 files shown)


poeschlr:
The drill hole diameter is given as the minimum (tolerance is given as -.../+0.1 which i would assume means the same as -0/+0.1, In the catalogue they only give positive tolerance for drill size so this also seems to suggest 0 tolerance towards smaller.)

I suggest to center the drill size in the tolerance range (1.15mm drill size would then be correct)

chschlue:
Ok, but note that these were already rather huge holes.

poeschlr:
Not really. The pins are 0.7mm square -> diagonal is already nearly 1mm. KLC would suggest adding 0.2mm to the maximum dimension so if Wago would not specify this then we would already need at least 1.2mm drill size (The KLC rule is taken from IPC suggestions)

chschlue:
Right.
I somehow had 0.7mm diameter circular pins on my mind.

poeschlr:
thanks

PR #406, opened 2019-07-31
add QFN-28-1EP 4x4mm with 2.6x2.6mm pad

lorem-ipsum:
add size definition for QFN-28 4x4mm with 2.6x2.6mm pad as defined [by Maxim](https://pdfserv.maximintegrated.com/package_dwgs/21-0139.PDF) with the T2844-1 marking.


evanshultz:
The dimensions are correct, but I would like a better way to differentiate this footprint. We need to capture `T2144-1` for this one and the proper code for the other Maxim footprint, if it has one.

@poeschlr 
I don't see any way to add some additional information to the description at https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/ipc_noLead_generator.py to capture what I mentioned above. Have I missed this capability? If not, how would you like to handle this situation? We could add a bogus parameter to the YAML file to indicate this, which could be supported by the script in the future. A YAML parameter like `unique_descriptor` which gets inserted before the datasheet field?

poeschlr:
You can add a bit more info to the datasheet field. Or make the script more powerful.

lorem-ipsum:
I just realised I messed up the reference URLs, should be fixed now. 

I can add some annotation to the script, but why for this one especially? Just to reference which one it is based on from the URL or for further information?

lorem-ipsum:
I added the code to the description, it shows up nicely in the generated symbol. Only disadvantage is that the field is no longer a valid url which might be an issue for scripts trying to parse that data.

evanshultz:
Fair point, but perhaps parsing would be best done by looking for a URL string instead of just grabbing whatever is inside the parens.

That being said, the ability to add an additional string to the description seems worthwhile to me as Rene said. But that need not block this footprint.

@poeschlr @pointhi 
Footprint is merged.

poeschlr:
thanks

PR #407, opened 2019-08-02
add Mini-Circuits DB1627

lorem-ipsum:
add size-definition for [Mini-Circuits DB1627](https://ww2.minicircuits.com/case_style/DB1627.pdf) package. The land pattern looks similar to a DFN to me, so I used this for the definition. However the package itself is not a regular DFN, so I put it in the `RF_Mini-Circuits` library and used the naming scheme of this library.

### DIFF
--- a/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/mini-circuits.yml
+++ b/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/mini-circuits.yml
@@ -1,0 +1,27 @@
+Mini-Circuits_DB1627:
+  device_type: DFN
+  library: RF_Mini-Circuits
+  manufacturer: Mini-Circuits
+  custom_name_format: "{man}_{mpn}"
+  part_number: DB1627
+  size_source: https://ww2.minicircuits.com/case_style/DB1627.pdf
+  ipc_class: dfn
+  body_size_x:
+    nominal: 3.81
+    tolerance: 0.127
+  body_size_y:
+    nominal: 4.06
+    tolerance: 0.127
+  body_height:
+    maximum: 4.06
+
+  lead_width:
+    nominal: 0.71
+    tolerance: 0.127
+  lead_len:
+    nominal: 1.02
+    tolerance: 0.127
+
+  pitch: 1.27
+  num_pins_x: 3
+  num_pins_y: 0


scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/mini-circuits.yml
```diff
@@ -0,0 +1,27 @@
...
+  device_type: DFN
+  library: RF_Mini-Circuits
+  manufacturer: Mini-Circuits
+  custom_name_format: "{man}_{mpn}"
+  part_number: DB1627
+  size_source: https://ww2.minicircuits.com/case_style/DB1627.pdf
+  ipc_class: dfn
+  body_size_x:
+    nominal: 3.81
+    tolerance: 0.127
```
evanshultz: Isn't the +/-0.01 inch in the datasheet translated into 0.254 mm here? I think your tolerance is half the actual value.

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/mini-circuits.yml
```diff
@@ -0,0 +1,27 @@
...
+  body_size_y:
+    nominal: 4.06
```
evanshultz: This is the overall length. I don't see the body size given.

evanshultz:
1. Please name file extension `yaml` for consistency.
1. Please enclose string values in single quotes. Again for consistency. See other files.

I stopped here because this generator is no no-lead packages. Your package has leads. I think you will need to make it manually as I don't believe we have a generator for this package type.

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/mini-circuits.yml
lorem-ipsum: The tolerance should be correct. The tolerances are for the values in inches. In inches it has 3 places, so it is +-0.005 inch.

lorem-ipsum:
Based on your observation in the matching footprint PR (https://github.com/KiCad/kicad-footprints/pull/1762) this footprint is not a no lead, so the script is completely wrong and useless. Closing and hand building.
Anyways thanks for the review, will try to integrate it in future footprints.

PR #408, opened 2019-08-07
Fix bug in tantal definition of SMD_chip_package

lorem-ipsum:
There was a redundant definition. There is also a (much larger) commit which moves the footprint generation into a independent function, guarded by a try except block. This makes it a lot easier to trace errors.


poeschlr:
thanks

PR #411, opened 2019-08-12
fix bug in addLineWithKeepout for horizontal or vertical lines

ppelleti:
When `addLineWithKeepout` was used to draw horizontal or vertical lines, it would omit the last segment.  This fixes that bug.


poeschlr:
thanks

PR #412, opened 2019-08-18
Add WQFN-24-1EP_4x4mm_P0.5mm_EP2.45x2.45mm

jonas-schievink:
Used by TS3A27518E.

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1785

Symbol PR: https://github.com/KiCad/kicad-symbols/pull/2092


scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/qfn/wqfn.yaml
```diff
@@ -224,3 +224,59 @@ WQFN-24-1EP_4x4mm_P0.5mm_EP2.6x2.6mm:
...
+    maximum: 4.1
+  body_size_y:
+    minimum: 3.9
+    maximum: 4.1
+  overall_height:
+    maximum: 0.8
+
+  lead_width:
+    minimum: 0.2
+    maximum: 0.3
```
poeschlr: Datasheet gives it as 0.24 .. 0.34

reportingsjr:
I checked the dimensions on this footprint and they all look correct. The only thing I noticed is that the suggested paste coverage of 75% is for 5 thermal vias that avoid the paste, whereas this script has 9 thermal vias not avoiding the paste with no increase in paste coverage.

I'm not familiar enough with manufacturing including thermal vias in pad to know if that is too little paste for this change.

poeschlr:
78% paste coverage is insanely high. Most application notes i read (including TI) suggest something between 50% to 70% (The high end is normally only used for high power devices. Not sure why a multiplexer would need this.) See: http://www.ti.com/lit/an/slua271b/slua271b.pdf#page=11

The danger with too much paste is that your part lifts away and your normal pads are not correctly connected.

poeschlr:
thanks

PR #413, opened 2019-08-19
add 

Sorontik:
Hi,

Just like a SMD-Pin-Header can start with the first pin left or right, angled/horizontal pin-headers/sockets can also be 'bent' to the left or to the right:
Imagine a single column of e.g. 4 straight (vertical) pins with the top one being pin 1.
I can now "bend" those pins either to the left or to the right to get the respective horizontal version.
These two variants are axisymmetrical but can't be tranformed into one another by only rotating. (Unlike the straight/vertical ones, which can be rotated to "invert" the pin numbering)

Since these connectors are very basic and commonly used in many Projects, I think both variants should be included in the library, instead of forcing everybody to add them manually on their own.

Your script only generated the right-bent variant, where the first pin is on the right (when looking at the tips of the pins), so i added a new parameter `pin_num_reverse` to `makePinHeadAngled` and `makeSocketStripAngled`. The parameter was added at the end and defaults to False, so existing usages of the functions still behave in the same way.

If pin_num_reverse is True, the pin-number `p` starts with the max-value and gets decreased in the Pad-creating loops.
So the pins-positions are still created in the same order but assigned a different number:
The pin at the first position gets the last pin-number and vice versa.
Additionally, a 'reversed' is appended to footprint name, description and tags.

I also edited the Scripts in `/scripts/pin-headers_socket-strips/` to use the new parameter to create both versions of angled/horizontal pin headers.

### DIFF
--- a/scripts/pin-headers_socket-strips/make_pin_headers.py
+++ b/scripts/pin-headers_socket-strips/make_pin_headers.py
@@ -49,6 +49,9 @@ if __name__ == '__main__':
                                 [0, 0, 0], [1, 1, 1], [0, 0, 0])
             makePinHeadAngled(rows, cols, rm, rm, angled_pack_width, angled_pack_offset, angled_pin_length, angled_pin_width, ddrill, pad,
                               [], "${KISYS3DMOD}/Pin_Headers", "Pin_Header", "pin header", [0, 0, 0], [1, 1, 1], [0, 0, 0])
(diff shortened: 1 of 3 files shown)


evanshultz:
Another way of looking at this is selecting a different pin 1 for a physical part in the same orientation. Can you provide a datasheet showing this difference? From the subset of datasheets I've looked at the pin 1 position was always shown in the same location.

Note that I've made changes to some of the same files at https://github.com/pointhi/kicad-footprint-generator/pull/387. And I looked at similar, but not the same, header datasheet there.

PR #414, opened 2019-08-19
Add VQFN-32-1EP_5x5mm_P0.5mm_EP3.5x3.5mm

jonas-schievink:
Used by FTDI's FT4222H.

Based on the data in https://www.ftdichip.com/Support/Documents/TechnicalNotes/TN_166%20FTDI%20Example%20IC%20PCB%20Footprints.pdf#page=22

I had to fight the script a little, because it would automatically make the footprint larger than what I entered, but the values are based off the TN I linked, so they should be kept as-is. The leads are still longer than what the TN specifies (0.8mm instead of 0.75mm).

Symbol PR: https://github.com/KiCad/kicad-symbols/pull/2091

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1786


poeschlr:
The linked document does not contain the right data. This generator does not take a suggested footprint as input but the dimensions (including tolerances) of the packages itself. Meaning it takes the input as found in https://www.ftdichip.com/Support/Documents/DataSheets/ICs/DS_FT4222H.pdf#page=40

jonas-schievink:
@poeschlr That makes sense. Fixed the dimensions (only the lead length was off).

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/qfn/vqfn.yaml
```diff
@@ -560,3 +560,62 @@ VQFN-32-1EP_5x5mm_P0.5mm_EP3.1x3.1mm:
...
+  thermal_vias:
+    count: [3, 3]
+    drill: 0.33
+    # min_annular_ring: 0.15
+    paste_via_clearance: 0.1
+    # EP_num_paste_pads: [2, 2]
+    # paste_between_vias: 1
+    # paste_rings_outside: 1
+    EP_paste_coverage: 0.7
+    grid: [1.2, 1.2]
```
poeschlr: Is there a reason to manually enter the grid?

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/qfn/vqfn.yaml
```diff
@@ -560,3 +560,62 @@ VQFN-32-1EP_5x5mm_P0.5mm_EP3.1x3.1mm:
...
+  EP_size_y:
+    minimum: 3.45
+    nominal: 3.5
+    maximum: 3.55
+  # EP_paste_coverage: 0.65
+  EP_num_paste_pads: [2, 2]
+
+  thermal_vias:
+    count: [3, 3]
+    drill: 0.33
```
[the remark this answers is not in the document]
jonas-schievink: No, this is just copied from the footprint above. The datasheet doesn't mention thermal vias at all, so theoretically I could just remove them completely.

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/qfn/vqfn.yaml
jonas-schievink: No, this is also due to copy-pasting the footprint above.

poeschlr: In this case remove the grid definition.

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/qfn/vqfn.yaml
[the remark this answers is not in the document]
poeschlr: In this case i would simply use 0.3mm (The same package can be used by a part that requires them so it makes sense to have this footprint)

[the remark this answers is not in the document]
jonas-schievink: Okay, done that.

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/qfn/vqfn.yaml
jonas-schievink: Also removed this. The FP now looks like this:

screenshot-2019-09-01-22:18:19

poeschlr:
thanks

PR #415, opened 2019-08-19
Drop Python 3.3 on Travis

jonas-schievink:
It has reached EOL almost 2 years ago, and CI fails because apparently the archive is now gone.

Let's see if this fixes Travis. Closing this PR and waiting for Travis to fix this is also an option, but I see little value in testing on 3.3.


chschlue:
I second that.
I would also go a step further and drop 2.7. It deserves to rest in peace.

pointhi:
Sorry, missed it, did it in https://github.com/pointhi/kicad-footprint-generator/commit/56245b802d7d9370c3c3818899c95d4145684572

PR #416, opened 2019-08-23
Add HTSSOP-28-1EP_4.4x9.7mm_P0.65mm_EP2.85x5.4mm

chschlue:
FP PR: https://github.com/KiCad/kicad-footprints/pull/1788

### DIFF
--- a/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/htssop.yaml
+++ b/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/htssop.yaml
@@ -213,21 +213,6 @@ HTSSOP-28-1EP_4.4x9.7mm_P0.65mm_EP2.85x5.4mm:
         num_pins_y: 14
         EP_size_x: 2.6 .. 3.1
         EP_size_y: 5.2 .. 5.6
-
-HTSSOP-28-1EP_4.4x9.7mm_P0.65mm_EP2.85x5.4mm_ThermalVias:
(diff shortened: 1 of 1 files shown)


scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/htssop.yaml
```diff
@@ -199,6 +199,41 @@ HTSSOP-24-1EP_4.4x7.8mm_P0.65mm_EP3.4x7.8mm_Mask2.4x4.68mm_ThermalVias:
     # bottom_pad_size:
     paste_avoid_via: False
 
+HTSSOP-28-1EP_4.4x9.7mm_P0.65mm_EP2.85x5.4mm:
```
poeschlr: You do not need to duplicate the non thermal vias version. (if thermal vias section is given then non thermal via version is generated as well)

poeschlr:
Dimensions look ok. But you have them duplicated which is not necessary.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/htssop.yaml
chschlue: Done

poeschlr:
I also just noticed that all other parts use tab-width = 2 spaces while you used way more. Could you reduce the number of spaces per tab used to align with the rest of the file?

chschlue:
I'd rather use the "one-tab-per-tab" convention but that's apparently frowned upon by Python folks.

Here you go.

poeschlr:
thanks

PR #422, opened 2019-09-08
Ep roundradius handling

poeschlr:
Using large a rather large round radius for exposed pads was a bad idea. See https://github.com/KiCad/kicad-footprints/issues/1798

TlDr: use the same radius as the normal pads instead. Also added a parameter to overwrite this.
This required adding a more powerful round radius handler to avoid code duplication.


KicadModTree/nodes/base/Pad.py
```diff
@@ -24,6 +24,120 @@
...
+      (This is useful for IPC-7351C compliance as it suggests 25% ratio with limit 0.25mm)
+      Ignored for every shape except round rect.
+    * *round_radius_exact* (``float``) --
+      Set an exact round radius for a pad.
+      Ignored for every shape except round rect
+    """
+    def __init__(self, **kwargs):
+        radius_ratio = kwargs.get('radius_ratio', 0)
+        if type(radius_ratio) not in [int, float]:
+            raise TypeError('radius ratio needs to be of type int or float')
```
cpresser: use '__checkAbsoluteValue' here?

cpresser:
I only managed to review ~50% of the code. Its a rather big patch and not that simple. My goal is to get this done this weekend.

KicadModTree/nodes/base/Pad.py
```diff
@@ -210,22 +319,12 @@ def _initLayers(self, **kwargs):
...
-        if kwargs.get('maximum_radius') is not None:
-            maximum_radius = kwargs.get('maximum_radius')
-            if type(maximum_radius) not in [int, float]:
-                raise TypeError('maximum radius needs to be of type int or float')
+            self.round_radius_handler = RoundRadiusHandler(**kwargs)
 
-            shortest_sidelength = min(self.size)
-            if self.radius_ratio*shortest_sidelength > maximum_radius:
-                self.radius_ratio = maximum_radius/shortest_sidelength
+        self.radius_ratio = self.round_radius_handler.getRadiusRatio(min(self.size))
```
cpresser: why call min() with only one argument?

KicadModTree/nodes/specialized/ExposedPad.py
```diff
@@ -65,6 +65,7 @@ class ExposedPad(Node):
         * *paste_coverage* (``float``) --
           how much of the mask free area is covered with paste. (default: 0.65)
 
+
```
cpresser: extra newline?

poeschlr:
The change of this one parameter is part of this set because i could not generate the footprints without that fix. And this fix is connected to a single footprint pull request.

KicadModTree/nodes/base/Pad.py
```diff
@@ -24,6 +24,108 @@
...
+      Ignored for every shape except round rect.
+    * *maximum_radius* (``float``) --
+      The maximum radius for the rounded rectangle.
+      If the radius produced by the radius_ratio parameter for the pad would
+      exceed the maximum radius, the ratio is reduced to limit the radius.
+      (This is useful for IPC-7351C compliance as it suggests 25% ratio with limit 0.25mm)
+      Ignored for every shape except round rect.
+    * *round_radius_exact* (``float``) --
+      Set an exact round radius for a pad.
+      Ignored for every shape except round rect
```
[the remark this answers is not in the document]
poeschlr: removed

KicadModTree/nodes/specialized/ExposedPad.py
poeschlr: removed

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/ipc_gullwing_generator.py
```diff
@@ -34,8 +34,14 @@ def __init__(self, configuration):
                 self.ipc_defintions = yaml.safe_load(ipc_stream)
 
                 self.configuration['min_ep_to_pad_clearance'] = 0.2
+
+                #ToDo: find a settings file that can contain these.
```
evanshultz: Why not pile these settings into https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/Packages/ipc_definitions.yaml? It also has a `ipc_generic_rules` section. Also in `ipc_noLead_generator.py`.

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/ipc_noLead_generator.py
```diff
@@ -41,6 +41,11 @@ def __init__(self, configuration):
...
 
                 self.configuration['min_ep_to_pad_clearance'] = 0.2
+
+                #ToDo: find a settings file that can contain these.
+                self.configuration['paste_radius_ratio'] = 0.25
+                self.configuration['paste_maximum_radius'] = 0.25
+
                 if 'ipc_generic_rules' in self.ipc_defintions:
                     self.configuration['min_ep_to_pad_clearance'] = self.ipc_defintions['ipc_generic_rules'].get('min_ep_to_pad_clearance', 0.2)
             except yaml.YAMLError as exc:
```
evanshultz: This line has a space before it in `ipc_gullwing_generator.py `.

KicadModTree/nodes/specialized/ExposedPad.py
```diff
@@ -105,11 +105,22 @@ class ExposedPad(Node):
...
+          The radius ratio of the main pads.
         * *maximum_radius* (``float``) --
-          Only used if a radius ratio is given.
-          Limits the radius.
+          The maximum radius for the main pads.
+          If the radius produced by the radius_ratio parameter for the pad would
+          exceed the maximum radius, the ratio is reduced to limit the radius.
+        * *round_radius_exact* (``float``) --
+          Set an exact round radius for the main pads.
+
```
evanshultz: Is this empty line intentional?

KicadModTree/nodes/base/Pad.py
```diff
@@ -50,13 +149,20 @@ class Pad(Node):
...
+          exceed the maximum radius, the ratio is reduced to limit the radius.
           (This is useful for IPC-7351C compliance as it suggests 25% ratio with limit 0.25mm)
-          Ignored for every other shape.
+          Ignored for every shape except round rect.
+        * *round_radius_exact* (``float``) --
+          Set an exact round radius for a pad.
+          Ignored for every shape except round rect
+        * *round_radius_handler* (``RoundRadiusHandler``) --
+          An instance of the RoundRadiusHandler class
+          If this is given then all other round radius specifiers are Ignored
```
evanshultz: Is the capital "I" intentional? This is repeated in other files too.

KicadModTree/nodes/base/Pad.py
```diff
@@ -24,6 +24,105 @@
...
+
+        :param shortest_sidelength: shortest sidelength of a pad
+        :return: the resulting round radius to be used for the pad
+        """
+        return self.getRadiusRatio(shortest_sidelength)*shortest_sidelength
+
+    def roundingRequested(self):
+        r"""Check if the pad has a rounded corner
+
+        :return: True if rounding corners is required
```
evanshultz: "is" should be "are".

KicadModTree/nodes/base/Pad.py
```diff
@@ -210,22 +316,12 @@ def _initLayers(self, **kwargs):
...
-        if kwargs.get('maximum_radius') is not None:
-            maximum_radius = kwargs.get('maximum_radius')
-            if type(maximum_radius) not in [int, float]:
-                raise TypeError('maximum radius needs to be of type int or float')
+            self.round_radius_handler = RoundRadiusHandler(**kwargs)
 
-            shortest_sidelength = min(self.size)
-            if self.radius_ratio*shortest_sidelength > maximum_radius:
-                self.radius_ratio = maximum_radius/shortest_sidelength
+        self.radius_ratio = self.round_radius_handler.getRadiusRatio(min(self.size))
```
evanshultz: Why is `min()` called with just a single argument? This happens several times in the changed code. I thought I recalled seeing a comment like this earlier, but I don't see it nor so it must have been for a different PR.

evanshultz:
The resulting footprints seem OK to me and at a high-level this appears fine. I admit to not following everything you've written but am OK merging this if nobody else objects. My comments above are reduced mostly to semantics without any logical effect.

KicadModTree/nodes/base/Pad.py
cp-aquila: I spotted the same thing, not sure if I submited the same question and then delted it.
But then I did some checking and learned that calling min() also works with an interable object. It returns the smallest element.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/ipc_gullwing_generator.py
poeschlr: Make an issue about that as this change should really not include anything other than the changes to the ep rounding

KicadModTree/nodes/base/Pad.py
poeschlr: self.size is an itterable. So min is called to get the minimum of the itterable

evanshultz: Ahh. I didn't realize self.size was an iterable. Got it now.

evanshultz:
@poeschlr 
I have four comments about semantics above (the middle four). If you can make those fixes or help me understand the intent I'm OK to merge this.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/ipc_gullwing_generator.py
evanshultz: https://github.com/pointhi/kicad-footprint-generator/issues/435

KicadModTree/nodes/specialized/ExposedPad.py
poeschlr: yes
separates the description of the copper pad handling parameters from the paste handling ones

KicadModTree/nodes/base/Pad.py
poeschlr: fixed

KicadModTree/nodes/base/Pad.py
poeschlr: fixed

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/ipc_noLead_generator.py
poeschlr: fixed

KicadModTree/nodes/specialized/ExposedPad.py
evanshultz: There is no empty line in `Pad.py` before the paste stuff.

KicadModTree/nodes/base/Pad.py
evanshultz: Only fixed in this file. As I mentioned, it's repeated elsewhere. I see three other capital "I".

KicadModTree/nodes/base/Pad.py
evanshultz: Change "is" to "are", not "if" to "are". The wrong word was changed.

KicadModTree/nodes/specialized/ExposedPad.py
poeschlr: There are now

KicadModTree/nodes/base/Pad.py
```diff
@@ -24,6 +24,105 @@
...
+
+        :param shortest_sidelength: shortest sidelength of a pad
+        :return: the resulting round radius to be used for the pad
+        """
+        return self.getRadiusRatio(shortest_sidelength)*shortest_sidelength
+
+    def roundingRequested(self):
+        r"""Check if the pad has a rounded corner
+
+        :return: True are rounding corners is required
```
evanshultz: Let's try again: should be `True if rounded corners are required`.

KicadModTree/nodes/specialized/ChamferedPad.py
```diff
@@ -212,11 +212,17 @@ class ChamferedPad(Node):
...
-          Limits the radius.
+          The maximum radius for the rounded rectangle.
+          If the radius produced by the radius_ratio parameter for the pad would
+          exceed the maximum radius, the ratio is reduced to limit the radius.
+          (This is useful for IPC-7351C compliance as it suggests 25% ratio with limit 0.25mm)
+        * *round_radius_exact* (``float``) --
+          Set an exact round radius for a pad.
+        * *round_radius_handler* (``RoundRadiusHandler``) --
+          An instance of the RoundRadiusHandler class
+          If this is given then all other round radius specifiers are Ignored
```
evanshultz: 'Ignored' should have a lowercase 'i'.

KicadModTree/nodes/specialized/ChamferedPadGrid.py
```diff
@@ -180,11 +180,17 @@ class ChamferedPadGrid(Node):
...
-          Limits the radius.
+          The maximum radius for the rounded rectangle.
+          If the radius produced by the radius_ratio parameter for the pad would
+          exceed the maximum radius, the ratio is reduced to limit the radius.
+          (This is useful for IPC-7351C compliance as it suggests 25% ratio with limit 0.25mm)
+        * *round_radius_exact* (``float``) --
+          Set an exact round radius for a pad.
+        * *round_radius_handler* (``RoundRadiusHandler``) --
+          An instance of the RoundRadiusHandler class
+          If this is given then all other round radius specifiers are Ignored
```
evanshultz: 'Ignored' should have a lowercase 'i'.

evanshultz:
I've posted three comments above which are not addressed.

poeschlr:
I merged master into this to have all footprints fixed that were exported since this started.
I also moved the VSON footprint to a separate definition file and ensured it is put into the correct lib.
And i fixed the spelling mistakes.

The footprint pull requests are now split up.

evanshultz:
Rene, I assume you want to get this into 5.1.5. I've finally gotten time to work on KiCad again and I'm prepared to help make this happen. Sorry for being mostly absent lately.

Confirmed all typos are fixed. The code changes look OK to me. Thanks!

I noticed page 12 of https://ocipcdc.org/archive/What_is_New_in_IPC-7351C_03_11_2015.pdf shows that thermal pads are exempt from rounded corners. Tom confirmed to me that it's fine if the manufacturer says to use rounded corners, so I don't have any information that there's a technical reason that rounded corners are bad (but as you've seen the radius should at least be reduced as you have done here). So should we use square thermal pads for a non-manufacturer footprint?

poeschlr:
I made the generator use the smaller rounded corners because this was how we decided it in the issue. Unless you have new information i would say we stick to that.

evanshultz:
I'm not sure if it's new information, but before I see the discussion centered on using the radius of the outer pads. And I'm sharing a doc and conversation with someone on the IPC 7351 board that is saying thermal pads on generic IPC 7351 footprints should not have rounded corners. I thought it was new information but maybe you already considered this?

poeschlr:
Ok will make the default 0, allow the footprint definition to select "pad" or exact number. I will also allow this to be overwritten in the global config file.

And i will add a parameter for defining the round ratio (should the exact number not be given) in both the footprint definition and global config.

If everything goes to plan then this should be done within a few hours

poeschlr:
It is now done. The footprint pull requests are now updated with rectangle pads for the EP

evanshultz:
Thanks! I hope this is the right call and for the best in this library. It should ensure the metal tab never sticks out of the mask even if the corners are farther away than the sides.

I plan to review this and footprints on Monday. It seems we will have no problem to get this in 5.1.5.

evanshultz:
I have reviewed all the PRs except for the TI one because it's affected by a question below, and merged the ones that were ready. Some comments from the individual footprint PRs:
1. Should we allow generic footprints for TI packages if TI gives guidance on a special footprint? This might mean all TI footprints need to be split off into their own YAML files or maybe a special keyword?
1. Should footprints with beveled corner pads (I think only QFN-like footprints) get a suffix?

poeschlr:
To answer your questions:

> Should we allow generic footprints for TI packages if TI gives guidance on a special footprint? This might mean all TI footprints need to be split off into their own YAML files or maybe a special keyword?

Nearly all manufacturers provide suggested footprints within their documentation. TI however has a very high quality here. 

To be honest i would go the opposite direction and reduce the number of manufacturer specific footprints. Right now we have a lot of TI specific footprints (as an example) which really are just duplicates of generic packages. For TI i would say if there is a note that IPC is an alternative then i would go with fully generic.

My thinking is that the official lib focuses on providing generic footprints and the kicad project adds a IPC footprint generator that allows making footprints to ones personal standard with the exact dimensions of the current package at hand. (I am already working on porting the footprint generator over to footprint wizards but hit a few roadblocks regarding the options available for the graphical interface. Example https://bugs.launchpad.net/kicad/+bug/1846404)

Specialized footprints are mainly necessary for high frequency parts (impedance control) or MEMS (mechanical stress management). Footprints for these should go into their respective libs and not be made with generators.

---

> Should footprints with beveled corner pads (I think only QFN-like footprints) get a suffix?

Might be a good idea. I suggest making that an issue over on the footprint repo.
I however see no need to rush. Footprints that use this feature need to use it because the combination of pitch, lead count and body size does not really allow for an alternative.

0xdec:
Just a thought since this seems to already be decided, but:
In my experience, surface tension does a reasonably good job centering the part to the pad, but is rarely perfect. Since the exposed pad and its corresponding copper pad are usually 1:1 unlike most other pins (which account for solder fillet), any small deviation from perfect alignment would mean some portion of the part's EP lying over soldermask. We had this issue when trying to use 1:1 pads to rely on surface tension to align a part; it didn't work so well due to tolerances on the pads on the part. All this is to say, it seems to me that rounding the EP is actually the way to go since we can never rely on it being perfectly centered anyhow. Just my 2¢.

evanshultz:
All footprint PRs from this change are now merged. So if you fix the branch conflict we can merge this and then move on with merging other footprint PRs using this script. Thanks!

poeschlr:
@evanshultz i merged this and tested the result with diff (ran the scripts before and after merge against all their yaml files and checked if before and after are the same other than the timestamp)

evanshultz:
OK then. Thanks!

@pointhi 
Would you like to take a look at this? If you're OK just going ahead and merging, please do it or tell me to. Thank you!

pointhi:
@evanshultz I see a lot of CI errors. Those need to be fixed before thinking of merging

poeschlr:
@pointhi yes it was a difference regarding the default round radius handling of chamfered pads, chamfered pad grids and exposed pads. All of them used 0 as the default in the past but i had unified this over all nodes with the introduction of the centralized round radius handler class.
I now introduced a default value parameter to the init function of this class to allow customizing the default value to 0 for these 3 classes.

chschlue:
Is this seems to be (almost) finished), I collected all open PRs affected by it above.

evanshultz:
@pointhi 
I had in mind getting your approval for the concept introduced in this PR, and then merging once ready. But it's not obvious. Do you have any objections to the work done here?

evanshultz:
I'm good with everything here. Thanks @poeschlr !

@pointhi 
Are you OK with the concept being introduced here and having it merged? This is blocking quite a number of PRs so it would be great to get this merged.

pointhi:
@evanshultz I did a rough look, and the concept looks good. If you or @poeschlr are ok you can merge anytime.

From my side: It would be nice to add a few tests of this feature, to validate nothing breaks.

evanshultz:
@pointhi 
Thanks!

@poeschlr 
Do you have some unit tests you can implement? I suspect you'd rather merge this now to open up the dam on other PRs, and come back to the topic of testing this feature in a later PR.

poeschlr:
The unit tests i had already showed weaknesses which i solved in my last commit. So yes there are unit tests already there. Could there be more? Of course. I invite anyone to write some, i simply do not have the time fir this.

evanshultz:
Alright then. Thank you!

chschlue:
Looks like these slipped through:
https://github.com/KiCad/kicad-footprints/pull/1542
https://github.com/KiCad/kicad-footprints/pull/1922
https://github.com/KiCad/kicad-footprints/pull/1931
https://github.com/KiCad/kicad-footprints/pull/1932
https://github.com/KiCad/kicad-footprints/pull/2026

PR #437, opened 2019-10-01
Support 43045-xx2x MPNs

evanshultz:
- These have solderable metal clips to support the header instead of plastic pegs (https://www.molex.com/pdm_docs/sd/430450221_sd.pdf)
- Corrected one peg location in the old footprints (just changed by 0.02mm)
- Also more use of variables

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1885


scripts/Connector/Connector_Molex/conn_molex_micro-fit-3.0_tht_side_dual_row.py
```diff
@@ -232,4 +273,5 @@ def generate_one_footprint(pins, configuration):
     configuration['kicad4_compatible'] = args.kicad4_compatible
 
     for pincount in pins_per_row_range:
-        generate_one_footprint(pincount, configuration)
+        for type in types:
```
cpresser: `type` is a python build-in function. a rather bad choice for a variable name.
perhaps use variant?

evanshultz: Good point. Will fix.

PR #438, opened 2019-10-02
Generate separate C, L, and R chip footprints

### LINKED ISSUE #420: Chip body sizes are too generic?
@poeschlr 
A few things with https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/SMD_chip_package_rlc-etc/size_definitions/size_default_chip_devices.yaml.

1. The tortai-tech.com link seems broken. I assume that was from IPC-SM-782? Perhaps a comment to indicate the document source, even if we can't find it available online?
2. SMD_0603 matches IPC-SM-782 sizes so I assume that was the reference. But for SMD_0805, the only significant deviation from the IPC doc is in terminator spacing. I suppose we could use the IPC spec for the other two dimensions and then just modify the terminator spacing to suit our needs (we both will recall the huge variation we found and the limits we placed on this parameter based on gathering data from parts in the market).
2. Here is the major issue that I found which caused me to make this issue: I'm concerned the generic SMD_0603, SMD_0805, etc. size definitions aren't specific enough. IPC-SM-782 shows different dimensions for chip resistors (page 72) and chip capacitors (page 76) which are NOT the same for the same body size. So instead of SMD_0603 should we have C_0603 and R_0603, and the same for other body sizes and part types?

If I'm right about #3, it means that our footprints are using correct IPC calculations for fillet goals, but the body sizes that are an input to the IPC calculations could (should?) be more specific for different part types.

evanshultz:
Fixes https://github.com/pointhi/kicad-footprint-generator/issues/420

Because IPC tells us there are differences in the body sizes, not all chip parts should result in the same footprints. This generates unique footprints for each part type.
- Add unique body size info for C, L, and R part types in accordance with IPC-SM-782 (with a slight deviation for 0805 capacitors)
- File renames to remove 'chip' if file contained more than chip info
- Aligned size definition file names
- Print part type and definition during execution
- Corrected descriptions and keywords in SMD_chip_devices.yaml
- Fixed a few typos in comments

See a comparison of resulting footprints at https://docs.google.com/spreadsheets/d/1OAOwxWGSfmM8BzPS2x2mySmMnPzjYmWQ7ZMfdHtCbws/edit?usp=sharing.

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/2280

Related: https://github.com/KiCad/kicad-footprints/issues/711


evanshultz:
@poeschlr 
Would you be willing and able to take a look at this?

evanshultz:
@poeschlr 
I made a separate commit I found for an error in one YAML file, but I should have made all the files changes in one commit and then done a separate commit for all the file renames (or put that in a separate branch). That probably would have been easier to review. My apologies for not thinking about how to do this better earlier. Is there anything I can do now to help?

And one question: because there's now a YAML file just for resistors, should the wide-body YAML file be merged into the regular resistor file? I kept them separate for continuity reasons but I really don't see why they can't be together in one file.

evanshultz:
Should I merge in master the same way I did at https://github.com/pointhi/kicad-footprint-generator/pull/439? Or is that not the right way?

evanshultz:
@poeschlr 
Could you take a look at this? Do you see problems with using unique body sizes for chip footprints as defined by IPC?

Also, merge in master as I did above to fix the conflict?

poeschlr:
I do not see a problem with using unique body sizes for different devices. However, be aware that we did not use the outdated IPC document everywhere. And we did avoid it for 0805 completely as we did our own research.

The reality is that such devices are much less standardized than we would like. Which means the footprints of the official lib can only ever be a compromise. Unless we really want to add one footprint per component series (not really feasible).

I hope that 5.1.6 finally fixes a deal breaker bug for the footprint wizard allowing me to continue working on a IPC footprint wizard for these sort of devices that will give users the option to generate fully compliant footprints for the device they really use.

poeschlr:
Regarding fixing the merge conflicts: You can either merge master into this one or rebase this onto the current master. The latter might be better in the sense of a cleaner git history but it also requires more git skills.

evanshultz:
Yes. As this was a while ago I may have forgotten some details, but from memory and looking at the Google Sheets I linked the IPC-SM-782 0805 cap size and the body size you came up with from reviewing datasheets is quite similar. So I think the initial issues were related from us not realizing that C and L body sizes were given in the IPC doc, and instead using the R body size for all part types.

I will try a rebase, then. First time for everything...

evanshultz:
@poeschlr 
Now I see what you meant. I had already merged in master above to rebase wouldn't work.

I merged in master again with a little manual work. Branch conflicts resolved but the size definitions added at https://github.com/pointhi/kicad-footprint-generator/pull/440 have a `SMD_` prefix instead of `C_`. When I've encountered this before I didn't find a way to resolve it in the same PR without GH seeing more branch conflicts so either made a second PR or a new replacement PR. Is there a way to fix that in this PR?

Also, I noticed a typo of 'tantalum' (there is a trailing 'n') so can that be done here too?

poeschlr:
One option to fix up the branch is to locally rename it, make a new branch with the current online name and then cherry-pick the commits from your original (now renamed branch) to your new branch correcting merge conflicts as necessary.

This might help a bit: https://learngitbranching.js.org/

evanshultz:
@poeschlr 
The branch is now fixed without conflicts. In addition, I fixed a typo and also an error when running the script. Not sure how that remained.

New footprints are submitted at https://github.com/KiCad/kicad-footprints/pull/2280.

evanshultz:
@poeschlr 
I created a footprint PR and requested your review, but most of the discussion was here so this is where I'm commenting. Will you be able to take a look at this? I have documented more info in the footprint PR.

evanshultz:
@poeschlr 
Ping!

scripts/SMD_chip_package_rlc-etc/size_definitions/size_resistor_chip.yaml
```diff
@@ -0,0 +1,112 @@
...
+        maximum: 5.15
+    body_width:
+        minimum: 2.35
+        maximum: 2.65
+    terminal_length:
+        minimum: 0.35
+        maximum: 0.85
+    size_info: '(Body size source: IPC-SM-782 page 72, https://www.pcb-3d.com/wordpress/wp-content/uploads/ipc-sm-782a_amendment_1_and_2.pdf)'
+
+SMD_1812:
```
diegoherranz, line 71: Should this be R_1812 instead of SMD_1812?

evanshultz: Yes. Initially that broke my branch but I think I know better how to fix it now so I went for it again. However, the branch can still be merged so I guess we're good.

evanshultz:
@diegoherranz 
If you're feeling up to it and want to review this that's fine with me. It seems @poeschlr is busy and @pointhi generally will merge whatever is approved into /scripts. But of course, take your time to review and make sure everything makes sense and you don't see anything I messed up. I don't want to break anything when I'm trying to fix things!

evanshultz:
@poeschlr 
Ping again!

evanshultz:
@cpresser 
You wanna take a look at this one?

evanshultz:
@cpresser 
Are you able to work on this now? I would quite like to have this in 5.1.7. I know we have a lot of footprints to sort out now as well.

Since Rene wrote the original generator script he knew the history. Let me try to capture what I recall here. There are also a few links above with more background info.

When Rene wrote the script we found that the cap footprints, specifically 0805, didn't see correct. Rene then surveyed info from cap datasheets and faked some body size for caps based on real datasheets. IPC-SM-782 has body size info and it turns out that the resistor dimensions were being used for all part types. When inputting the correct cap dimensions, the resulting footprints are almost exactly the same as the ones from Rene's research. While caps were noted by Wayne and precipitated making some manual adjustments earlier, other part types also have specific body sizes in that IPC doc. I captured all those body sizes here and then made updates so the correct body sizes are being used for each specific part type.

You may be able to search online and find IPC-SM-782, but if not I guess you just have to trust me on the dimensions??

In essence, it was just an oversight when developing the script to miss that IPC-SM-782 shows resistor body sizes first but there are other body sizes in later pages.

The reason why I kept pushing on this and marked it as a bug is that:
1. We claim to have IPC 7351 footprints, but the body sizes are wrong for some parts. So bad input gives bad output. Some of our chip footprints are known to not meet the sizes they should according to IPC documentation.
1. For some part types, the body size is quite wrong. There could be issues because the pad sizes don't match the part size. If there are soldering issues, I'd be very sad since this is a known problem and should have been corrected so users of the library weren't affected. Nobody wants to pay money and get back a bad board that is due to our library. Yes, this is only one of many known issues but it's one I found and care about.

cpresser:
Sorry for the delay, I do have this open in a browser-tab for quite a while now.
It is of medium size, so I guess It will be possible to get this done by september 23.
image


However, I will invest my time the next few days into finishing some footprint-reviews I already started. My brain has a limited stacksize.

scripts/SMD_chip_package_rlc-etc/size_definitions/size_capacitor_chip.yaml
```diff
@@ -0,0 +1,143 @@
...
+    body_length:
+        minimum: 1.02
+        maximum: 1.32
+    body_width:
+        minimum: 0.77
+        maximum: 1.27
+    terminal_length:
+        minimum: 0.13
+        maximum: 0.38
+    size_info: '(Body size source: IPC-SM-782 page 76, https://www.pcb-3d.com/wordpress/wp-content/uploads/ipc-sm-782a_amendment_1_and_2.pdf)'
```
cpresser, line 13: This one is smaller than 0603 and should go into the respective file.

scripts/SMD_chip_package_rlc-etc/size_definitions/size_capacitor_chip.yaml
```diff
@@ -0,0 +1,143 @@
...
+    body_length:
+        minimum: 4.2
+        maximum: 4.8
+    body_width:
+        minimum: 3
+        maximum: 3.4
+    terminal_length:
+        minimum: 0.25
+        maximum: 0.95
+    size_info: '(Body size source: https://www.nikhef.nl/pub/departments/mt/projects/detectorR_D/dtddice/ERJ2G.pdf)'
```
cpresser, line 83: Why did you chose that datasheet? It is also in the one from AVX and in IPC-SM-782.
IPC has which is identical, So I suggest to take that value.
```yaml
body_length: 4.2 .. 4.8
body_width: 3.0 .. 3.4
terminal_length: 0.25 .. 0.95
```

scripts/SMD_chip_package_rlc-etc/size_definitions/size_capacitor_chip.yaml
```diff
@@ -0,0 +1,143 @@
...
+    code_metric: "2012"
+    body_length:
+        minimum: 1.8
+        maximum: 2.2
+    body_width:
+        minimum: 1.05
+        maximum: 1.45
+    terminal_length:
+        minimum: 0.25
+        maximum: 0.55 # only deviation from IPC is reducing this from 0.75mm
```
cpresser, line 40: Why that deviation? (IPC has 0.75)

scripts/SMD_chip_package_rlc-etc/size_definitions/size_capacitor_chip_smaller_0603.yaml
```diff
@@ -0,0 +1,12 @@
...
+    code_imperial: "0402"
+    code_metric: "1005"
+    body_length:
+      minimum: 0.9
+      maximum: 1.1
+    body_width:
+      minimum: 0.4
+      maximum: 0.6
+    terminator_spacing_min: 0.3
+    terminator_spacing_max: 0.65
```
cpresser, line 11: That is wrong. And uses the old style.
`terminator_spacing: 0.1 .. 0.3` according to the spec.

scripts/SMD_chip_package_rlc-etc/size_definitions/size_capacitor_tantal.yaml
```diff
@@ -1,4 +1,4 @@
-SMD_1608-08:
+C_1608-08:
```
cpresser, line 1: There is no `1608-08` in that datasheet. Only `1608-09`. The difference is the height.

scripts/SMD_chip_package_rlc-etc/size_definitions/size_default_chip.yaml
```diff
@@ -60,19 +62,20 @@ SMD_1210:
...
-      nominal: 4.5
-      tolerance: 0.2
-    body_width:
-      nominal: 1.6
-      tolerance: 0.2
-    terminal_length:
-      nominal: 0.7
-      tolerance: 0.3
-    size_info: '(Body size source: https://www.modelithics.com/models/Vendor/MuRata/BLM41P.pdf)'
+#Could not find any diode, fuse, or LED in this body size
```
cpresser, line 65: The murata datasheet does have a ferrite bead in that size.
LEDs also exist: https://www.mouser.de/datasheet/2/216/APDA1806ZGCK-1374637.pdf
For Diodes I agree, at least a quick search did not return anything.

scripts/SMD_chip_package_rlc-etc/size_definitions/size_inductor.yaml
```diff
@@ -1,4 +1,18 @@
-SMD_1008:
+L_0805:
```
cpresser, line 1: I don't see L_0603 here. So The one from size_default_chip is used.
Is that by design? Or should we also include a explicit 0603 inductor?

scripts/SMD_chip_package_rlc-etc/size_definitions/size_resistor_chip_smaller_0603.yaml
```diff
@@ -0,0 +1,12 @@
...
+    code_imperial: "0402"
+    code_metric: "1005"
+    body_length:
+      minimum: 1.0
+      maximum: 1.1
+    body_width:
+      minimum: 0.48
+      maximum: 0.6
+    terminator_spacing_min: 0.4
+    terminator_spacing_max: 0.7
```
cpresser, line 11: Why not have `terminal_length` here?
And why use the old style?

scripts/SMD_chip_package_rlc-etc/size_definitions/size_resistor_chip_wide_body.yaml
```diff
@@ -13,7 +13,7 @@ SMD_0612:
     ipc_reference: "ipc_spec_larger_or_equal_0603"
     size_info: '(Body size source: https://www.vishay.com/docs/20019/rcwe.pdf)'
 
-SMD_0815:
+R_0815:
```
cpresser, line 16: The datasheet for that no longer exists.

I suggest to use this one: https://www.bourns.com/docs/Product-Datasheets/CRK.pdf
image

cpresser (commented):
I did not review the `ipc7351_smd_two_terminal.yaml` or `ipc7351B_smd_two_terminal.yaml` files since they are just renamed.

Regarding the tantalum specs file
* It uses an outdated style for size definitions
* There are remnant lines saying stuff about Kemet
* The code-letter is AVX, but the size-source is from vishay
* IMHO, that whole file needs some rework, I did not review it to the end since it is inconsistent.

scripts/SMD_chip_package_rlc-etc/SMD_packages.yaml
```diff
@@ -72,7 +72,7 @@ capacitor_smaller_0603:
 
 capacitor_smaller_0402:
     fp_lib_name: 'Capacitor_SMD'
-    size_definitions: ['size_default_chip_devices_smaller_0402.yaml']
+    size_definitions: ['size_default_chip_smaller_0402.yaml']
```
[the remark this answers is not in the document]
evanshultz: OK. Done.

scripts/SMD_chip_package_rlc-etc/size_definitions/size_capacitor_chip.yaml
evanshultz: That was the datasheet already there, and I don't know why. I also confirmed the values are in IPC-SM-782 and updated the link so that used for other chip Cs.

scripts/SMD_chip_package_rlc-etc/size_definitions/size_capacitor_chip.yaml
evanshultz: It's been quite a while, but as best I can recall it was to maintain the footprints more close to the existing ones. Rene had compiled a list of Cs when the R body dimensions were being used and there were issues. See https://github.com/KiCad/kicad-footprints/issues/711 and https://github.com/KiCad/kicad-footprints/pull/712. I noted this above so Rene may give some feedback, but at this point if you want to go strictly with IPC that's fine with me. In fact, I like it even better but we should be prepared for any comments about it here or on the forum. So we will need to evaluate the change in pad size/location before deciding, I think. Want me to take a look at that or do you want to?

scripts/SMD_chip_package_rlc-etc/size_definitions/size_capacitor_chip_smaller_0603.yaml
evanshultz: Note we're looking at terminal length in some definitions and terminal spacing in others. It's been a while so I didn't catch that at first. With L=0.9 to 1.1mm and T=0.1 to 0.3mm, the spacing I calculate is 0.3mm min and 0.9mm max. As mentioned above, if there is a notable change to what I expect is a very common footprint we should carefully evaluate and be prepared to answer any questions from users of the library. Or leave ourselves a comment for later and wait for v6's library to do it. I'm just saying to be thoughtful if changing high-impact footprints.

Note that this dimension style was carried over from https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/SMD_chip_package_rlc-etc/size_definitions/size_default_chip_devices_smaller_0603.yaml.

Perhaps we should look at this a bit closer. Probably we should consolidate on the same style everywhere where possible, and I see a lot of hits for 'spacing' if I search for that term in the YAML files. Many of them are in comments, however. I'm thinking to remove all the comment lines and then see if we can consolidate on terminal length and not spacing. That will require looking into the code a bit so that's why I'm asking first.

scripts/SMD_chip_package_rlc-etc/size_definitions/size_capacitor_tantal.yaml
evanshultz: I'm thinking the datasheet may have changed over time. Look at the dimensions for the 1608-10 in the file comment and compare with the latest datasheet. Two mistakes are too many to consider as simply a typo IMO. I reckon it makes sense to update this to the current 1608-09 in the datasheet and leave the 1608-08 footprint in our library for now, removing it in the v6 library. OK?

scripts/SMD_chip_package_rlc-etc/size_definitions/size_default_chip.yaml
evanshultz: This comment was already there, and it doesn't mention inductors (which should cover ferrite beads). As commented on line 126 of SMD_packages.yaml, for inductors there is a specific L_1806 definition in size_inductor.yaml using the IPC-SM782 body dimensions (annoying only given as 4516 and not showing the 1806 equivalency). Good catch on the LED; I will add it in a separate commit soon. You didn't mention anything about fuses but I quickly checked DigiKey and didn't see any with 1.8x0.6mm body size.

scripts/SMD_chip_package_rlc-etc/size_definitions/size_inductor.yaml
evanshultz: See page 80 of the IPC-SM-782 doc linked: there is no 0603 body size for chip inductors. So it is intentional. I added a more explicit comment in SMD_packages.yaml about this.

scripts/SMD_chip_package_rlc-etc/size_definitions/size_resistor_chip_smaller_0603.yaml
evanshultz: As mentioned above, probably we should consolidate but I didn't touch things that were outside of the scope of correcting C and L body size if I could help it. Since I wrote more about this above, I'll mark this comment as resolved and we can discuss in the other comment about this topic.

scripts/SMD_chip_package_rlc-etc/size_definitions/size_resistor_chip_wide_body.yaml
evanshultz: OK. Updated. Note that there are dimensional changes here, especially a much tighter terminal width dimension so the resulting footprint will change. Again, shall I evaluate before doing this or would you like to?

Also, please take a look at my question at https://github.com/pointhi/kicad-footprint-generator/pull/438#issuecomment-538133156. Should those all be merged into one YAML file? I'm not sure why wide body needs a separate YAML file, but if there's an advantage to breaking them up I have no problem leaving it alone.

scripts/SMD_chip_package_rlc-etc/size_definitions/size_default_chip.yaml
evanshultz: The IPC-7351 fillet goals are very specific. The ones for rectangular-end components must be the one to use here, and this is the only reference to LEDs I could find (see 'LEDM' in section 8.4:
image

I don't see anything that makes me think this LED footprint wouldn't work based on the image in the datasheet, as it has leads wrapping around just the ends like a chip resistor, so I added it. But like I said, I'm not 100% sure. The resulting pads are a bit smaller than shown in that datasheet but fairly close (0.85x0.75mm) so it seems OK.

evanshultz:
@cpresser 
Thank you!

Yes, let's wrap everything else up and then go back to the tantalum YAML file.

scripts/SMD_chip_package_rlc-etc/size_definitions/size_capacitor_chip.yaml
evanshultz: Currently, our C 0805 (non-handsolder) footprint is:
- Pad centers: +/-0.9375mm
- Pad width: 0.975mm
- Pad inside: 0.9mm
- Pad outside: 2.85mm

The script currently results in:
- Pad centers: +/-0.95mm
- Pad width: 1.0mm
- Pad inside: 0.9mm
- Pad outside: 2.9mm

If I use the 0.75mm terminal length from IPC I get this:
- Pad centers: +/-0.875mm
- Pad width: 1.15mm
- Pad inside: 0.6mm
- Pad outside: 2.9mm

So we kinda end up back where we were when Wayne filed the issue I linked above. The pad spacing is quite small. The adjustment I made here was to roughly match the existing 0805 cap footprint.

We could enter the IPC-SM-782 terminal spacing instead of terminal width, and then we get this:
- Pad centers: +/-0.8mm
- Pad width: 1.3mm
- Pad inside: 0.3mm
- Pad outside: 2.9mm

So even worse, at least from the original comment made by Wayne. My solution was to align with the IPC numbers for everything but this dimension, which happily resulted in good matching with what Rene pulled out of datasheets once the proper capacitor body dimension info was used.

Do we have any other reference? Something to check our dimensions against? Another IPC-compliant footprint generator? Wayne's concern seemed valid to me and once Rene got actual part dimensions this seemed reasonable. I wouldn't object to making further changes, but without a reason I'm somewhat resistant to matching the IPC numbers here and going against what feels like the right approach (as opposed to every other situation involving our IPC generators where I saw go with it and that's how it should be!).

Thoughts?

scripts/SMD_chip_package_rlc-etc/size_definitions/size_capacitor_chip_smaller_0603.yaml
evanshultz: OK. A little bit more about this. The 0.1-0.3mm you wrote above is terminal length. The given dimensions of 0.3-0.65mm are terminal spacing. Both are in IPC-SM-782 so one would think we are good. I believe what you're pointing out is a difference between terminal _spacing_ vs terminal _length.

That being said, we do end up with slightly different footprints if we use spacing vs length. Here is current:
- Pad centers: +/-0.45mm
- Pad width: 0.62mm
- Pad inside: 0.28mm
- Pad outside: 1.52mm

And if I use terminal length from IPC instead of spacing:
- Pad centers: +/-0.48mm
- Pad width: 0.56mm
- Pad inside: 0.4mm
- Pad outside: 1.52mm

Here the pad spacing is farther apart. Above it was closer. I'm wondering if we should standardize on spacing or length for the IPC body sizes, but I don't know which is better. Ugh. Again, I'll ask your thoughts.

scripts/SMD_chip_package_rlc-etc/size_definitions/size_resistor_chip_wide_body.yaml
evanshultz: As I did above, here is the current footprint:
- Pad centers: +/-0.85mm
- Pad width: 1.35mm
- Pad inside: 0.35mm
- Pad outside: 3.05mm

And with the dimensions from above:
- Pad centers: +/-1.0375mm
- Pad width: 1.125mm
- Pad inside: 0.95mm
- Pad outside: 3.2mm

That is a huge increase in pad minimum distance! I'm thinking perhaps a bit more surveying of the parts in the market to ascertain more of an average might be in order. You?

evanshultz:
While we're talking about this, I really dislike some of the naming here. I think it's because this was done first and Rene did things a little better with the gullwing and SON generator scripts.
1. Each of the YAML files in /size_definitions starts with 'size_' which seems unnecessary. Remove it?
1. The `series_config` parameter is the same file name as the `global_config` parameter. I would like to rename to `package_config_KLCv3.yaml` like the SON script. OK?
1. The folder and script name of `SMD_chip_package_rlc-etc` is just ugly to type. This script does chip and tantalum capacitors body types, and other part types besides Rs and Ls and Cs, but there has to be a better name. Any ideas?

scripts/SMD_chip_package_rlc-etc/size_definitions/size_resistor_chip_wide_body.yaml
cpresser: Here is a comparison of the recommended burns landing pattern (top) with the current master footprint (bottom). 
image

The only other manufacturer on digikey I could find was this - which is identical to the Burns part.
image
https://www.susumu.co.jp/common/pdf/n_catalog_partition07_en.pdf

Asssuming the extreme values on tolerance (Burns) the resistor would be 3.45mm wide, with the terminal 0.7mm. So the inner-most-edge of metal is 3.45/2 - 0.7 = 1.025. 
So a pad-inside position of 0.95mm seems perfectly fine for me.

scripts/SMD_chip_package_rlc-etc/size_definitions/size_capacitor_tantal.yaml
cpresser: A quick search shows that both `1608-08` and `1608-09` are sold.
I am resolving this conversation to reduce noise now. Its part of the `tantal` file that we should discuss as a whole in dedicated PR.

scripts/SMD_chip_package_rlc-etc/size_definitions/size_capacitor_chip.yaml
cpresser: I opened my old Eagle 6.3 to see what is in there.
* Pad centers +/- 0.95mm
* Pad width: 1.5mm
* Pad inside: 0.6mm
* Pad outside: 3.2mm

Some more stuff I researched:
* Kemet Datasheets have exactly the same dimensions and tolerances as IPC-SM-782
* Yaego Datasheets have exactly the same dimensions and tolerances as IPC-SM-782. But they also sell caps with tighter tolerances where the spacing between the pads could be considerably larger

Yaego also has a nice table for footprints:
image

Following their recommendation (Nominal Density) we would get:
* Pad centers +/- 0.9mm
* Pad width: 1.15mm
* Pad inside: 0.325mm
* Pad outside: 2.95mm

scripts/SMD_chip_package_rlc-etc/size_definitions/size_resistor_chip_wide_body.yaml
evanshultz: The 3.75+/-0.3mm matches on Bourns and Susumu, but annoyingly the other dimensions is 2mm vs 2.3mm. I also found https://www.seielect.com/catalog/sei-csr_csrn.pdf which is 3.75x2mm, matching Susumu. And 80mil is 2.032mm, so 2mm nominal is much closer to matching the '08' in the package name than 2.3mm. Which dimensions should you like to go with?

scripts/SMD_chip_package_rlc-etc/size_definitions/size_capacitor_chip_smaller_0603.yaml
cpresser: > I believe what you're pointing out is a difference between terminal spacing vs terminal _length.
Yes, that is what my original comment was about. It is inconsistent with other definitions using `terminal_length`.

> I'm just saying to be thoughtful if changing high-impact footprints.

:+1: 

> Perhaps we should look at this a bit closer. Probably we should consolidate on the same style everywhere where possible, and I see a lot of hits for 'spacing' if I search for that term in the YAML files. Many of them are in comments, however. I'm thinking to remove all the comment lines and then see if we can consolidate on terminal length and not spacing. That will require looking into the code a bit so that's why I'm asking first.

I would love to see the same yaml style used everywhere. That makes reviews like this much easier. Less confusion. Which we should prevent with high-impact-footprints.

Looking at the Kemet Datasheet again, they get for IPC-Nominal
*   Pad centers: +/-0.45mm
*   Pad width: 0.62mm
*   Pad inside: 0.28mm
*   Pad outside: 1.52mm

I also looked for Wuerth, they have good datasheets and Sales-reps I could call: https://www.we-online.de/katalog/datasheet/885012205012.pdf
*   Pad centers: +/-0.75mm
*   Pad width: 0.5mm
*   Pad inside: 0.5mm
*   Pad outside: 1.50mm

One more Data point from taiyo-yuden. They clearly sell parts with different `terminal_length` values.
image

scripts/SMD_chip_package_rlc-etc/size_definitions/size_default_chip.yaml
cpresser: TBH i have no clue which footprint to use for a ferrite bead. I just had to look which fp-filters are set for the symbol.

Regarding the LED I made a mistake. It is a side-emitting LED. Its only `1806 Metric` when viewed from the side.
But it seems you fell for it as well
```yaml
LED_1806:
    code_imperial: "1806"
    code_metric: "4516"
    body_length:
      nominal: 1.8
      tolerance: 0.15
    body_width:
      nominal: 0.6
      tolerance: 0.1
    terminal_length:
      nominal: 0.3
      tolerance: 0.15
    ipc_reference: "ipc_spec_larger_or_equal_0603"
    size_info: '(Body size from: https://www.mouser.de/datasheet/2/216/APDA1806ZGCK-1374637.pdf)'
```
That does not make sense. `code_imperial` is 1806, but the metric size is 1806.

scripts/SMD_chip_package_rlc-etc/size_definitions/size_capacitor_chip.yaml
evanshultz: Good info. Thanks! Do you have links? Especially for the tighter-tolerance part. (I'm just lazy but I should search myself.)

The footprint in that screenshot looks like quite close to the IPC one, then.  And they only note IPC-7351, not IPC-73651B, so that could account for the slight difference. The fillet goals do tend to change a bit as the doc was updated.

Here is an image, top to bottom, of 0603, 0805, 0805 using strictly IPC, and 1206:
image

I'm still a bit concerned. Wayne's comments I linked above about the 0805 cap footprint still apply, even with the corrected dimensions. And especially since there are caps in the market with tighter tolerances. Should we have multiple 0805 cap footprints? Or adjust this one dimension from IPC as I originally was thinking? Something else?

scripts/SMD_chip_package_rlc-etc/size_definitions/size_default_chip.yaml
evanshultz: Ahh! Yes. Good catch. I'll undo that.

cpresser:
> 
>     1. Each of the YAML files in /size_definitions starts with 'size_' which seems unnecessary. Remove it?

Agree.

> 
>     2. The `series_config` parameter is the same file name as the `global_config` parameter. I would like to rename to `package_config_KLCv3.yaml` like the SON script. OK?

I don't quite understand that.
TBH the difference between `series_config` and `global_config` is not clear from the command-line help. It does not really describe what it does.

> 
>     3. The folder and script name of `SMD_chip_package_rlc-etc` is just ugly to type. This script does chip and tantalum capacitors body types, and other part types besides Rs and Ls and Cs, but there has to be a better name. Any ideas?

Agree. Here are some of my ideas. None of them sound really appealing to me. Perhaps they give you the right inspiration.
* Package_Chip_SMD
* SMD_2Terminal_Packages
* Generic_Chip_SMD
* Generic_2Terminal_SMD

cpresser:
One more point, should we move the discussion about the optimal pad geometry of the high-impact footprints somewhere else?
This PR is getting quite cluttered. Also for documentation issues it would be nice if the discussion can easily be found.

scripts/SMD_chip_package_rlc-etc/size_definitions/size_capacitor_chip_smaller_0603.yaml
evanshultz: Heh. So the Kemet is similar to using terminal spacing and the Wuerth is similar to using terminal length.

In the Taiyo Yuden datasheet, it looks like that is different heights, not terminal widths. Part height is T.

Here are some posts from Tom Hausherr, who really, knows, in order of date. I haven't fully digested it yet but it looks like there was some grappling with this also and they settled on tweaking the IPC 7351 numbers to get a single footprint:
https://www.pcblibraries.com/forum/standard-chip-component-dimensions_topic1729.html
https://www.pcblibraries.com/forum/new-2016-solder-joint-goals_topic1921.html
https://www.pcblibraries.com/forum/ipcjstd001-chip-component-solder-joints_topic2588.html

scripts/SMD_chip_package_rlc-etc/size_definitions/size_capacitor_chip.yaml
cpresser: https://www.yageo.com/upload/media/product/productsearch/datasheet/mlcc/UPY-GPHC_X7R_6.3V-to-50V_18.pdf
image

evanshultz:
global_config is for the entire library, which all scripts can use. series_config is only for a particular generator (part/package type). It's simple and I just did it so let me know if you have an issue. We can now make further changes for all generator scripts since they will be aligned.

scripts/SMD_chip_package_rlc-etc/size_definitions/size_capacitor_chip_smaller_0603.yaml
cpresser: I was to quick to judge the Taiyo Yuden picture - my bad. But the height also seems to be an important thing.

From https://www.pcblibraries.com/forum/standard-chip-component-dimensions_topic1729.html
> Good luck with finding a single footprint to successfully accommodate every 0603 package. The Height is the main driver of this component lead form.
 > Issue: if you use a 0603 land pattern considering the highest package, you may be dumping too much solder on an 0603 that is 50% of that height, introducing potential tomb-stoning. 

From my personal experience:
This is especially a problem with capacitors. Chip-resistors seem to have mostly the same height.

This is an optimization problem:
* From all commercially available 0402 caps most should show perfect solder results
* The distance between pads should be maxed
* The courtyard should be minified
* Users should not question the footprint on first sight

cpresser:
> global_config is for the entire library, which all scripts can use. series_config is only for a particular generator (part/package type). It's simple and I just did it so let me know if you have an issue. We can now make further changes for all generator scripts since they will be aligned.

Sorry, my comment also was not clear. I was referencing to the help text of the generator. Which currently has those two lines:
`the config file defining how the footprint will look like. (KLC)`
`the config file defining series parameters.`
From what I understand they have the same effect, but series overrides global. That should be reflected in the help text.

scripts/SMD_chip_package_rlc-etc/size_definitions/size_resistor_chip_wide_body.yaml
cpresser: I think it is reasonable to go with 2.0mm nominal.
These resistors sit in lots of paste and are relatively heavy with lots of leverage, I don't expect them to tombstone even when the pad geometry is bad.

I suggest to add part of of this discussion as comment to the .yaml file. Reference both the sumusu and sei datasheets. That ensures users, ourselves, and future contributors can understand why that value is there.

Unfortunately the recommended landing pattern of sei is quite different.
I would suggest to let the generator do its job. Then we can inspect the resulting footprint and compare it with sei and burns recommended footprints.

evanshultz: Using 2+/-0.2mm gives:
Pad centers: +/-0.9375mm
Pad width: 1.125mm
Pad inside: 0.75mm
Pad outside: 3mm

I think that should be fine based on the available 0815 resistors we found so I'm resolving this. I've added comments.

evanshultz:
For the global_config vs series_config, this text matches the SON generator. I agree better wording could be used. Would it be OK to shelve this and update all the generators together in another PR?

Here are the three IPC 7351B tables that are specific to this generator:
image
image

It handles more than just chip packages. And SMD crystals and aluminum electrolytic caps also have 2 pins and a different table (I previously implemented the electrolytic cap IPC generator). Maybe `ipc_chip_molded.py`? The others are ipc_ plus the package types they handle.

scripts/SMD_chip_package_rlc-etc/size_definitions/size_capacitor_chip_smaller_0603.yaml
evanshultz: Yes, but the 5th comment from the 2nd link, noting that I posted them in order or date above, says that height was ruled out because it was too specific and wouldn't allow multiple sourcing with only one ideal footprint. Again, that was from 2016 and it sounds like that is the preferred direction but for now more sloppiness is allowed. The last link I posted is less than a year old and seems to confirm that.

I agree with what you said. The generators are available if anyone wants to utilize them, but let's try to go for the 80/20 rule and support at least 80% of users and situations without question.

I would like to change from using terminal spacing to terminal length for the 0402 part, in combination with the IPC size info. See my edited initial reply above where I added the existing footprint info. Are you OK with that?

scripts/SMD_chip_package_rlc-etc/size_definitions/size_capacitor_chip.yaml
evanshultz: So the 0805 body size (length and width) tolerance drops from +/-0.2mm to +/-0.1mm, but the terminal length stays the same. (And just for kicks, the 0402 shows a terminal length of 0.15-0.35mm. IPC shows 0.1-0.3mm. Not tighter, just shifted.)

If I plug that in, using 0.75mm max terminal length per IPC, I get:
Pad centers: +/-0.8625mm
Pad width: 1.125mm
Pad inside: 0.6mm
Pad outside: 2.85mm

Not any substantial change.

I'm still uncomfortable with such closely-spaced pads. I prefer my slight adjustment for max terminal length from 0.75mm to 0.55mm. IMHO this makes our library most compatible for the most users in the most circumstances.

cpresser: Agree, we should hack/change that footprint to get more spacing between the pads.
0.55mm seems like a good choice.

scripts/SMD_chip_package_rlc-etc/size_definitions/size_capacitor_chip_smaller_0603.yaml
cpresser: Agree. IPC size info and use `terminal_length` everywhere.

evanshultz: Done.

evanshultz:
OK. I think that wraps up everything for chip parts. Yay!

While tantalum caps are out-of-scope for what I intended to do here, what are you thinking?

cpresser:
Proposed process to continue with this PR and the footprints. We have a little more than 1 week left to get footprints merged for 5.1.7:
* Merge this generator+yaml changes even when we are not 100% sure about the high-impact (0402 0603 0805) RLC footprints.
* Compare and discuss the resulting footprints in https://github.com/KiCad/kicad-footprints/pull/2280
* When changes are needed, open a new PR for footprint-generators, iterate the process



> For the global_config vs series_config, this text matches the SON generator. I agree better wording could be used. Would it be OK to shelve this and update all the generators together in another PR?

Yes. We are building quite a backlog. I am going to open an issue for that right away so it does not get lost.

evanshultz:
I find terminal spacing in the following YAML files:
- resistor_chip_smaller_0603.yaml (only 0402)
- default_chip_smaller_0603.yaml
- default_chip.yaml
- capacitor_tantal.yaml

The first one is simple enough so I'll share the difference when using terminal length here:
- Pad centers: +/-0.485mm to +/-0.51mm
- Pad width: 0.59mm to 0.54mm
- Pad inside: 0.38mm to 0.48mm
- Pad outside: 1.56mm to 1.56mm

image
Current is above and new is below. In an absolute sense it a relatively large change but it because in comparison it's not as much. Pad spacing is much wider than the Kemet datasheet you shared at https://github.com/pointhi/kicad-footprint-generator/pull/438#discussion_r488921448, though. And compared with https://www.yageo.com/upload/media/product/productsearch/datasheet/mlcc/UPY-GPHC_X7R_6.3V-to-50V_18.pdf the min lead spacing could be 0.25mm (0.95mm body length - 2 * 0.35mm terminals) so even our existing footprint fails there. I'm not totally sure about this one. Here IPC pads seem to far apart compared to this datasheet while with 0805 they were too close together...

IPC-SM-782 has 3216, 3528, 6032, and 7343 tantalum cap body sizes.

For the others, however, there is no 'default' part in IPC-SM-782 so we don't really have any reference to use that I see. The link http://www.tortai-tech.com/upload/download/2011102023233369053.pdf is referenced for many of the non-tantalum parts but the link doesn't work so I'm not sure where to go for those. Thoughts? Just leave them if we don't have a reason to change?

And also, rename the generator. And then any tantalum stuff if you want to look into it.

cpresser:
> 
> And also, rename the generator. And then any tantalum stuff if you want to look into it.

Tantalum: not as part of this PR. Rename: yes please.

The `R_0402` change looks good to me at first glance. I like that there is more space between pads. That allows to route a signal between them.

I would like to merge this ASAP. And if necessary open another PR for fixes. See my answer here https://github.com/pointhi/kicad-footprint-generator/pull/438#issuecomment-693662901
Do you agree, or do you want to discuss all the high impact footprints as part of this PR?

evanshultz:
0402 resistor dimensions updated. One extraneous bit of info found and deleted. And file/folder renamed. I'm open to comments on the name as I'm not really in love with it.

Now that I also discovered that IPC-SM-782 has some tantalum cap body sizes I think perhaps it is worth taking a look at. Perhaps I should start by updating the format and then comparing the existing footprints to the ones that we would get if we used IPC dimensions?

@poeschlr 
http://www.tortai-tech.com/upload/download/2011102023233369053.pdf is used as a reference many times here, but that link is dead. Do you have the document or, if it's some kind of standard, can you share what it is? I have no idea and it may help.

chschlue:
I don't have that Tortai document but I may be able to add confusion by posting IEC 60115 cap sizes:
iec60115-8

cpresser:
> 0402 resistor dimensions updated. One extraneous bit of info found and deleted. And file/folder renamed. I'm open to comments on the name as I'm not really in love with it.

The name `chip_molded` sounds like it is about molded chips. 
Chip-Capacitors and Resistors are not molded, so that does not sound right.

Or should I read the name as `molded or chip?`

> Now that I also discovered that IPC-SM-782 has some tantalum cap body sizes I think perhaps it is worth taking a look at. Perhaps I should start by updating the format and then comparing the existing footprints to the ones that we would get if we used IPC dimensions?

Looking at the time-frame I would suggest to postpone that until we have the basic RLC stuff merged into kicad-footprints.

> @poeschlr
> http://www.tortai-tech.com/upload/download/2011102023233369053.pdf is used as a reference many times here, but that link is dead. Do you have the document or, if it's some kind of standard, can you share what it is? I have no idea and it may help.

I assume it is the IPC-SM-782 doc (searched for the filename)
https://de.scribd.com/document/369788654/2011102023233369053-pdf

cpresser:
I have one more question: `IPC-SM-782` is really old, and according to the IPC website superseded by `IPC-7351`. Why are we using the tables from the old document then?

evanshultz:
I was thinking 'molded or chip' as the generator name since it handles chip parts and molded body parts, as described in the three IPC 7351 tables I pasted above. But I'm not in love with the name and I'd like to find something better. That was a first attempt and hopefully not the final name.

Sure thing. We can delay tantalum caps for later.

Good idea to search for the file name. However, there is another link used for IPC-SM-782 and that IPC doc doesn't have fuses or diodes or LEDs in it so I'm still a bit confused.

IPC-7351 doesn't have any part body size info. It only has tables for fillet goals and calculations for determining pad size. So that's why we couldn't use it as a reference for body dimensions. Seems strange, doesn't it?

cpresser:
Comparison of C and R, high-impact footprints **master** (left)  vs. **generator output** (right)

| Size |  C(master), C(#438), R(master), R(#438) |
|------|----|
|0805 | image |
| 0603 | image |
| 0402 | image |
| 0201 |image |

evanshultz:
Thanks! Just a hair of change on a few footprints. Did you see anything concerning?

Are you ready for me to drop these footprints into the footprint PR so we can compare with master over there?

I updated the script file name to what I intended to submit earlier. But I'd still like something better. And maybe the file `part_definitions.yaml` should be something better, like 'family_definitions'? Any ideas?

Lastly, I took a quick look at the tantalum stuff just to quickly make it less ugly. The YAML file seems clean and uniform now and doesn't affect the footprints (I did a spot check but once this is settled I'll update the footprints in the footprint PR so we can check all of them). However, something does concern me: lines 101 and 113 of part_definitions are not using the tantalum fillet goal table. If I change that, the footprint changes wildly. I didn't chase it down in case you want to defer the tantalum stuff for later, but removing each ipc_spec_tantalum key in the tantalum had no effect and it's those two lines driving the pad size calculations. I would expect to see ipc_size_tantalum on those two lines and that changing it for a particular size definition changes the footprint, but neither of those is working. So I'm a bit concerned we have tantalum cap footprints that aren't what we advertise. Let me what, if anything, you want to do about this now.

cpresser:
> Thanks! Just a hair of change on a few footprints. Did you see anything concerning?

Nope, the R & C footprints seem fine. 
* The changes are hardly visible
* I can get a trace through each of them
* They haven't gotten bigger (0805 did, but only 0.04mm in both directions)

> Are you ready for me to drop these footprints into the footprint PR so we can compare with master over there?

Yes, lets do that!

> tantalum

* You missed to remove one `n` in `ipc7351_smd_two_terminal.yaml`
* I did a little bit of chasing. Its the `heel` parameter in line 88. Change it to anything not positive and the pads don't crash together. I tried to follow where that parameter, but got lost in `scripts/tools/ipc_pad_size_calculators.py`. I have no clue what the functions there are supposed to do. They are in dire need of documentation and test-cases.
* It seems to me the generator is buggy here. 


From https://content.kemet.com/datasheets/KEM_T2005_T491.pdf page 15
image

The (current) KiCad footprint is quite different: 
image

| 3528-21 | W | L | S |
|--|--|--|--|
|Kemet| 2.23| 1.8 | 1.2|
|KiCad| 2.25 | 1.325 | 1.34|

We are missing about 0.5mm on the L parameter, which corresponds to the 0.5mm heel in the IPC spec file.

cpresser:
Here is the equivalent info from AVX. I will add it to the table in the above post.
http://www.avx.com/docs/techinfo/TechSumAppGuide.pdf Page 16
image

evanshultz:
I just pushed the footprints into that PR. Seems to me everything looks good, but take a look through the diff and see what you see.

Reviewing those footprints also made two things pop out at me which wouldn't take long to incorporate here but I suspect it's best to do as a later step:
1. There are wide body capacitor footprints available but they aren't here.
1. The inductor footprints are, at least in some cases, used for ferrite beads. I view ferrite beads and indutors as different parts, and they have unique schematic symbols in our library. We have a Ferrite_THT footprint library. Should the chip 'inductor' footprints go into Ferrite_SMD or somewhere else instead of being with inductors?

I'll look at the tantalum stuff a bit later today. We can always decide to punt on it if the current footprints are OK. Waiting also gives more time to get into https://github.com/pointhi/kicad-footprint-generator/pull/491 and (possibly... hopefully??) https://github.com/pointhi/kicad-footprint-generator/pull/439.

cpresser:
ad1: I vote for a different PR
ad2: There are both 
* chip ferrites for example murata BL
* and chip inductors (https://www.digikey.com/product-detail/en/tdk-corporation/MLZ2012N6R8LT000/445-6761-1-ND/2523582)
* The chip-inductors are usually higher, comparable to chip-capacitors (they are also manufactured with screen printing).
* I think we need more research into this topic.
* Thus, I vote for postponing that. Should be another PR.

evanshultz:
`ipc7351_smd_two_terminal.yaml` isn't being used (the B version should be used) but I fixed the typo. Thanks.

IPC 7351 give equations for determining pad size given package dimensions and fillet goals. The heel fillet is the inside fillet, which would be under the part body. I verified all the fillet goals so unless IPC7351B has a typo in it that is correct. And I've done hand calculations to verify the equations in the calculator script so I think that is correct as well. A mistake in the fillet goals seems most likely, or perhaps we're using the wrong table. Here is the table and part construction paragraph:
image
image

It sure seems like the right stuff. But I notice to phrases in parenthesis `to find G dim` and `to find Z dim`. That's strange. And backwards compared to the 'typical' equation usage (see lines 253-256 in `ipc_pad_size_calculators.py`). Swapping the numbers on lines 87 and 88 of `ipc7351B_smd_two_terminal.yaml` didn't quite do it, but it's much closer (pad centers at +/-0.7125mm with pad size 1.375x0.95mm). I gotta run right now but I'll keep on this a bit later. The equations aren't quite the same so I may need to read in toe and heel fillets and then swap the numbers in the `ipc_data` variable just for tantalum caps.

2. I'd rather not duplicate footprints. I thought KLC said this but I don't see it now. Anyway, it feels funny to me but if it works for inductor footprints to also be for ferrite beads then I'm OK leaving it alone.

Lastly, I happened to find this in IPC 7351B:
image

I haven't looked at that, but it may be another possible set of body dimension data. Just posting it her for posterity since I think we should finish this PR as it is.

cpresser:
Looks like I mixed up heel and toe in my head. Then the current behavior of the generator makes sense (pads grow inward). Swapping heel and toe might be what happened in that table. But that is a whole different story to investigate.

ad 2) I also don't see any reason to duplicate the footprints. In my personal use I selected resistor-footprints for ferrites so far (because I did not know better). They worked fine (a few thousand boards without problems).

Regarding that last block from `IPC-7351B`. Those are good points. We should discuss having different libraries for v6. 
Perhaps like so
* `Resistor_SMD_IPC7351_Nominial`
* `Resistor_SMD_IPC7351_MaxDensity`
* ...
I am not sure if that is a good idea. Lets discuss that outside of this PR as well.

cpresser:
I think this one is ready to merge. Do you want anything else done/checked @evanshultz 
It is quite late in my Timezone already, I will come back to this and the footprint PR in ~16 hours so we can merge.

evanshultz:
Right. Because the end of the lead is wrapped under the part on tantalum caps (and other molded body parts), the toe is under the part body instead of sticking out on the ends of the part. Confusing. But it finally clicked for me, too.

I checked the footprints at the end of https://www.vishay.com/docs/40182/tmch.pdf, and it's not too different. Pad centers just a bit off and the pads are a bit bigger. If I use the LMC (smallest pads) they are smaller than the datasheet. So with the tantalum fillet goals with heel and toe flipped I think we might be OK.

The only caveat are the 1608-08 and 1608-10 footprints where the pads are just too close with nominal fillet goals. I checked body dimensions and they appear correct. With LMC pads the pad inner gap is 0.2mm apart which should be fine. My thought with this would be to update the script to set a minimum pad spacing that automatically makes adjustments if the pads are too close. That distance would be in the series config YAML file.

However, that would take too much time. So I see a few options now:
1. Merge it just like it is. No tantalum cap changes.
1. Push updated tantalum caps except for the 1608 ones above.
1. Push all new tantalum caps, but make some simple manual adjustments to give more clearance on the 1608 ones. I suggest 0.2mm inner pad gap which is what we've done many other places (though not codified in KLC or elsewhere AFAIK).

Any of those are OK with me. Just let me know. But regardless, I will push another commit to this PR. Fix the extra 'n' on tantalum and a few other things that I figured out so far (even if he tantalum caps aren't re-generated). So please don't merge this first.

FYI, there already is an issue about having chip footprints with varying IPC densities: https://github.com/KiCad/kicad-footprints/issues/1836. And a PR to add that functionality to the script: https://github.com/pointhi/kicad-footprint-generator/pull/439. We can pick this up later. :)

evanshultz:
I pushed the tantalum updates, but I expect we will merge them as they are. And later take a closer look at the footprints and add some feature to adjust things if pad-pad spacing is too tight. But we'll see.

As noted, I didn't copy over tantalum footprints to the footprint PR. So if you want to hold off on other tantalum changes there are only two minor questions:
1. Generator script name. Is there something better?
1. Remove `_PadXX.XxYY.Y` in hand solder footprint names?

cpresser:
> 
> As noted, I didn't copy over tantalum footprints to the footprint PR. So if you want to hold off on other tantalum changes there are only two minor questions:
> 
>     1. Generator script name. Is there something better?
> 
>     2. Remove `_PadXX.XxYY.Y` in hand solder footprint names?

0. Regarding the tanatal caps: Lets keep them as they are. I am afraid we (mostly me) will fuck it up when trying to finish that in a rush.
1. How about `Package_2Terminal_SMD` ? Its not perfect, but the best I could come up with. If you want to stay with `chip_molded` I suggest to capitalize the first `c` as almost all other folders are captialized. 
Also, I would set the +x bit on the generator to make it executable.
2. I vote for **not now**. Lets not change their filenames for a minor release. I was somehow confused because I assumed that 5.x did not include the `_PadXX.XxYY.Y` information. But I was wrong there.
3. You did duplicate the `scripts/chip_molded/ipc_chip_molded.py` file in the last commit. Please remove the one that does not have generator in its name.

evanshultz:
I figured out what happened with the missing footprints. Since the YAML files for body size info has been split, new entries were needed in the YAML file that contains the info for each part type. I would like to simplify the part_definitions.yaml file which I think is possible right now. My fix for this issue added a lot of entries. If this PR can wait to be merged for a few hours I'll do that tonight when I have a chance to work on it.

1.  I went with `SMD_two_terminal_chip_molded`. There are crystals and SMD electrolytic caps and other 2-terminal SMD footprints, but chip and molded are two terms specific to this generator. So we could add a `SMD_2terminal_crystal` later, for example. And I realized this was like the existing IPC 7351 fillet goal files. It's awfully long, though.

We should also update the datasheets and dimensions for generic parts since some (all?) use R or C datasheets which don't apply to the diodes, fuses, LEDs, etc. I think we end up with a single YAML file for each part type and get rid of the generic YAML files entirely. But I don't think it's a showstopper to get this merged and it will affect footprints so let's not do it now.

evanshultz:
Good catch on the L_1806 footprint at https://github.com/KiCad/kicad-footprints/pull/2280#issuecomment-698017922 and fixed at https://github.com/KiCad/kicad-footprints/pull/2487. Pads end up 1.275x1.9mm at +/-2.1125mm. A little more narrow but a bit taller than the pads now. And wider than the actual part's body width now!

I was able to remove about half the entries in part_definitions.yaml by allowing an ipc_reference for one size definition in the YAML file to overwrite the general one for the entire part group. That was easy. Supporting a paste pad overwrite requires changing parameters in IPC calculator functions that I plan to pull out and share across all generators so I'm leaving it for now.  I checked each footprint compared to the current master, and excluding L_1806 only the timestamp changed (except for some castellated footprints; see below). So I think this change is good.

I also noticed there are 2816 diode, fuse, and LED footprints which should have been removed from the footprint repo. Oh well.

All the castellated footprints are wrong. The castellated table applies only the LCC footprints, not chips. So these footprints should be removed IMO. We are saying they are something when they clearly are not. What do you want to do about that? Here is the table and generic package drawing, from which you can compare the fillet goals with the YAML file to see for yourself:
image
image

If you have no further comments, I believe this can be merged. Or I can whack out the castellated stuff first. Let me know.

evanshultz:
@cpresser 
How do you want to proceed?

cpresser:
>
> Edit: Actually, I'd rather remove the castellated footprints, the 2816 ones, and update L_1806. And remove the castellated stuff from this PR before merging. Are you OK with that?

Yep, that sounds good.
Once that is done, I will give the whole thing another full review. I am losing track of individual changes/features.  During all the incremental reviews the stack in my brain got corrupted.

One option would be so squash all the commits and rework the commit message so it properly reflects all the changes. What do you think?

evanshultz:
I submitted https://github.com/KiCad/kicad-footprints/pull/2495 and made the corresponding cleanup changes to the script.

I'm fine with a squash merge. That's what I usually do in this repo. Maybe that's a mistake?

cpresser:
> I submitted [KiCad/kicad-footprints#2495](https://github.com/KiCad/kicad-footprints/pull/2495) and made the corresponding cleanup changes to the script.
> 
> I'm fine with a squash merge. That's what I usually do in this repo. Maybe that's a mistake?

Looks like my comment was not clear at all. Sorry about that. What I meant to say was:
"Please (as in optional, if you have the time and motivation for it) cleanup the git-history of this PR".
* Squash fixups, squash related commits
* Rework the commit messages
* Then force-push the results to this PRs branch
* That might be an insane amount of work to do properly, so one option would be to squash it into one commit
  * That would still require to write one nice commit message 

Reasons:
* There are currently 35 commits in this branch
* Most of them overlap each other or undo each other
* The scope of this PR changed as we were working on it
* I feel like I lost track of individual changes

evanshultz:
Oh. I see. I did misunderstand.

I usually squash merge in the web interface and can change the commit message then if I want. As a reviewer and merger (is that a word?), https://github.com/desktop/desktop/issues/3580 captures what I usually do.

I've not actually done all the stuff above. Doesn't look too tough, just new to me. I can figure it out. But let me work on the gullwing footprint name format thing first. Hang on...

cpresser:
The difference is squashing everything into one commit vs. explicitly choosing which commits to squash.

I have no clue how to clean up the history in a gui. In this case, something like `git rebase -i HEAD~36` is how I would to that on the commandline.

A quick google-search did return this video: https://mattstauffer.com/blog/squashing-git-commits-with-interactive-rebase/ that explains this workflow.

PR #439, opened 2019-10-05
Support generating two-terminal footprints with IPC density argument

evanshultz:
Branched from #438.

This adds a parameter to support IPC density settings. If the user does not enter a value for the density, nominal density is assumed and footprints are generated exactly as they are now (or, rather, would be if the aforementioned PR is merged). Adding an argument to this parameter generates footprints with correspondingly changed pads and a suffix on the footprints indicates the density.

It appears the density is used during pad calculation for the handsolder footprints and I've left that alone since it gives further options for handsoldering pads. Or, those footprints can be discarded if the density argument is given and only the non-handsolder footprints added to the library.

I feel like there might be a better way to sort out the density during argument parsing than what I added at the end of `SMD_package_rlc-etc.py`, but my brain feels empty now so I may need to come back to this later. Or get suggestions.

I also renamed the function `TwoTerminalSMDchip()` to `TwoTerminalSMD()` because the script can generate non-chip footprints.


scripts/SMD_chip_package_rlc-etc/SMD_package_rlc-etc.py
```diff
@@ -334,10 +344,23 @@ def generateFootprint(self, device_size_data, footprint_group_data):
...
     parser.add_argument('--series_config', type=str, nargs='?', help='the config file defining series parameters.', default='config_KLCv3.0.yaml')
-    parser.add_argument('--ipc_definition', type=str, nargs='?', help='the ipc definition file', default='ipc7351B_smd_two_terminal_chip.yaml')
+    parser.add_argument('--ipc_definition', type=str, nargs='?', help='the ipc definition file', default='ipc7351B_smd_two_terminal.yaml')
+    parser.add_argument('--ipc_density', type=str, nargs='?', help='IPC density level (L,N,M)')
     parser.add_argument('--force_rectangle_pads', action='store_true', help='Force the generation of rectangle pads instead of rounded rectangle (KiCad 4.x compatibility.)')
     args = parser.parse_args()
 
+    # if the user requests an IPC density, put that and footprint suffix in a list
+    # otherwise, use nominal density with no suffix
+    if args.ipc_density is None or args.ipc_density not in ['l', 'L', 'n', 'N', 'm', 'M']:
```
poeschlr:
```suggestion
    if args.ipc_density is None or args.ipc_density.lower() not in ['l', 'n', 'm']:
```

evanshultz: I think there's an even better way to do this. I believed this could be improved but my brain was feeling worn out as I mentioned above; hopefully now inspiration has struck.

poeschlr:
You might have done something wrong with git here. The list of changed files looks strange.

poeschlr:
Also a general thing: The size definitions used for this script are massive compromises as these sort of devices are not really standardized. I am therefore not really sure the official lib will benefit from having other densities of these footprints in it. I am however not opposed to adding this feature to the script to allow users to generate their own set of footprints with differing density levels (really only makes sense if the user then also adds their own size definition fitting their device)

cpresser:
@evanshultz while looking at #438  I also found this PR. The history of this one looks broken. I think its best to fix this branch after #438 is merged.

evanshultz:
@cpresser 
Yes, that was my intent. Maybe that wasn't clear above. Sorry.

PR #440, opened 2019-10-05
size definitions Tantal AVX F72/F65 series

lorem-ipsum:
This adds size definitions for the missing Tantal capacitors from the [F72 & F75 series](http://datasheets.avx.com/F72-F75.pdf). Only the F75 AVX-D was already present.


poeschlr:
thanks

PR #443, opened 2019-10-08
Add MSOP-8_3x3mm_P0.65mm

chschlue:
https://github.com/KiCad/kicad-footprints/pull/1901

https://www.jedec.org/system/files/docs/mo-187F.pdf

### DIFF
--- a/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/msop.yaml
+++ b/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/msop.yaml
@@ -3,10 +3,10 @@ FileHeader:
   device_type: 'MSOP'
 
 MSOP-8_3x3mm_P0.65mm:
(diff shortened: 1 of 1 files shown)


evanshultz:
1. Please update `size_source` to capture that this is `MO-187AA`. There are other variantions.
1. There is tolerance for the body size in the JEDEC doc: both width and height are +/-0.2mm (`aaa`).
1. Similarly, overall size tolerance is +/-0.25mm.
1. Can you include package height?

chschlue:
1. I know, but size_source is supposedly of {url} type. What am I supposed to do, add a pseudo page anchor?

evanshultz:
We did this on some other part that Rene added. I think he just overloaded the string since there was no validation. That probably works here since I don't believe a dedicated field for info has been added yet.

chschlue:
Done.

evanshultz:
Thanks! This doesn't have an EP so it's OK to do.

PR #445, opened 2019-10-16
Add VQFN-46

rleh:
Source/datasheet: http://www.ti.com/lit/ds/symlink/lp5036.pdf#page=59

VQFN-46-1EP_5x6mm_P0.4mm_EP2.8x3.8mm

VQFN-46-1EP_5x6mm_P0.4mm_EP2.8x3.8mm_ThermalVias

I am not sure how to add the thermal via pattern from the datasheet:

image


poeschlr:
thanks

PR #449, opened 2019-10-26
Adding TE 1734839 connectors

ki5libs:
Adding TE 1734839 connectors.
Script based on conn_ffc_te_84952-84953.py


poeschlr:
thanks

PR #450, opened 2019-10-29
Add DFN-4

DanSGiesbrecht:
Used for Sensiron SHTC3

---

image

image

---

From [datasheet](https://www.sensirion.com/fileadmin/user_upload/customers/sensirion/Dokumente/0_Datasheets/Humidity/Sensirion_Humidity_Sensors_SHTC3_Datasheet.pdf):

image

image

### DIFF
--- a/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/dfn.yaml
+++ b/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/dfn.yaml
@@ -1,3 +1,38 @@
+DFN-4-1EP_2x2mm_P1mm_EP0.7x1.6mm:
+  device_type: 'DFN'
+  #manufacturer: 'man'
+  #part_number: 'mpn'
+  size_source: 'https://www.sensirion.com/fileadmin/user_upload/customers/sensirion/Dokumente/0_Datasheets/Humidity/Sensirion_Humidity_Sensors_SHTC3_Datasheet.pdf'
+  # ipc_class: 'qfn' # 'qfn_pull_back'
+  #ipc_density: 'least' #overwrite global value for this device.
+  body_size_x:
+    nominal: 2
+    tolerance: 0
+  body_size_y:
+    nominal: 2
+    tolerance: 0
+  lead_width:
+    nominal: 0.35
+    tolerance: 0
+  lead_len:
+    nominal: 0.35
+    tolerance: 0
+
+  EP_size_x: 0.7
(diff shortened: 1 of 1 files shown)


scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/dfn.yaml
```diff
@@ -333,7 +368,7 @@ DFN-8-1EP_4x4mm_P0.8mm_EP2.3x3.24mm:
 
   EP_size_x_overwrite: 2.30
   EP_size_y_overwrite: 3.24
-
```
DanSGiesbrecht: Save automatically removed whitespace.

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/dfn.yaml
```diff
@@ -696,5 +731,3 @@ TDFN-6-1EP_2.5x2.5mm_P0.65mm_EP1.3x2mm:
   suffix: ''
   #include_suffix_in_3dpath: 'False'
   #include_pad_size: 'fp_name_only' # 'both' | 'none' (default)
-
```
DanSGiesbrecht: Save automatically removed extra lines at end.

poeschlr:
I am not really happy about the available documentation of this part. Is there possibly some other document available that contains tolerance information? Otherwise, we might be better off to use the suggested footprint as is instead of using the ipc generator.

DanSGiesbrecht:
Thanks for the review, @poeschlr. I agree that there is not great documentation for the package. I cannot find any other official source for package dimensions.

So, do you suggest the following steps?
1) close this PR
2) implement the manufacturer-recommended footprint

If you agree with 1) and 2) above:
- I assume that if it's not IPC-compliant, it should not be a generic `DFN-4` part, but rather `Sensirion_DFN-4...`. Is that correct?
- where should this footprint be located? Should I create a `Sensor_Humidity.pretty`?

poeschlr:
Sensor_Humidity sounds like a good library name. (A humidity sensor will most likely have some opening in the package that could be represented in the 3d model. Another reason for a speialized footprint here)

Prefixing with manufacturer name should be ok

I think there is no harm in having rounded rectangle pads using IPC rules.

DanSGiesbrecht:
Thanks! I'll start work on this. It'll give me a good excuse to learn the 3D CAD workflow!

PR #457, opened 2019-11-04
Added Wuerth WE-XHMI.

cnieves1:
Datasheet: https://katalog.we-online.de/pbs/datasheet/74439358068.pdf

imagen

### DIFF
--- a/scripts/Inductor_SMD/Inductor_SMD.yml
+++ b/scripts/Inductor_SMD/Inductor_SMD.yml
@@ -138,3 +138,15 @@ L_TDK_SLF12575:
   size: [12.5, 12.5]
   smd_tht: "smd"
   pins: [["smd", "1", -5.55, 0, 2.6, 3.2, 0.0], ["smd", "2", 5.55, 0, 2.6, 3.2, 0.0]]
+
+L_Wuerth_WE-XHMI:
+  description: "Inductor, Wuerth, WE-XHMI, 8.3mmx8.8mm"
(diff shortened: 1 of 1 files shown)


scripts/Inductor_SMD/Inductor_SMD.yml
```diff
@@ -138,3 +138,15 @@ L_TDK_SLF12575:
   size: [12.5, 12.5]
   smd_tht: "smd"
   pins: [["smd", "1", -5.55, 0, 2.6, 3.2, 0.0], ["smd", "2", 5.55, 0, 2.6, 3.2, 0.0]]
+
+L_Wuerth_WE-XHMI-8080:
+  description: "Inductor, Wuerth, WE-XHMI 8080, 8.3mmx8.8mm"
+  datasheet: "https://katalog.we-online.de/pbs/datasheet/74439358068.pdf"
+  tags: "Inductor Wuerth WE-XHMI 8080"
+  extratexts: [[-49.0, -0.05, "", "F.SilkS", 3.0, 3.0]]
```
aris-kimi, line 146: @cnieves1 do you think something here might cause this empty text?

cnieves1: Ok. I understand now where it is coming from.
The question is: why all other inductors in that file have an empty string as the first text field in extratexts keyword?

aris-kimi: hm.. I am not sure. Could it be a mistake?
@evanshultz may i ask for some input here?

PR #460, opened 2019-11-09
Add SOT-363

chschlue:
WIP

SOT-363_SC-70-6

http://www.ti.com/lit/ml/mpds114c/mpds114c.pdf
https://www.analog.com/media/en/package-pcb-resources/package/pkg_pdf/ltc-legacy-sc-70/SOT_6_05-08-1638.pdf
https://assets.nexperia.com/documents/outline-drawing/SOT363.pdf
https://www.jedec.org/sites/default/files/docs/MO-203C.pdf


cpresser:
I suggest to do the whole SC-70-xx series (The generator now supports missing pins).
All but the 4-pin variants are in the JEDEC document. 
The 4-pin-variant is used for `APE8865U4-12-HF-3`. https://www.tme.eu/Document/ced3461ed31ea70a3c416fb648e0cde7/APE8865-3.pdf

PR #473, opened 2019-12-24
Use numpy to accelerate keepout computation if available

lorem-ipsum:
The current keepout code becomes very slow when using many keepouts (a few seconds with about 2000 keepouts which I got by using addKeepoutRound on ~100 pads.

This first fixes the line endings back to unix style and then adds a faster version of the current code which uses [numpy](https://docs.scipy.org/doc/numpy/). If numpy is not available the existing code is used.

The keepout code is converted to a numpy array once when calling a function. On this array optimised array routines from numpy can be used to check all keepouts at once. This reduce the calculation time down significantly (not tested how much exactly, but no longer really noticable).


evanshultz:
@pointhi @poeschlr 
Do you have any objection to this? Seems beneficial if numpy is available and innocuous to other users.

PR #474, opened 2019-12-27
add Texas Power PAD SO-20 (pick relevant changes from #340)

aewallin:
This adds a Texas PowerPAD SO-20.

Used by e.g. OPA569:
http://www.ti.com/lit/ds/symlink/opa569.pdf

Further details on footprint at:
http://www.ti.com/lit/an/slma004b/slma004b.pdf

#340 had the original work but this PR accidentally contained too much else also (rebase on master etc)
This closes #340


poeschlr:
thanks

PR #476, opened 2019-12-31
Adding QFN-56-1EP_8x8mm_P0.5mm_EP4.3x4.3mm

ki5libs:
Adding QFN-56-1EP_8x8mm_P0.5mm_EP4.3x4.3mm

### DIFF
--- a/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/qfn.yml
+++ b/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/qfn.yml
@@ -159,25 +159,6 @@ QFN-36-1EP_6x6mm_P0.5mm_EP4.1x4.1mm:
   center_pad_v_div: 4
   paste_margin_ratio: -0.1
   thermal_vias: false
-QFN-56-1EP_8x8mm_P0.5mm_EP4.3x4.3mm:
-  description: "56-Lead Plastic Quad Flat - 8x8 mm Body [QFN]; http://ww1.microchip.com/downloads/en/DeviceDoc/00002142A.pdf#page=41"
-  pkg_width: 8.0
-  pkg_height: 8.0
-  pitch: 0.5
-  n_horz_pads: 14
-  n_vert_pads: 14
-  pad_width: 0.70
-  pad_height: 0.28
-  pad_h_distance: 7.68
-  pad_v_distance: 7.68
-  corner_pads: false
-  center_pad: true
-  center_pad_width: 4.3
-  center_pad_height: 4.3
-  center_pad_h_div: 1
-  center_pad_v_div: 1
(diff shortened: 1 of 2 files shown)


evanshultz:
1. Please link to page 40 since that's the package drawing.
2. Give tolerances as shown in the datasheet. For example, `pkg_width: 7.85 .. 8.0 .. 8.15`.
3. Make sure you're entering the values from the table in the document. The pad size values in the script aren't even the nominal values.
4. The biggest issue is that you're using the wrong YAML file. Put the package info into https://github.com/pointhi/kicad-footprint-generator/tree/083b79e039dbef88f4b39ce28bf0a064d957311c/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions (under the `qfn` folder in this case). Then generate the footprint using `kicad-footprint-generator/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/ipc_noLead_generator.py`.

ki5libs:
Hi @evanshultz 

sorry for inconvenient, I really confused the old and the new generators. Hope it's much better now. I'm only not sure about thermal vias. Datasheet tells that it should be 0,3-0,4mm diameter, but for me it looks not the drill size but the whole ring diameter. May be I'm not right, but other case it looks too big.

ki5libs:
Another issue - I've got errors during Travis checking:
https://travis-ci.org/KiCad/kicad-footprints/builds/632431076?utm_source=github_status&utm_medium=notification
And I have no idea how to fix it.

evanshultz:
Thanks!

1. Note 2 on page 40 indicates that lead width min should be 0.15mm. So why does that not match the table. It's confusing to me. Do you know how to reconcile that?
1. 0.2mm will be fine. That's the common size in this library. We aren't strictly following the datasheet footprint because we're using our script.
1. Those errors are, unfortunately, normal. Our check script hasn't learned the newer way the script builds footprints with thermal vias.

Also note that I asked for an updated screenshot in the footprint PR.

ki5libs:
I've updated screenshots. But unfortunately I couldn't say anything about Note 2. I didn't understand what it tries to tell so I just took table values. By the way dimension X on page 41 (pad width) is 0.28mm whereas generated value is 0.25 mm.

evanshultz:
Thanks for the screenshots.

Let's adjust the terminal width min to 0.15mm and add a note in the YAML file stating the table and note have different minimum values. Then re-generate the footprint and we may be good to merge.

Our script generates footprints according to IPC 7351 based on the package drawing. It does not know or care about the recommended footprint.

ki5libs:
Done. But looks like it has no influence to the result of generation

evanshultz:
I'm not surprised there is no change, but glad you checked it. Thanks!

PR #477, opened 2020-01-02
Adding ST_uTFBGA-36_3.6x3.6mm_Layout6x6_P0.5mm

ki5libs:
Adding ST_uTFBGA-36_3.6x3.6mm_Layout6x6_P0.5mm


evanshultz:
The link given with a 'b' suffix doesn't work for me, but https://www.st.com/resource/en/datasheet/stulpi01a.pdf does. Can you check that?

Also, the ball size is given as 0.3mm but no footprint is shown. We can then use a 0.25mm pad, according to IPC, so please adjust your pad size.

PR #479, opened 2020-01-07
Add Texas_X2SON-4-1EP_1.1x1.4mm_P0.5mm_EP0.8x0.6mm

rnd2:
Footprint definition for Texas package DMR0004A.
Uses custom pad number scheme using ccw_dual generator.


chschlue:
Pending #422

rnd2:
I have rebased the branch based on the master to avoid any conflicts.

rnd2:
I just noticed that the silkscreen on the left is located next to pin 2 (first image).
And if pin 1 is at top right corner (second image) the silkscreen stays the same.

Based on https://kicad-pcb.org/libraries/klc/F5.1/ the Pin-1 designator is provided on the `F.SilkS` layer.

So, there should be another merge request that handles the silkscreen before this on is merged.

DMR0004A_silk
DMR0004A_silk2

evanshultz:
Pin 1 marked by the absence of silk. This is in accordance with IPC standards.

rnd2:
> Pin 1 marked by the absence of silk. This is in accordance with IPC standards.

Yes, that is correct for the first case. But on the second one, the silkscreen should have been mirrored.

rnd2:
Pending #508

PR #483, opened 2020-01-12
Add JST XAD Vertical (top entry)

fenugrec:
Adds "JST_XAD_B**B-XADSS-N-A_2x**_P2.50mm_Vertical" connectors, two variants, with and without a plastic localizing boss (NPTH PCB hole).
I based this on `conn_jst_PHD_vertical.py` in general and `conn_jst_ze_tht_top.py` for the boss-variant handling.


fenugrec:
Rebased and remove an invalid variant (no 38-pin version). Also created matching PR in kicad-footprints, https://github.com/KiCad/kicad-footprints/pull/2171

PR #487, opened 2020-01-15
Cleanup organization of script directory

poeschlr:
It was getting quite hard to find a script so i consolidated a few directories to reduce the number of entries within the script directory.
I left out the remaining connector scripts as there is still a pull request open for at least one of them.


poeschlr:
@pointhi, @evanshultz any input? (Such a reorg should really happen rather quickly to avoid unnecessary merge conflicts)

scripts/Inductors/Inductor_generic/Readme.md
```diff
@@ -0,0 +1 @@
+Generic generator for inductor footprints. Use to duplicate the suggested footprint given in the datasheet. Used to generate footprints for inductors with rectangular base and possibly some corners chamfered, rounded or cut out. (Example TDK SLF6028)
```
evanshultz: Can you look at the second sentence? Maybe it's me, but I'm not making sense of it.

evanshultz:
Good idea!

A few thoughts:
- Could `/PadGenerator` go under `/general` or `/tool`?
- Can both `/Connector*` and `/Multicomp` and `/pin-header*` folders go under the root `/Connector`? Maybe one of these is what you alluded to above?
- I think the following go under `/Converter_DCDC`: `/Recom_DCDC` and `/Vigortronix`.
- Would you like to consider a 'Passive' or 'Device' folder to collect resistor and potentiometer and inductor and maybe LED and oscillator and other folders? This would impact a number of open PRs so it's just a conversation starter.

So this PR doesn't affect any open PRs? You mentioned a connector PR affecting at least one possible reorganization you didn't make, but no open PRs will be affected right now if this is merged?

pointhi:
@poeschlr @evanshultz I'm ok with any organization you want to have of the scripts folder.

poeschlr:
This does not affect any footprint as far as i am aware, it is however rather hard to check this easily (github does not offer a search for affected files)

And this does not need to be the last reorg. I currently collected the things that i think are the most annoying.

scripts/Inductors/Inductor_generic/Readme.md
```diff
@@ -0,0 +1,3 @@
+Generic generator for inductor footprints. Use it to exactly duplicate the suggested footprint given in the datasheet (as oposed to the IPC generators which do not rely on the suggested footprint).
```
evanshultz: 'opposed' is missing a 'p'.

scripts/Inductors/Inductor_generic/Readme.md
```diff
@@ -0,0 +1,3 @@
+Generic generator for inductor footprints. Use it to exactly duplicate the suggested footprint given in the datasheet (as oposed to the IPC generators which do not rely on the suggested footprint). 
+
+It supports generating footprints for inductors with rectangular base and possibly some corners chamfered, rounded or cut out. (Example TDK SLF6028)
```
evanshultz: This sentence is rather awkward. Could change `...`with rectangular base and possibly some corners...` to `...with a rectangular base and corners which are...` or something else you prefer.

evanshultz:
@poeschlr 
Can you respond to the first three bullets I mentioned above? Those seem straightforward enough to include now.

poeschlr:
I did not move the pin header stuff because you have a pull request open there and i do not want to create extra work for you. (Might be easier if you move it in your pull request)

The pad generator is a full script so it does not fit under tools. It does not really fit in general because that folder currently contains only the obsolete smd resisotr/capacitor/... generator.

I am also not sure what the difference is between the Connector_PinSocket and pin-headers_socket-strip generators (i would assume one of them is obsolete)

The Multicom generators are now moved but i fear all of them are obsolete as they use the old discontinued kicad_mod library instead of the current KicadModTree.

Creating a Passsives folder might be something for a future reorg.

scripts/Inductors/Inductor_generic/Readme.md
poeschlr: Fixed but had a better idea than your suggestion

scripts/Inductors/Inductor_generic/Readme.md
poeschlr: Fixed

evanshultz:
I can move the connector stuff in my PR. No problem.

Connector_PinSocket just does sockets, while pin-headers_socket-strip does those plus many other types. I'm not sure which footprints are in the library now, but I find the latter to be perfectly serviceable and much, MUCH easier to maintain.

Can you look at the third bullet above? I didn't see a reply to it.

poeschlr:
I answered your bullet points above already. Well kind of i overlooked the Converter_DCDC stuff but will move those two as well.

antoniovazquezblanco:
It is nice to see this reorganization. I think it helps clarify the structure of the repo.

All points made by @evanshultz seem to have been addressed but if anything is missing I am willing to help fordward in case @poeschlr is busy.

poeschlr:
I think it is just a case of evan having too little time to merge. I fear right now only evan, thomas and I have write access and i generally don't like to merge my own work.

evanshultz:
Oops! My mistake. I didn't see a commit, or I forgot to respond if I did.

The Converter_DCDC folder has several files and two subfolders, each with just a single script. Can those folders be removed and the scripts moved up a level?

Also, I don't believe I made any changes to the header scripts in my PR for IDC footprints that you merged. It looks the same now on MASTER as what I described above. Would you care to organize that a bit better also? I made a suggestion above.

PR #491, opened 2020-01-22
Modify SMD_chip_package_rlc-etc.py to allow manufacturer PN naming.

cnieves1:
I needed to generate a footprint for a ceramic chip antenna, like this one:
  https://www.johansontechnology.com/datasheets/2450AT18A100/2450AT18A100.pdf

I thought it would be great to use SMD_chip_package_rlc-etc.py for this, but it doesn't allow to modify the naming to a manufacturer_part-number scheme.

This patch allows to add 3 more data fields: manufacturer, manufacturer_pn and description.
Using this, I can generate footprints for the above antenna, with correct naming.

### DIFF
--- a/scripts/SMD_chip_package_rlc-etc/SMD_chip_package_rlc-etc.py
+++ b/scripts/SMD_chip_package_rlc-etc/SMD_chip_package_rlc-etc.py
@@ -151,31 +151,40 @@ class TwoTerminalSMDchip():
         code_metric = device_size_data.get('code_metric')
         code_letter = device_size_data.get('code_letter')
         code_imperial = device_size_data.get('code_imperial')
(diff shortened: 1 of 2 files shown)


poeschlr:
Looks good to me, thanks

cpresser:
I am currently working on https://github.com/pointhi/kicad-footprint-generator/pull/438, thus this will need re-basing soon.
The good news is, since I am already working on that generator I can also review and merge this PR.

evanshultz:
For consistency, I'd rather implement a `custom_name_format` keyword for YAML files like in our other generators. At least the SON one. That keyword and it's behavior, as implemented elsewhere, gives more flexibility than a special custom name format that is fine-tuned just for manufacturer and MPN.

PR #255, opened 2018-12-28
add definition for LFCSP-8 3x3mm from Analog

lorem-ipsum:
Add a yaml definition for a LFCSP chip from Analog Devices to the DFN_QFN package generator. [Package documentation](https://www.analog.com/media/en/package-pcb-resources/package/pkg_pdf/lfcspcp/cp-8/CP_8_13.pdf)

PR #493, opened 2020-01-28
Move orphaned LFCSP definition (#255)

chschlue:


evanshultz:
This is using the old dimension format but since it's a move and not truly new I'll merge it.

PR #494, opened 2020-01-30
Hvqfn 24

jneiva08:
Created footprint for HVQFN 24

Datasheet: https://www.nxp.com/docs/en/package-information/SOT616-3.pdf
and https://www.nxp.com/docs/en/data-sheet/NHS3100.pdf


jneiva08:
When try to generate the footprint with the last version  shows the following error:

`$ ./ipc_noLead_generator.py size_definitions/qfn/hvqfn.yaml Building footprint for parameter set: HVQFN-24-1EP_4x4mm_P0.5mm_EP2.5x2.5mm
The use of manual heel reduction is deprecated. It is automatically calculated from the minimum EP to pad clearance (ipc config file)`

what can do to fix-it?

chschlue:
Remove the `heel_reduction` parameter.

jneiva08:
thanks for the help, it worked

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/qfn/hvqfn.yaml
```diff
@@ -0,0 +1,53 @@
+HVQFN-24-1EP_4x4mm_P0.5mm_EP2.5x2.5mm:
```
chschlue: Should be EP2.6x2.6mm

chschlue:
FP merged.
Size definition still needs to be corrected.

jneiva08:
@chschlue , Only needs to change EP size on name? and not the actual size?
It's recommended to make new request to update in footprint?

chschlue:
EP size is already 2.6x2.6, just the name of the definition here is wrong.

jneiva08:
tks

PR #495, opened 2020-02-03
[Fix] Platform-agnostic path

chschlue:


evanshultz:
Thanks!

Fix for https://github.com/KiCad/kicad-footprints/pull/2079.

PR #470, opened 2019-12-17
Add more Maxim 5x5mm TQFN footprints

matthijskooijman:
This adds all Txx55 footprints from the Maxim 21-0140 datasheet that were not already added. In addition, this addds thermal vias to two existing TQFN footprints that did not have them yet.

I have some remarks about these footprints and the way they are generated. To keep discussion in one place, I've noted them in https://github.com/KiCad/kicad-footprints/pull/2008

PR #496, opened 2020-02-05
Fix V/H typo in ipc_noLead_generator.py

matthijskooijman:
For footprints with a non-standard lead center the horizontal lead
length was used in place of the vertical one. Since there are no
footprints that have both a non-standard center and different V/H lead
lengths, this did not actually affect any current footprints.


evanshultz:
@matthijskooijman 
How did you come across this? I didn't take a close look yet, but I'd like to understand how you found this for perspective. Thank you.

matthijskooijman:
I was trying to understand the generator a bit for #470 and noticed this bit of code looked incorrect :-)

As I mentioned, there are not generators actually using this (i.e. different `lead_len_H` and `lead_len_V`) currently AFAICS, so I did not really test this change, but seemed like just a copy-paste typo to me.

evanshultz:
@matthijskooijman 
Ah. Got it. Thank you.

@poeschlr 
This is your code and it looks like a good fix. Can you take a look?

PR #504, opened 2020-02-27
Capacitor_THT: Update and regenerate models

antoniovazquezblanco:
Update scripts to allow to generate capacitors withouth splitting the lib. An atempt on continuing https://github.com/KiCad/kicad-footprints/pull/1715

### DIFF
--- a/scripts/Capacitors_THT/make_Capacitors_THT.py
+++ b/scripts/Capacitors_THT/make_Capacitors_THT.py
@@ -181,7 +181,7 @@ if __name__ == '__main__':
         makeResistorRadial(seriesname=seriesname, rm=rm, w=w, h=d, ddrill=ddrill, R_POW=R_POW,
                                     type=type, w2=w2, x_3d=[0, 0, 0], s_3d=[1, 1, 1], has3d=1,
                                     specialfpname="", add_description=add_description, name_additions=name_additions,
-                                    specialtags=specialtags, classname="C", lib_name="${KISYS3DMOD}/Capacitor_THT_Rectangular", script3d=scr, height3d=h3d)
+                                    specialtags=specialtags, classname="C", lib_name="Capacitor_THT", script3d=scr, height3d=h3d)
 
 
     ###########################################################
@@ -208,7 +208,7 @@ if __name__ == '__main__':
         makeResistorRadial(seriesname=seriesname, rm=rm, rm2=rm2, w=w, h=d, ddrill=ddrill, R_POW=R_POW,
                                     type=type, w2=w2, x_3d=[0, 0, 0], s_3d=[1, 1, 1], has3d=1,
                                     specialfpname="", add_description=add_description, name_additions=name_additions,
-                                    specialtags=specialtags, classname="C", lib_name="${KISYS3DMOD}/Capacitor_THT_Rectangular", script3d=scr, height3d=10)
+                                    specialtags=specialtags, classname="C", lib_name="Capacitor_THT", script3d=scr, height3d=10)
 
 
 
@@ -276,7 +276,7 @@ if __name__ == '__main__':
         makeResistorRadial(seriesname=seriesname, rm=rm, w=w, h=d, ddrill=ddrill, R_POW=R_POW,
                            type=type, w2=w2, x_3d=[0, 0, 0], s_3d=[1, 1, 1], has3d=1,
                            specialfpname="", add_description=add_description, name_additions=name_additions,
-                           specialtags=specialtags, classname="C", lib_name="${KISYS3DMOD}/Capacitor_THT", script3d=script3mkt)
+                           specialtags=specialtags, classname="C", lib_name="Capacitor_THT", script3d=script3mkt)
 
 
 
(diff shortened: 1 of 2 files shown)


scripts/tools/footprint_scripts_resistorlike.py
```diff
@@ -280,16 +280,18 @@ def makeResistorAxialHorizontal(seriesname, rm, rmdisp, w, d, ddrill, R_POW, typ
...
     # print(kicad_mod.getCompleteRenderTree())
 
     # write file
     file_handler = KicadFileHandler(kicad_mod)
-    file_handler.writeFile(footprint_name+'.kicad_mod')
-
+    directory = os.path.dirname(lib_name + ".pretty/")
+    if not os.path.exists(directory):
+        os.makedirs(directory)
+    file_handler.writeFile(lib_name + ".pretty/" + footprint_name + '.kicad_mod')
```
antoniovazquezblanco, line 294: Simply add a check for the output directory before actually writing the footprint. This avoids errors if the directory does not exist or if library name is changed (for example in future library split...)

scripts/tools/footprint_scripts_resistorlike.py
```diff
@@ -1025,12 +1027,15 @@ def makeResistorRadial(seriesname, rm, w, h, ddrill, R_POW, innerw=0,innerh=0,rm
...
     # print(kicad_mod.getRenderTree())
     # print(kicad_mod.getCompleteRenderTree())
 
     # write file
     file_handler = KicadFileHandler(kicad_mod)
-    file_handler.writeFile(footprint_name + '.kicad_mod')
+    directory = os.path.dirname(lib_name + ".pretty/")
+    if not os.path.exists(directory):
+        os.makedirs(directory)
+    file_handler.writeFile(lib_name + ".pretty/" + footprint_name + '.kicad_mod')
```
antoniovazquezblanco, line 1041: Simply add a check for the output directory before actually writing the footprint. This avoids errors if the directory does not exist or if library name is changed (for example in future library split...)

scripts/tools/footprint_scripts_resistorlike.py
```diff
@@ -280,16 +280,18 @@ def makeResistorAxialHorizontal(seriesname, rm, rmdisp, w, d, ddrill, R_POW, typ
 
     # add model
     if (has3d!=0):
-        kicad_mod.append(Model(filename=lib_name + ".3dshapes/"+footprint_name+".wrl", at=x_3d, scale=s_3d, rotate=[0, 0, 0]))
+        kicad_mod.append(Model(filename="${KISYS3DMOD}/" + lib_name + ".3dshapes/"+footprint_name+".wrl", at=x_3d, scale=s_3d, rotate=[0, 0, 0]))
```
antoniovazquezblanco, line 283: Prepend 3D path environment variable to the generated 3D model path as the library name now does not contain such path.

scripts/tools/footprint_scripts_resistorlike.py
```diff
@@ -1025,12 +1027,15 @@ def makeResistorRadial(seriesname, rm, w, h, ddrill, R_POW, innerw=0,innerh=0,rm
 
     # add model
     if (has3d != 0):
-        kicad_modg.append(Model(filename=lib_name + ".3dshapes/" + footprint_name + ".wrl", at=x_3d, scale=s_3d, rotate=[0, 0, 0]))
+        kicad_modg.append(Model(filename="${KISYS3DMOD}/" + lib_name + ".3dshapes/" + footprint_name + ".wrl", at=x_3d, scale=s_3d, rotate=[0, 0, 0]))
```
antoniovazquezblanco, line 1030: Prepend 3D path environment variable to the generated 3D model path as the library name now does not contain such path.

scripts/Capacitors_THT/make_Capacitors_THT.py
```diff
@@ -181,7 +181,7 @@
         makeResistorRadial(seriesname=seriesname, rm=rm, w=w, h=d, ddrill=ddrill, R_POW=R_POW,
                                     type=type, w2=w2, x_3d=[0, 0, 0], s_3d=[1, 1, 1], has3d=1,
                                     specialfpname="", add_description=add_description, name_additions=name_additions,
-                                    specialtags=specialtags, classname="C", lib_name="${KISYS3DMOD}/Capacitor_THT_Rectangular", script3d=scr, height3d=h3d)
+                                    specialtags=specialtags, classname="C", lib_name="Capacitor_THT", script3d=scr, height3d=h3d)
```
antoniovazquezblanco, line 184: Avoid prepending "${KISYS3DMOD}" in the library name as it is not part of it. The new generator script will take care of adding this to the 3D model path.

Change library name parameter to avoid library splitting.

antoniovazquezblanco:
It has been a while since I last had a look at this. To make things easier I have added comments to code to make it easier for the reviewer. Summary of the changes:

- Avoid using `${KISYS3DMOD}` in library name in the top level script. Underlying generator script will add this to corresponding places instead.
- Avoid capacitor splitting by libraries yet (Someone had introduced this into the scripts but the models were not updated and changes have been refused so I fixed this situation by reversing those changes). It will probably be done in the future but planning is required.
- Check that the output folders exist before writing the footprint files to avoid errors.

PR #507, opened 2020-03-08
Move IPC helpers from tools to utils

poeschlr:
Move IPC helpers from tools directory into the utils directory of the KicadModTree and Package scripts.

This is in preparation to make footprint wizards.


poeschlr:
I will look into the style issues as soon as i find time for it.

evanshultz:
Looks OK to me once the style issues are addressed.

evanshultz:
Everything here is under /scripts, so @pointhi should be fine with it. Can codeclimate be run again to check if there are any remaining style issues, or has it already run and approved things but I don't know where to look to see that?

poeschlr:
Code climate just updates its comment and travis includes formatting checks as well.

PR #371, opened 2019-06-07
Allow custom numbering scheme in PadArray and QFN scripts

rnd2:
Custom pin numbers for the pad array can be passed to PadArray
as a list through the keyword argument `pad_numbers`.
This bypasses the increment functionality.

Additonally, add the ability to define custom pin numbers per side on a QFN
package, as some packages use a non-standard numbering scheme.

This is needed for KiCad/kicad-footprints#1553 and it seems that there is also related issues (#323, #280) 

As an example, here is the definition from the pull request mentioned above using the changes I suggest. Notice the `EP_pin_number` for the exposed pad and `pad_numbers_{top,right,left,bottom}`

```yaml
Texas_X2SON-4-1EP_1.1x1.4mm_P0.5mm_EP0.8x0.6mm:
  device_type: 'X2SON'
  library: Package_SON
  manufacturer: 'Texas'
  #part_number: 'mpn'
  size_source: 'http://www.ti.com/lit/ds/symlink/drv5032.pdf#page=28'
  ipc_class: 'qfn' #'qfn' | 'qfn_pull_back'
  #ipc_density: 'least' #overwrite global value for this device. (if used, pad values are closer to TI's suggested fp)

  body_size_x:
    minimum: 1.05
    maximum: 1.15
  body_size_y:
    minimum: 1.35
    maximum: 1.45
  overall_height:
    maximum: 0.4

  lead_width:
    minimum: 0.17
    maximum: 0.27
  lead_len:
    minimum: 0.15
    maximum: 0.25

  EP_size_x:
    nominal: 0.8
    tolerance: 0.05
  EP_size_y:
    nominal: 0.6
    tolerance: 0.05
  # EP_paste_coverage: 0.65
  EP_num_paste_pads: 1 #[2, 2]
  EP_pin_number: 5

  pitch: 0.5
  num_pins_x: 2
  num_pins_y: 0
  #include_suffix_in_3dpath: 'False'
  #include_pad_size: 'fp_name_only' # 'both' | 'none' (default)

  pad_numbers_top: [1, 4] # left to right
  #pad_numbers_right: [5, 6] # top to bottom
  pad_numbers_bottom: [2, 3] # left to right
  #pad_numbers_left: [8, 7] # top to bottom

  thermal_vias:
    count: 1
    drill: 0.2
    # min_annular_ring: 0.15
    paste_via_clearance: 0.1
    EP_num_paste_pads: 1 # [2, 2]
    EP_paste_coverage: 0.75
    grid: [1, 1]
[...]

PR #508, opened 2020-03-14
Add ipc noLead silkscreen calculator

rnd2:
When the custom pin number generator is used pin 1
is in a different location than the expected.

Pin 1 can be in one of the four corners:
  top left/right and bottom left/right

Add calc_silk_(x,y) to calculate the silkscreen line coords.

### DIFF
--- a/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/ipc_noLead_generator.py
+++ b/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/ipc_noLead_generator.py
@@ -16,6 +16,7 @@ sys.path.append(os.path.join(sys.path[0], "..", "..", "tools"))  # load parent p
 from footprint_text_fields import addTextFields
 from ipc_pad_size_calculators import *
 from quad_dual_pad_border import add_dual_or_quad_pad_border
+from silkscreen_calc import calc_silk_x, calc_silk_y
 
 sys.path.append(os.path.join(sys.path[0], "..", "utils"))
 from ep_handling_utils import getEpRoundRadiusParams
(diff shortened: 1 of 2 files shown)


rnd2:
This request is related to #507 and #479.

If #507 is going to be merged first the this PR needs to be rebased to relocate the file `silkscreen_calc.py` to `util`.

When this PR is merged #479 can be merged (after a rebase) too.

rnd2:
A few examples to demonstrate:

For `num_pins_x = 0`,

left: `cw_dual` generator
right: `ccw_dual` generator, `start = 'bottom_right'`

  

For `num_pins_y = 0`,

left: `cw_dual` generator, `start = 'bottom_left'`
right: `ccw_dual` generator, `start = 'top_right'`

cpresser:
Question: in the F.Fab layer the Pin1 marker does not match the silkscreen marker. Why?

evanshultz:
This is what is shown for IPC 7351 and built by the generator. See https://ocipcdc.org/archive/What_is_New_in_IPC-7351C_03_11_2015.pdf#page=19 for an example of the silk and the fab is the historical way IPC docs have shown pin 1, though I don't see an example if that doc.

rnd2:
> Question: in the F.Fab layer the Pin1 marker does not match the silkscreen marker. Why?

You are right. The fab layer must be changed too.

Should I push more commits here to address this or make a new PR just for (F/B).Fab layer?

evanshultz:
Sorry for my stupid comment earlier. I didn't understand at that time all that was going on. I vote to push more commits here, otherwise MASTER will be able to generate footprints that have known issues. A squash merge can be used to keep history clean so there's no issue.

rnd2:
Ready for merging,

top_lefttop_right

bottom_leftbottom_right

cpresser:
Another question: KLC F4.2 says Pin1 should be top left (https://kicad-pcb.org/libraries/klc/F4.2/).
All the above examples could be rotated to full fill that requirement.
Is that something that should be addressed as well?

rnd2:
The defaults are set so that the resulting footprint is conforming with KLC.
The rest is just for completeness.

rnd2:
For some background info.

This all started so that the generator could generate the footprint for DMR0004A (http://www.ti.com/lit/ml/mpss084/mpss084.pdf). 

For an unknown reason Texas decided to use non-standard numbering. I had a look in the datasheets for all SON packages from TI and this is the only one that uses counter-clockwise numbering.

evanshultz:
@cpresser is right that the footprints don't conform to KLC since pin 1 is not in the top left (orientation A). Are you willing and able to update this PR to address that? While I haven't tried using this part of the framework myself, I believe https://github.com/pointhi/kicad-footprint-generator/blob/master/KicadModTree/nodes/specialized/Rotation.py may make this easy.

We can merge this anyway and Rene can clean up https://github.com/pointhi/kicad-footprint-generator/pull/507 later.

evanshultz:
@cpresser 
What do you want to do with this? I quickly looked at just the changed file (not the new one) and I'm not thrilled with some variable names but it is a needed fix. The custom pad generators are in the script right now but they don't work as expected without this code (and then the further footprint rotation fix for pin 1 location). Thoughts?

cpresser:
I looked at DMR0004A (http://www.ti.com/lit/ml/mpss084/mpss084.pdf) to understand the problem.
Disclaimer: **I have not yet looked at this problem in Detail.**

My first idea how to solve this would be to patch the pin-number-generator. Or introduce a new one. How about adding a property
```yaml
pin_numbers: [1, 4, 3, 2]
```
The generator iterates that list and just yields one number after another. If we ensure that the list always starts with 1, no changes to component orientation, Silkscreen and F.Fab are required. That will reduce the amount of code needed significantly.

rnd2:
Hi @cpresser , did you have a look at the first PR for this (#371). Initially I had the following properties:

```yaml
pad_numbers_top: [1, 4] # left to right
#pad_numbers_right: [5, 6] # top to bottom
pad_numbers_bottom: [2, 3] # left to right
#pad_numbers_left: [8, 7] # top to bottom
```
Later on that PR there was a suggestion to use a generator instead, which was then merged.

It won't be too difficult to add a generator that yields the items of the property.

cpresser:
> Hi @cpresser , did you have a look at the first PR for this (#371). Initially I had the following properties:
> 

I was not aware of that. Thanks for the hint.

> 
> It won't be too difficult to add a generator that yields the items of the property.

Sounds good to me. But that seems to simple given. Is there another angle to the original problem not solved by this? It feels like I am missing something.

evanshultz:
I don't think pin 1 has to be the first pin as the entire footprint can be rotated as I mentioned above. I haven't tried this, but it looks from the documentation that it would work.

PR #514, opened 2020-03-20
Added TDFN-8 2x2mm footprint with EP0.8x1.38mm for MAX17620.

matyas1995:
Hello,

I have created a matching TDFN-8 footprint for the MAX17620 step-down regulator, as discussed in the PR https://github.com/KiCad/kicad-symbols/pull/2559 in the kicad-symbols repo.

Specification can be found here: http://pdfserv.maximintegrated.com/land_patterns/90-0065.PDF

grafik

A PR for this footprint was also created in the official kicad-footprints library: https://github.com/KiCad/kicad-footprints/pull/2159

### DIFF
--- a/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/dfn.yaml
+++ b/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/dfn.yaml
@@ -688,7 +688,7 @@ TDFN-6-1EP_2.5x2.5mm_P0.65mm_EP1.3x2mm:
   #include_suffix_in_3dpath: 'False'
   #include_pad_size: 'fp_name_only' # 'both' | 'none' (default)
 
-TDFN-8-1EP_2x2mm_P0.5mm_EP0.8x1.20mm:
+TDFN-8-1EP_2x2mm_P0.5mm_EP0.8x1.30mm:
   device_type: 'TDFN'
   size_source: 'https://pdfserv.maximintegrated.com/package_dwgs/21-0168.PDF'
   body_size_x: 2.0 +/-0.1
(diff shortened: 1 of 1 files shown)


chmorgan:
Hi @myfreescalewebpage it looks like the dimensions in the commit match the pads linked by the url. I haven't reviewed a change like this before however so I'll defer to someone else to do a final check.

chmorgan:
@chschlue done.

matyas1995:
Thanks for the feedback, I learn something new every day :)

I have changed the footprint generator script as requested by @chschlue. I also changed the package name to TDFN-8-1EP_2x2mm_P0.5mm_EP0.8x1.2mm as this is the true size the package drawing for the MAX17620 dictates.

chschlue:
This is OK now, just one more comment for completeness:
It would have been fine to enlarge the EP to 1.3mm (the maximum package pad size Maxim specifies) by using the EP_size_y_overwrite parameter to get closer to their recommended land size or, alternatively, to create a solder mask defined footprint with 0.8x1.38mm copper (probably not very useful as we're talking about 0.4mm mask cover per side in this case).

matyas1995:
Ok, thanks for the help, I have modified the package now. Just a question for a learning effect:
What is the EP_size_y_overwrite for and what is the difference to EP_size_y?

Greetings,
Matyas

> This is OK now, just one more comment for completeness:
> It would have been fine to enlarge the EP to 1.3mm (the maximum package pad size Maxim specifies) by using the EP_size_y_overwrite parameter to get closer to their recommended land size or, alternatively, to create a solder mask defined footprint with 0.8x1.38mm copper (probably not very useful as we're talking about 0.4mm mask cover per side in this case).

chschlue:
> What is the EP_size_y_overwrite for and what is the difference to EP_size_y?

EP_size_[xy]_overwrite is exactly for cases like this.
It allows us to tell the actual package dimensions apart from intentional deviations.
The output is the same, being explicit about a certain EP land size just keeps the definitions cleaner.

chschlue:
FP merged

PR #516, opened 2020-03-23
Add vertital  JST ZH THT connectors

Teslafly:
Adds a generator for vertical through hole jst zh series connectors.
Based on the jst ph vertical generator, but better parameterized/commented.


Teslafly:
image

PR #518, opened 2020-03-30
Added UQFN-20_3x3mm_P0.4mm without exposed pad

cnieves1:
Datasheet: https://resurgentsemi.com/wp-content/uploads/2018/09/MPR121_rev5-Resurgent.pdf?d453f8&d453f8
Captura de pantalla de 2020-03-30 14-40-55
Captura de pantalla de 2020-03-30 14-41-08

### DIFF
--- a/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/qfn/uqfn.yaml
+++ b/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/qfn/uqfn.yaml
@@ -123,6 +123,66 @@ UQFN-16-1EP_4x4mm_P0.65mm_EP2.6x2.6mm:
   #suffix: '_Pad{pad_x:.2f}x{pad_y:.2f}mm_HandSolder'
   #include_suffix_in_3dpath: 'False'
 
+UQFN-20_3x3mm_P0.4mm: # No exposed pad
+  device_type: 'UQFN'
+  #manufacturer: 'man'
+  #part_number: 'mpn'
+  size_source: 'https://resurgentsemi.com/wp-content/uploads/2018/09/MPR121_rev5-Resurgent.pdf?d453f8&d453f8'
+  ipc_class: 'qfn' # 'qfn_pull_back'
+  #ipc_density: 'least' #overwrite global value for this device.
+  # custom_name_format:
+  body_size_x:
+    nominal: 3
+  body_size_y:
+    nominal: 3
+  overall_height:
+    minimum: 0.5
+    nominal: 0.55
+    maximum: 0.60
+
+  lead_width:
+    minimum: 0.15
+    nominal: 0.20
+    maximum: 0.25
+  lead_len:
+    minimum: 0.35 # Would be 0.4mm, but set to 0.35 to make this compatible with 'http://ww1.microchip.com/downloads/en/PackagingSpec/00000049BQ.pdf#page=561'
+    nominal: 0.5
+    maximum: 0.6
+
+#  EP_size_x:
+#    minimum: 1.65
+#    nominal: 1.75
+#    maximum: 1.85
+#  EP_size_x_overwrite: 1.85
+#  EP_size_y:
+#    minimum: 1.65
+#    nominal: 1.75
+#    maximum: 1.85
+#  EP_size_y_overwrite: 1.85
+#  EP_num_paste_pads: [2, 2]
+
+#  thermal_vias:
+#    count: [2, 2]
+#    drill: 0.2
+#    # min_annular_ring: 0.15
+#    paste_via_clearance: 0.1
+#    EP_num_paste_pads: [2, 2]
+#    #paste_between_vias: 1
+#    #paste_rings_outside: 1
+#    EP_paste_coverage: 0.75
+#    #grid: [1, 1]
+#    # bottom_pad_size:
+#    paste_avoid_via: False
+
+  pitch: 0.4
+  num_pins_x: 5
+  num_pins_y: 5
+  chamfer_edge_pins: 0.12
+  #pin_count_grid:
+  #pad_length_addition: 0.5
+  #suffix: '_Pad{pad_x:.2f}x{pad_y:.2f}mm_HandSolder'
+  #include_suffix_in_3dpath: 'False'
+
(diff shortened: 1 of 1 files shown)


cnieves1:
Similar package at: http://ww1.microchip.com/downloads/en/PackagingSpec/00000049BQ.pdf#page=561
Captura de pantalla de 2020-03-30 14-43-06

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/qfn/uqfn.yaml
```diff
@@ -123,6 +123,66 @@ UQFN-16-1EP_4x4mm_P0.65mm_EP2.6x2.6mm:
...
+  size_source: 'https://resurgentsemi.com/wp-content/uploads/2018/09/MPR121_rev5-Resurgent.pdf?d453f8&d453f8'
+  ipc_class: 'qfn' # 'qfn_pull_back'
+  #ipc_density: 'least' #overwrite global value for this device.
+  # custom_name_format:
+  body_size_x:
+    nominal: 3
+  body_size_y:
+    nominal: 3
+  overall_height:
+    minimum: 0.5
```
evanshultz: Should be 0.51.

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/qfn/uqfn.yaml
```diff
@@ -123,6 +123,66 @@ UQFN-16-1EP_4x4mm_P0.65mm_EP2.6x2.6mm:
...
+    nominal: 0.55
```
evanshultz: No nominal given.

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/qfn/uqfn.yaml
```diff
@@ -123,6 +123,66 @@ UQFN-16-1EP_4x4mm_P0.65mm_EP2.6x2.6mm:
...
+    maximum: 0.60
+
+  lead_width:
+    minimum: 0.15
+    nominal: 0.20
```
evanshultz: No nominal given.

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/qfn/uqfn.yaml
```diff
@@ -123,6 +123,66 @@ UQFN-16-1EP_4x4mm_P0.65mm_EP2.6x2.6mm:
...
+    maximum: 0.25
+  lead_len:
+    minimum: 0.35 # Would be 0.4mm, but set to 0.35 to make this compatible with 'http://ww1.microchip.com/downloads/en/PackagingSpec/00000049BQ.pdf#page=561'
+    nominal: 0.5
```
evanshultz: No nominal given.

scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/qfn/uqfn.yaml
```diff
@@ -123,6 +123,66 @@ UQFN-16-1EP_4x4mm_P0.65mm_EP2.6x2.6mm:
...
+    maximum: 0.6
+
+#  EP_size_x:
```
evanshultz: Remove the unneeded lines instead of commenting them out.

cnieves1:
@evanshultz 
I tried to make this compatible with both Resurgent Semiconductor and Microchip's package, as they are very similar. Microchip's package specifies nominal values (page 563 of the above link):
Captura de pantalla de 2020-03-30 21-49-04

Please advice with the following:

-overall height: 0.51-0.60 (Resurgent), 0.50-0.60 (Microchip). I put 0.50-0.60 in the script, since it covers both. Microchip _does_ specify nominal dimension as 0.55.
- lead width: 0.15-0.25, with 0.2 as nominal (Resurgent). Resurgent specifies nominal in detail G. 0.20-0.30, with 0.25 as nominal (Microchip). In this case I put Resurgent's values. What should I put here? 0.15-0.30 with no nominal value?
 -lead length: 0.4-0.6 (Resurgent), 0.35-0.45 with 0.4 as nominal (Microchip). I put 0.35-0.6, with 0.5 nominal, as it seems the worst case. Would this be ok? should I remove nominal value?

evanshultz:
@cnieves1 
This is a bit tricky, and we've discussed this before but not had to cross this bridge with a generic footprint.

One option would be two footprints that are manufacturer-specific.

@poeschlr 
Several times it's come up where packages are totally compatible according to our naming convention because lead length, lead width, etc. aren't in the footprint name. Here is a great example. How would you like to handle this if a single generic footprint would work for you?

I'm personally inclined to do a 'max' kinda of thing by choosing the package which would appear to generate a safe footprint that would work for both packages. Two manufacturer-specific packages, as mentioned above, is another option but it seems less elegant to me.

cnieves1:
ping @poeschlr 
Can I do a 'max' footprint, as suggested above?
That is:
  - overall height: max=0.60
  - lead width: min=0.15, max=0.30
  - lead length: min=0.35, max=0.60
Would this be ok?

Thanks!

nickoe:
@evanshultz  Friendly poke. What do you think about @cnieves1 suggestion?

myfreescalewebpage:
@evanshultz ping

myfreescalewebpage:
@cnieves1 are you still working on this subject.
To answer your question, I think you can. Make it and check if the generator is happy. If yes you can update the footprint. I will do a review after, ping me when it's ready.
Joel

cnieves1:
@myfreescalewebpage Hi Joel, changes done... Thanks,

chschlue:
Not sure if I'm missing something here, but the Resurgent package has no EP while the Microchip one does, so there's no naming collision and no need for compatibility, right?

myfreescalewebpage:
I'm on this too with you @chschlue you're right there is no name collision and no need to find compatibility solution.
@cnieves1 also can you remove all lines which are commented please ?

myfreescalewebpage:
Thanks
@cnieves1 also can you remove all lines which are commented please ?

myfreescalewebpage:
Fp has been merged

PR #524, opened 2020-04-08
QFN-64 with 5.2x5.2mm exposed pad.

ObKo:
Documentation: https://www.silabs.com/documents/public/data-sheets/Si5345-44-42-D-DataSheet.pdf#page=51


chschlue (approved):
Thanks.

PR #525, opened 2020-04-08
Add JST XA

jhalmen:
used to be #484 
created this pull request, as i had no permission to update the original request with my commits.

@fenugrec mind taking a look at my updates?

### DIFF
--- a/scripts/Connector/Connector_JST/conn_jst_XA_horizontal.py
+++ b/scripts/Connector/Connector_JST/conn_jst_XA_horizontal.py
@@ -135,15 +135,31 @@ def generate_one_footprint(pincount, variant, configuration):
     part_x_max = x_max
     part_y_min = y_min
     part_y_max = y_max
+    offset = configuration['courtyard_offset']['connector']
+    grid = configuration['courtyard_grid']
 
-    cx1 = roundToBase(part_x_min-configuration['courtyard_offset']['connector'], configuration['courtyard_grid'])
-    cy1 = roundToBase(part_y_min-configuration['courtyard_offset']['connector'], configuration['courtyard_grid'])
+    cx1 = roundToBase(part_x_min - offset, grid)
+    cy1 = roundToBase(part_y_min - offset, grid)
 
-    cx2 = roundToBase(part_x_max+configuration['courtyard_offset']['connector'], configuration['courtyard_grid'])
-    cy2 = roundToBase(part_y_max+configuration['courtyard_offset']['connector'], configuration['courtyard_grid'])
+    cx2 = roundToBase(part_x_max + offset, grid)
+    cy2 = roundToBase(part_y_max + offset, grid)
 
-    kicad_mod.append(RectLine(
-        start=[cx1, cy1], end=[cx2, cy2],
+    cx3 = roundToBase(part_x_max - body_back_protrusion_width - offset, grid)
+    cy3 = roundToBase(-pad_size[1]/2 - offset, grid)
+
+    cx4 = roundToBase(part_x_min + body_back_protrusion_width + offset, grid)
+
+    kicad_mod.append(PolygoneLine(polygone=[
+            {'x':cx1, 'y':cy1},
+            {'x':cx1, 'y':cy2},
+            {'x':cx2, 'y':cy2},
+            {'x':cx2, 'y':cy1},
+            {'x':cx3, 'y':cy1},
+            {'x':cx3, 'y':cy3},
+            {'x':cx4, 'y':cy3},
+            {'x':cx4, 'y':cy1},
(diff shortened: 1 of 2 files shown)


jhalmen:
@fenugrec in case you're satisfied with them, it would be cool if you could also update https://github.com/KiCad/kicad-footprints/pull/2169

fenugrec:
ok will check later today

fenugrec:
done

scripts/Connector/Connector_JST/conn_jst_XA_horizontal.py
```diff
@@ -0,0 +1,333 @@
...
+pad_to_pad_clearance = 0.8
+pad_copper_y_solder_length = 0.5 #How much copper should be in y direction?
+min_annular_ring = 0.15
+
+hook_size = [2.8, 2]    #oval hole
+
+# Connector Parameters
+y_max = 9.2
+y_min = y_max-12.6
+y_main_min = y_max - 9.7    #plastic body extent behind row of pins
```
evanshultz, line 39: Please note this was measured from the CAD model or a real part in the comment.

scripts/Connector/Connector_JST/conn_jst_XA_horizontal.py
```diff
@@ -0,0 +1,333 @@
...
+
+    kicad_mod.append(PadArray(initial=1, start=[0, 0],
+        x_spacing=pitch, pincount=pincount,
+        size=pad_size, drill=drill_size,
+        type=Pad.TYPE_THT, shape=Pad.SHAPE_OVAL, layers=Pad.LAYERS_THT,
+        **optional_pad_params))
+
+    #add mounting hole for hook
+    if hook:
+        kicad_mod.append(Pad(at={'x': A/2, 'y': 6.5}, type=Pad.TYPE_NPTH, shape=Pad.SHAPE_OVAL, layers=Pad.LAYERS_NPTH,
```
evanshultz, line 249: For the y position, please calculate from the datasheet values (9.2 - 2.7) instead of a magic number.

evanshultz, line 249: You've already got `x_mid` above so you don't need to calculate the X location again.

scripts/Connector/Connector_JST/conn_jst_XA_vertical.py
```diff
@@ -0,0 +1,271 @@
...
+
+#Note : this script should cover both standard and "N type" variants. Apparently the N type only uses a different platic:
+# https://jsttac.blogspot.com/2011/08/what-is-difference-in-xa-series-header.html
+
+fab_pin1_marker_type = 1
+pin1_marker_offset = 0.3
+pin1_marker_linelen = 1.25
+
+drill_size = 0.95 #Datasheet: 0.9 +0.1/-0
+mh_drill = 1.25 #for boss
```
evanshultz, line 32: Please note the asymmetrical tolerance in the datasheet for this hole, too.

scripts/Connector/Connector_JST/conn_jst_XA_vertical.py
```diff
@@ -0,0 +1,271 @@
...
+pad_to_pad_clearance = 0.8
+pad_copper_y_solder_length = 0.5 #How much copper should be in y direction?
+min_annular_ring = 0.15
+
+
+
+pitch = 2.5
+
+# as of 2020 : available in 2- to 15-pin, 18 and 20-pin variants.
```
evanshultz, line 41: Please check this since the latest datasheet doesn't show 18 pin. I don't see it as of 30 Sep, anyway.

evanshultz:
1. I don't like the use of `silk_pin1_marker_type` and `fab_pin1_marker_type` having numeric values that aren't descriptive. Can you change the values to something salient? If this is done in other scripts then it's fine to leave it.
1. Please add `tht` to the script file names since they don't do the SMT footprints.
1. Add explicit comments to anywhere I missed that show if a dimension not in the datasheet was measured from a real part, from the 3D model, or somewhere else.
1. I made some comments on the horizontal script that also apply to the vertical one, but I didn't repeat them again. Please assume comments cover both orientations.
1. If the 18 pin doesn't exist, at least as of today, you can simplify the vertical script. But that being said, you could also do something more quick and dirty since perhaps things are apt to change. Either way.

If you can make each change in a separate commit that might be easier to review.

evanshultz:
@chschlue 
Can you please confirm that this repo won't be migrated? At least not yet? I merged the footprints but would like to get a few of these under-the-hood updates to the script before merging this.

chschlue:
@evanshultz yes

PR #528, opened 2020-04-15
Added TSOP-II-86_10.16x22.22mm_P0.5mm

prasetekus:
Datasheet: http://www.issi.com/WW/pdf/42-45S32800J.pdf


poeschlr:
The datasheet includes nominal dimensions in addition to min and max so please include them.
Include the package height (overall_height)
Include the pdf page number in the linke (suffix the link with `#page=60`)

PR #530, opened 2020-04-17
Added QFN-56 8x8 EP5.9

awjlogan:
Added YAML for QFN-56 with 5.9mm EP.


awjlogan:
> Please add nominal dimensions (even if they happen to be mean(min, max) and especially if they don't (like lead width in this case))

Done, thanks for the review @chschlue  :+1:

chschlue:
FP merged.

PR #532, opened 2020-04-17
Espressif typo

evanshultz:
https://www.espressif.com/sites/default/files/documentation/esp32_datasheet_en.pdf#page=47

As you can see, the pitch is 0.35mm, and correctly entered as a parameter, but a typo in the definition of this package. Also fixed the datasheet page link, which could have changed since the doc version when the size definition was added.


nickoe:
@evanshultz I just reviewed this, it does indeed look like the pitch should be 0.35 mm. But when I download the datasheet pdf the info is in  "Figure 9: QFN48 (5x5 mm) Package" on page 42.  Why do we even borther to add a page number in that URL?

Currently the version of the datasheet is listed as "ESP32 Datasheet V3.4" dated 2020-04-27.

evanshultz:
It's a convenience. Yes, the doc is liable to get stale but at least it's usually only a page or a few pages off so it's faster to find. Without a page anchor at all there's much more scrolling and hunting, we've found.

chschlue (approved):

PR #537, opened 2020-04-26
Add Texas NS/PS packages

chschlue:
https://www.ti.com/lit/ml/msop001a/msop001a.pdf
https://www.ti.com/lit/ml/msop002a/msop002a.pdf

https://github.com/KiCad/kicad-footprints/pull/2245


poeschlr:
thanks

PR #539, opened 2020-04-26
Append newline to output

chschlue:
All the "no newline at EOF" warnings keep triggering my OCD.

Please excuse bad Python style. I had a difficult childhood, I grew up with C.


poeschlr:
What do you mean with "no newline at EOF warning"? Which tool makes such warnings for KiCad files?

Also does KiCad itself add a newline?

chschlue:
eof

GitHub does, for example. So does `git diff`.

Yes, ModEdit output ends in 0x0A linefeed.

PR #511, opened 2020-03-16
Add 5.3mm SSOP packages from JEDEC MO-150-B

mansr:

PR #558, opened 2020-05-24
Add TQFP-48_7x7mm_P0.5mm definition

chschlue:
https://www.jedec.org/system/files/docs/MS-026D.pdf variation ABC
https://github.com/KiCad/kicad-footprints/pull/2300
TQFP-48_7x7mm_P0

### DIFF
--- a/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/tqfp.yaml
+++ b/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/tqfp.yaml
@@ -2,6 +2,19 @@ FileHeader:
   library_Suffix: 'QFP'
   device_type: 'TQFP'
 
+TQFP-48_7x7mm_P0.5mm:
+  size_source: 'https://www.jedec.org/system/files/docs/MS-026D.pdf var ABC'
+  body_size_x: 7.00 # basic
+  body_size_y: 7.00 # basic
+  overall_height: 1.20 # max
+  overall_size_x: 9.00 # basic
+  overall_size_y: 9.00 # basic
+  lead_width: 0.17 .. 0.22 .. 0.27
+  lead_len: 0.45 .. 0.60 .. 0.75
+  pitch: 0.50
+  num_pins_x: 12
+  num_pins_y: 12
+
 TQFP-48-1EP_7x7mm_P0.5mm_EP5.0x5.0mm:
   size_source: 'https://www.trinamic.com/fileadmin/assets/Products/ICs_Documents/TMC2100_datasheet_Rev1.08.pdf (page 45)'
   body_size_x:
--- a/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/tqfp.yaml
+++ b/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/tqfp.yaml
@@ -3,10 +3,10 @@ FileHeader:
   device_type: 'TQFP'
 
 TQFP-48_7x7mm_P0.5mm:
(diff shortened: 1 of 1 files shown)


evanshultz:
I'm not sure we want to specify footprints this way. Personally, I agree that using the standards-body doc makes the most sense. But since users require an account to get the doc from the official website and it's not in a datasheet is seems contrived to me. I'd rather see a part datasheet with the package in it since that seems the best user experience. The existing footprint appears to reference a Micrel package drawing, but it's not a full URL and I didn't bother to research it. @poeschlr ?

A couple other things:
1. I think a parameter is assumed to be nominal if only one value is given. So the height doesn't work to just put that one value since it's 1.2mm from board to the top of the package.
1. All other dimensions look good to me.
1. I don't see 'TQFP' spelled out in the doc, but it does use the word 'THIN' in the title. The column headers on page 5 all use 'PQFP'. I'm not sure what is the best way to call this package.

chschlue:
First of all, it's the _IPC 7351_ footprint generator, for crying out loud. That spec isn't freely accessible at all.
Also, what are we going to do if we use a datasheet (rightly) claiming JEDEC-compliance but with tighter tolerances? We'd need another FP for a different manufacturer's package specified with the original tolerances.
For generic footprints, I don't see a cleaner way than using the actual standard if available.

Regarding the other things:
1. True, but the height isn't used at all currently. If and when we're actually going to generate 3D models from the same size definitions, using A1 (standoff) and A2 (actual body height) instead of A (overall height) seems to be the proper way to go to me. It's also pretty common for manufacturers (and JEDEC for that matter) to specify max overall height only.
2. Great.
3. I think using the vernacular ((T)QFP) is a good compromise towards user friendliness. While the official designator (**T**hin-**P**lastic**Q**uad**F**lat**P**ack-**G**ullwing in this case) would be cleaner, those get a little cryptic at times. Semi-OT: @poeschlr ordered the removal of height prefixes for v6, BTW.

Finally: I didn't bother looking up the Micrel drawing either. http://ww1.microchip.com/downloads/en/DeviceDoc/48L_TQFP_7x7x1_0mm_PT_C04-00300d.pdf is the package I was going to use this with.

Even more finally: I think we should be happy about bulk additions like #500 or #511 instead of complaining about the size source and asking for "real" datasheets for each of those footprints.

poeschlr:
While JEDEC does not give access to their data without login it is easy to get a login. I mean i would love they just put their info out there for all to see but well we are not the ones making this decision. And yes i am aware that i was originally the one against it but you guys brought me around.

---

Regarding height: for all other parts we simply add the maximum specifier. However, i did not add a way to specify it with the new style so use the old style for that dimension at least for now (i could imagine we could add ..<maximum> and <minimum>.. as options to indicate that something is given as one of the extremes.)

myfreescalewebpage:
I agree with @evanshultz that available datasheet are probably better for most of the user. Registering to JEDEC need to have a professional email isn't ?

This footprint is used for example by ATSAMD21G15A-A symbol, datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/SAM_D21_DA1_Family_Data%20Sheet_DS40001882E.pdf#page=1062 and based on this I agree on the dimensions but I'm surprised of the way to declare them in the yaml file. I was more nominal+tolerance fields.

I also see the datasheet above indicate the min/max values while @chschlue your definition for this footprint has only the nominal value and no tolerance.

Joel

chschlue:
Here you go.

myfreescalewebpage:
@evanshultz is it good for you ?

evanshultz:
Yes. Since height isn't used, more work would have to be done there anyway if we were to use it.

PR #575, opened 2020-07-15
csp.yaml - Add X1-WLB0909 footprint

chmorgan:
Datasheet: https://www.diodes.com/assets/Datasheets/AP22913.pdf

### DIFF
--- a/scripts/Packages/Package_BGA/csp.yaml
+++ b/scripts/Packages/Package_BGA/csp.yaml
@@ -474,12 +474,12 @@ WLCSP-8_1.551x2.284mm_P0.5mm:
   pad_diameter: 0.2
   layout_x: 2
   layout_y: 4
-WLCSP-4-X1-WLB0909-4_0.890x0.890_P0.500mm:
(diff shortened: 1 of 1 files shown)


scripts/Packages/Package_BGA/csp.yaml
```diff
@@ -473,4 +473,13 @@ WLCSP-8_1.551x2.284mm_P0.5mm:
   pitch: 0.5
   pad_diameter: 0.2
   layout_x: 2
-  layout_y: 4
\ No newline at end of file
+  layout_y: 4
+WLCSP-4-X1-WLB0909-4_0.890x0.890_P0.500mm:
```
cpresser, line 477: The `mm` is missing from the xy-size. Also the trailing zero on 0.890 could be removed.

cpresser (approved):

cpresser:
Thanks, looks good.
Next time, please don't force-push so I can easier review changes. In this case it was not an issue since its only a few lines though.

PR #576, opened 2020-07-18
csp.yaml - Add Lattice WLCSP-16 package

chmorgan:


scripts/Packages/Package_BGA/csp.yaml
```diff
@@ -456,3 +456,14 @@ ST_WLCSP-180_Die451:
...
   row_skips: [[], [], [], [], [], [], [7], [7], [], [], [], [], [], []]
+WLCSP-16_1.409x1.409mm_P0.35mm:
+  description: "WLCSP-16"
+  size_source: 'http://www.latticesemi.com/view_document?document_id=213'
+  body_size_x: 1.409
+  body_size_y: 1.409
+  pitch: 0.35
+  pad_diameter: 0.218
+  mask_margin: 0.01
+  paste_margin: 0.000001
```
cpresser, line 467: Why did you choose those parameters?

* mask_margin and paste_margin seem to be arbitraty values. If there is no manufacturer spec, just leave them empty. Then the boards settings will be applied when generating paste and mask layers.
* As for the pad_diameter, I suggest to go for 0.2mm

cpresser (changes requested):
I got the 0.2mm from here:
image

Page8: http://www.latticesemi.com/-/media/LatticeSemi/Documents/ApplicationNotes/PT2/FPGA-TN-02024-4-1-PCB-Layout-Rec-BGA-Packages.ashx?document_id=671

scripts/Packages/Package_BGA/csp.yaml
chmorgan: @cpresser I might have copied them from another part not realizing that leaving them out is ok. Pushed an update that adjusts the pad_diameter to 0.2mm after confirming from the datasheet you linked above.

cpresser (approved):

PR #587, opened 2020-08-24
Added Dialog_VFBGA-86_6x6mm_Layout10x10_P0.55mm_Ball0.25mm_Pad0.2mm

aectaan:
[footprint PR](https://github.com/KiCad/kicad-footprints/pull/2188)


scripts/Packages/Package_BGA/bga.yaml
```diff
@@ -345,3 +345,14 @@ WLP-9_1.448x1.468mm_Layout3x3_P0.4mm_Ball0.27mm_Pad0.25mm:
   paste_margin: 0.000001
   layout_x: 3
   layout_y: 3
+Dialog_VFBGA-86_6x6mm_Layout10x10_P0.55mm_Ball0.25mm_Pad0.2mm:
```
cpresser, line 348: Ball diameter is in the name, but not set as a property.

scripts/Packages/Package_BGA/bga.yaml
```diff
@@ -345,3 +345,14 @@ WLP-9_1.448x1.468mm_Layout3x3_P0.4mm_Ball0.27mm_Pad0.25mm:
...
+  description: "VFBGA-86"
+  size_source: "https://www.dialog-semiconductor.com/sites/default/files/da1469x_datasheet_3v1.pdf"
+  body_size_x: 6.0
+  body_size_y: 6.0
+  pitch: 0.55
+  pad_diameter: 0.205
+  mask_margin: 0.026
```
cpresser, line 355: Where did you get those values from (pad_diameter, mask_margin)? I feel stupid since I can't find it.

cpresser (commented):
Looks pretty good. I have two questions/remarks though.

scripts/Packages/Package_BGA/bga.yaml
aectaan: It's experimental data. I've made 7 boards with this MCU since my PR

scripts/Packages/Package_BGA/bga.yaml
aectaan: It's not really clear for me. What do you mean? I looked at the previous part (WLP-9_1.448x1.468mm_Layout3x3_P0.4mm_Ball0.27mm_Pad0.25mm) and made similar.

cpresser: I was stupid, misread a line, made a stupid assumption and wrote that comment. Ignore it.

scripts/Packages/Package_BGA/bga.yaml
cpresser: Fair, but we try to not base footprints on experimental value if possible. The manufacturers recommendations are to be followed. But here we don't have any.

In this case I suggest to use 0.2mm for the pad as specified in IPC-7351B (https://github.com/pointhi/kicad-footprint-generator/blob/d3b8ffd00e656a0ccff3a75be2c9cb8aeec3f6df/scripts/Packages/Package_BGA/ipc_definitions.yaml)

Also, I checked the snapeda.com export (dialog links to them for footprints/symbols). They have 0.2mm as well.
```
  (pad A1 smd circle (at -2.475 -2.475) (size 0.2 0.2) (layers F.Cu F.Mask F.Paste))
  (pad A2 smd circle (at -1.925 -2.475) (size 0.2 0.2) (layers F.Cu F.Mask F.Paste))
``` 

Regarding mask margin I am actually not sure how to continue. Never had that issue before.
* https://github.com/KiCad/kicad-footprints/pull/1170#issuecomment-447583481 is about the same value
* I agree that having a value set that is proven in production is a good idea.
* The KLC does explicitly allow to set the mask in [F9.2](https://kicad-pcb.org/libraries/klc/F9.2/).
* From https://www.ti.com/lit/an/sprabb3/sprabb3.pdf?#page=10
  * image
*image
* One option would be to remove the `mask_margin` parameter completely. Then the settings as defined in the board-file where the footprint is used will be applied to the pads.
* Or take the value from TI (0.05mm) which is 10 years old
* Or take 0.035 which lots of other parts in that file use
* Or we just go with your proven value which is close to other values in the same file. One question here though, why did you choose that value?

evanshultz:
We should use IPC7351 pads sizes for ball sizes if no recommended footprints are given, and arguably even if they are as we do the same thing with gullwing and SON packages so the same justification works here.

That info is captured at https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/Packages/Package_BGA/ipc_definitions.yaml, but I haven't taught the generator how to use it yet. You can still read that file and manually pick the proper pad size, though.

That is a really tight mask margin. I don't see guidance from IPC on that, but I also suggest to leave it alone and have it controlled by the board. I added the full complement of parameters to the generator to make the recommended footprints for ST parts, but if the vendor doesn't tell us how they want the footprint made then we can give the user control once we get past IPC guidelines.

Lastly, this package like isn't only for Dialog. It looks like a generic package. And Dialog doesn't have a recommended footprint, so there's nothing special about the resulting footprint. Therefore, don't make it Dialog-specific but just make it generic.

scripts/Packages/Package_BGA/bga.yaml
aectaan: Regarding your last question: it's a value, recommended by the local PCB manufacturer. Anyway, I liked your idea to take the most common value from this file (0.035)

aectaan:
I changed pad diameter and mask margin (0.2/0.035). I'm disagree, that it is generic package: if you try to google "vfbga-86", 6/10 results will be about DA14691/DA14695, and 4 will be not relevant.

evanshultz:
There's nothing special about the package or the footprint that will result from it I can see. Dialog can call it VFBGA while another company calls the same package something else, but the package can still be a standard, off-the-shelf package available from the typical package vendors. Do you have more information about the package and if it is specific? And without a recommended footprint I really don't see anything tying this footprint to Dialog.

If you do want it to be specific, it appears https://www.dialog-semiconductor.com/sites/default/files/da14695_devkt-b-usb_vc_allegro_source.zip is the layout in Allegro. I could open it and extract any unknown information for you. Then we'd have Dialog's recommendation, as best as I can see, and it would make sense for this to be a specific footprint. Let me know if you would like to me do that (I'm guessing you don't have Allegro). We are missing pad diameter and mask clearance, right?

Lastly, we often add an anchor to the datasheet page as a convenience. You could do that to the link for page 740.

aectaan:
Ok, I will remove "Dialog_" prefix. Could you please provide pad diameter and mask opening?

cpresser:
Lets go with Pad 0.2mm (per IPC) and no mask parameter at all.

I also tried the free Allegro-Viewer for fun. But was not that successful - I don't know how to properly use it. The grid was not fine enough to measure it exactly. But I estimate the pad in that design is 0.2 and the mask 0.25 (both diameter).

cpresser:
Please just omit the mask-margin parameter.
Currently it has no effect since 0 is treated as *use board settings*. Which is the same as not specifying it at all.
But we never know how KiCad will change. So better leave it empty.

cpresser:
Forgot one thing: Please update the `kicad-footprints` PR with the generated file.

aectaan:
done

cpresser:
The `mask_margin` line is still part of the change-set. Please remove that line. See my comment above.
That's the last item, after that we can merge it.
I will go ahead and merge the generated Footprint since it won't be affected by that change.

cpresser (approved):

PR #591, opened 2020-08-26
Add dict_tools.py with dictionary inheritance/merge functions

calebreister:
This is a script I created as part of #397 to create an "inheritance" system for Python dict objects (actually it should work with any instance of `collections.abc.Mapping`). As per @chschlue's request, I've created this PR so that the script can be introduced into the main repo sooner for others to use. Specifically, it may be useful for solving an issue in #586. Although I do not have unit tests, I have used the script and tested it in a variety of conditions.

### DIFF
--- a/scripts/tools/dict_tools.py
+++ b/scripts/tools/dict_tools.py
@@ -1,0 +1,119 @@
+#!/usr/bin/env python
+
+import collections.abc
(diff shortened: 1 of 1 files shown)


pointhi:
I'm fine with merging it

Issue #55, opened 2017-09-12
Create some sane external behaviour of the scripts


pointhi:
@jkriege2, @hackscribble, @SchrodingersGat, @poeschlr 

We are a small group of people which created most of the script generated parts inside KiCad. First of, thanks for using my stuff ^^.

In contrast to the framwork itself, I'm merging contributions into scripts with almost no review. Simply because most scripts are used mostly a few times by mostly the same people.

This results in some collection of code where it's hard to use/change existing scripts, as well as the inventing of scripts which get better placed into the framework itself ([like the KLC definitions](https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/tools/footprint_global_properties.py)).

I would like to start a discussion about how much we can unify our development of scripts, to help new contributors as well as ourself.

The main proposal could be described as [separate code and data](http://wiki.c2.com/?SeparationOfDataAndCode). This is a main statement when talking about secure coding, but has other reasons as well. Everyone invented his system to get the data into the system. (I invented at least 3 different ones ^^). This means people can not simply use those data for other purposes (like generating 3d-models), and people are required to decrypt the way of data storage to push their own values into the system.

I added [ModArgparser.py](https://github.com/pointhi/kicad-footprint-generator/blob/master/KicadModTree/ModArgparser.py) as my proposed fix for this problem into the framwork. Although no one really uses it right now beside me. This class would allow us to unify the external behaviour of all the scripts, means everyone can push data to it using .csv or .yml files. Furthermore, every script get's the same comandline interface, for interaction.

I would ask about your opinion about those problems, as well as what are you thinking about my proposal. I'm especially thinking about the usage of ModArgparser for 3d-file generation as well. This would allow us to use the same part definitions for both tasks.

poeschlr:
Seems like a good idea. Currently i am a bit short on time to look into it in great detail. If i don't report back in the next few weeks give me another notification. (I might forget about this otherwise.)

hackscribble:
Hi @pointhi 

I'm also short of time at present, but here are some initial thoughts - probably raising more questions than giving answers :disappointed: 

1. Your framework is a *tremendous* contribution to the KiCad ecosystem.  Many thanks for creating and supporting it.

[...]

PR #592, opened 2020-08-26
Derive pad size from IPC rules

evanshultz:
When possible, derive BGA pad sizes from IPC 7351. If not, fall back to pad sizes specified in the YAML file as currently done.

Renamed the IPC land pattern info file so it's not the same as the general IPC info file. Also added a test YAML file to show this new behavior.


evanshultz:
Pushed another commit to specify the IPC info is for 7351 revision B since the rules might be different in a future rev C.

scripts/Packages/Package_BGA/bga.py
```diff
@@ -153,6 +153,19 @@ def crtYdRound(x):
 
     # Pads
     balls = layoutX * layoutY
+    
+    # use IPC-derived pad size if possible, then fall back to user-defined pads
+    if (("ball_type" in fpParams) and ("ball_diameter" in fpParams)):
+        padSize = configuration['collapsible'][fpParams["ball_diameter"]]["min"]
```
cpresser, line 159: why is `collapsible` hard-coded

cpresser: What happens if that ball_diameter is not found in the `.yaml` file?
I think some checking if the required values are present is needed here.

scripts/Packages/Package_BGA/bga.py
```diff
@@ -153,6 +153,19 @@ def crtYdRound(x):
...
+        padSize = [padSize, padSize]
+        if "pad_diameter" in fpParams:
+            print("Pad size is being derived using IPC rules even though pad diameter is defined.")
+    elif "pad_diameter" in fpParams:
+        padSize = [fpParams["pad_diameter"], fpParams["pad_diameter"]]
+    else:
+        print("\033[1;35mThe config file must include 'ball_type' and 'ball_diameter' or 'pad_diameter'. Exiting...\033[0m")
```
cpresser, line 166: I don't really like the escape-codes. It is hard to read. At least I don't know out of the back of my head which colors that are.
A similar print() is also used in `scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/ipc_noLead_generator.py` where the same comment applies.

Perhaps add a convenience function like `print_color()`

scripts/Packages/Package_BGA/ipc_test.yaml
```diff
@@ -0,0 +1,43 @@
+Dialog_VFBGA-86_6x6mm_Layout10x10_P0.55mm_Ball0.25mm_IPC:
```
cpresser, line 1: How about renaming all of them, to indicate what is tested.
Like 
* `test_only_ball_definition`
* `test_ball_and_pad`
* ...

And remove the stuff that is not needed for the test (datasheet, row_skip).
And add a file header saying what is tested and why.

cpresser (changes requested):
I would also some kind of documentation for the users. When to use which option? IMHO we should encourage users to just set ball_diameter and ball_type and let the script handle the rest.
But that can be a separate PR.

Regarding the `ipc_test.yaml` file: I don't like its location. But it is consistent with `test_so.yaml`. And I don't have a better location.

scripts/Packages/Package_BGA/bga.py
evanshultz: Whoops! Fixed the ball type being fixed and not read from the YAML file.

I made a separate function to handle argument hygiene and then pass onto the footprint generation. Much like the gullwing generator script. This would also allow an extension of the script to generate footprints with multiple IPC densities or something (this isn't done on any of the Package* scripts but it applies to all of them).

scripts/Packages/Package_BGA/bga.py
evanshultz: True. I switched to a `KeyError`, which applies well. Adding that helper function for printing in color would be nice. I'll do that later.

chschlue: Tangent:
Python probably has some kind of termcap/terminfo support.
How about only outputting color when supported by the output device?

scripts/Packages/Package_BGA/ipc_test.yaml
evanshultz: Followed the file name convention Rene used along with added comments to the file and streamlined as you mention.

evanshultz:
@cpresser 
Fixes and enhancements pushed. Thank you for looking.

Should the script be renamed to `ipc_bga_generator.py` to match the other Package* generator scripts and reflect the fact that it now adheres to IPC rules (which are far fewer for BGA than other packages types)?

Documentation is a good idea (and it will reveal ever more features missing that the gullwing and SON generators have. And I should have made the skipped_row letters a parameter with a default list instead of hard-coded. Doing that all later sounds good.

cpresser:
> @cpresser
> Fixes and enhancements pushed. Thank you for looking.

> 
> Should the script be renamed to `ipc_bga_generator.py` to match the other Package* generator scripts and reflect the fact that it now adheres to IPC rules (which are far fewer for BGA than other packages types)?

That sounds good. I like consistency.
But I can not really be a judge on how compliant that generator is to IPC standards. I still have no access to that spec.

> 
> Documentation is a good idea (and it will reveal ever more features missing that the gullwing and SON generators have. And I should have made the skipped_row letters a parameter with a default list instead of hard-coded. Doing that all later sounds good.

TBH, I have no clue what the `skipped_row` parameter is. A quick regexp search doesn't show anyting.

evanshultz:
Changed the generator script name. Anything else?

One line 256 you can see a list of the row names that are not used. I shouldn't have hard-coded that but it is, for now. I don't think it affects this, it was just something else that should be updated later that I noticed.

cpresser:
I am still not happy with the error messages generated.
I would prefer if the output would say something like  `Pad size 0.1337mm is not a valid IPC diameter. Please see ipc_7351b_bga_land_patterns.yaml for permissible values`, with the same for the `ball_type` key.
(Please refine those messages with proper English wording. My examples as a non native speaker are usually not that good).

To illustrate this I made  new test cases like this
```yaml
Collapsible_balls_wrong_diameter:
  description: 'BGA-100'
  size_source: ''
  body_size_x: 6.0
  body_size_y: 6.0
  pitch: 0.55
  ball_type: 'collapsible'
  ball_diameter: 0.1337
  layout_x: 10
  layout_y: 10

WrongBallType_balls_wrong_diameter:
  description: 'BGA-100'
  size_source: ''
  body_size_x: 6.0
  body_size_y: 6.0
  pitch: 0.55
  ball_type: 'foobar'
  ball_diameter: 0.2511
  layout_x: 10
  layout_y: 10

```
image

evanshultz:
Oh. I see what you mean now. I was thinking about when IPC rules weren't being used and I realized that this is valid since we can easily support non-IPC footprints with this generator (unlike the gullwing and SON ones). Hang on...

cpresser:
> Changed the generator script name. Anything else?
> 
> One line 256 you can see a list of the row names that are not used. I shouldn't have hard-coded that but it is, for now. I don't think it affects this, it was just something else that should be updated later that I noticed.

Interesting. Why are those letters not used? Could you add a comment that explains why they are omitted?

evanshultz:
Those letters are easy to confuse with numbers or other letters, especially when viewing printed schematics or when rendered in silkscreen, and so they're commonly disused. AFAIK this is the common set of unused letters across the industry but it may be that somebody wants a different set so we might as well give the user control to override this list if they wish.

evanshultz:
Your English above was perfect but I tried to be a bit more brief in case the console window is small. Let me know if this looks good.

scripts/Packages/Package_BGA/ipc_bga_generator.py
```diff
@@ -18,8 +18,38 @@
...
+    elif "ball_type" in fpParams and not "ball_diameter" in fpParams:
+        raise KeyError("Ball diameter is missing. No footprint generated.")
+    elif "ball_diameter" in fpParams and not "ball_type" in fpParams:
+        raise KeyError("Ball type is missing. No footprint generated.")
+    elif "pad_diameter" in fpParams:
+        fpParams["pad_size"] = [fpParams["pad_diameter"], fpParams["pad_diameter"]]
+        print("Pads size is set by the footprint definition. This should only be done for manufacturer-specific footprints.")
+        createFp = True
+    else:
+        raise KeyError("The config file must include 'ball_type' and 'ball_diameter' or 'pad_diameter'. No footprint generated.")
```
cpresser, line 47: I don't like that the behavior is not consistent. Some branches throw a KeyError, others print a message (and keep `createFp = false`)
I would change them all to just print the error message.
Or is there a good reason for the current scheme that I am missing?

scripts/Packages/Package_BGA/ipc_bga_generator.py
```diff
@@ -218,6 +253,12 @@ def rowNameGenerator(seq):
         # except yaml.YAMLError as exc:
             # print(exc)
     
+    with open(args.ipc_doc, 'r') as config_stream:
+        try:
+            configuration.update(yaml.safe_load(config_stream))
+        except yaml.YAMLError as exc:
+            print(exc)
```
cpresser, line 260: This exception should generate a better error message.
* File not found/not readable
* Yaml parsing error
Also the script should exit here.

The same applies to the block that reads the config file. Since both of those blocks basically do the same I suggest to refactor them into a function.
```python
for fn in [args.ipc_doc, ags.global_config]:
  update_configuration_from_yaml(fn)
`

cpresser (commented):
Sorry, I could not finish the review yesterday. It was to late in my timezone.

scripts/Packages/Package_BGA/ipc_bga_generator.py
evanshultz: OK Changed them all to print(). I think this is rather inconsistent in all the IPC generator scripts (and maybe others, too).

scripts/Packages/Package_BGA/ipc_bga_generator.py
evanshultz: This applies to all the IPC generator scripts, and in fact I think many non-IPC generator scripts. In fact, I think it falls into the purview of https://github.com/pointhi/kicad-footprint-generator/issues/55, and would best be left as a bigger exercise once there is an agreed-upon direction for how to interface with our scripts. Then figuring out what errors can result and implementing them can happen uniformly.

evanshultz:
@cpresser 
Thanks for looking again. I updated one item but the other one I think is a bigger issue that is best dealt with outside of this PR.

cpresser:
I agree, that is out of scope. I would prefer to squash-merge this. Or what is the usual approach on this repo?

evanshultz:
That is what I have done in the past. @pointhi , do you have any preference?

cpresser (approved):

pointhi:
@evanshultz I do not have any strong opinion regarding merge policy. Please do as you like.

PR #579, opened 2020-07-22
Add Traco DCDC parts in Single Inline Packages

Werni2A:
This script generates footprints for the following series:
TBA 1, TBA 1E, TBA 1HI, TBA 2, TMA, TMAP, TME, TMR 1, TMR 2, TMR 2E, TMR 2WI, TMR 3, TMR 3E, TMR 3HI, TMR 3WI, TMR 6, TMR 6WI, TMR 9WI, TMV, TMV-EN, TMV-HI, TMV 2HI, TMW 2WIN, TRV 1M, TEC 2, TEC 2WI, TEC 3, TEC 3WI, TMH, TRA 3, TEA 1, TEA 1E,  TEA 1HI.

- Some hand-made footprints can be replace with this script, should this be discussed in the footprint PR?

- Initially I wrote down all relevant distances in the footprint drawings in a table before I copied it over to the script. This might be helpful for checking the PR.
[Part_Data.zip](https://github.com/pointhi/kicad-footprint-generator/files/4959922/Part_Data.zip)

If there is anything to change please let me know.

Todo:
- [ ] Remove example
- [ ] Rename and copy into subfolder
- [ ] Rename `missing_pins` to `hidden_pins`

PR #596, opened 2020-08-27
Add Feature for Missing Pins in SIP Footprints

Werni2A:
This PR adds a functionality to leave out certain pins. This is required in e.g. [Tracos TEA 1HI](https://www.tracopower.com/sites/default/files/products/datasheets/tea1hi_datasheet.pdf) devices but there are a lot more.
image

In the following you can see an example footprint that was generated with the script (note that the footprint uses just random values and does not represent a specific device):
image

As a reference, I used this modification in #579 .


cpresser (approved):
Thanks, that looks good. Simple and straight forward.

PR #599, opened 2020-08-31
Added BGA-200 footprint definition for IS43LQ32256A-062BLI.

mhrdmhrd:
Corresponding PR to KiCad footprint repository: https://github.com/KiCad/kicad-footprints/pull/2437

Screenshot:
image

~~The chamfer looks like too large for this device, but I could not find any adjustment way for this issue. The chamfer parameters seemed to defined in the global config.~~

Datasheet: http://www.issi.com/WW/pdf/43-46LQ32256A-AL.pdf
image


evanshultz:
Thanks for making the first contribution to this repo.

The bevel size is controlled by KLC F5.2.2.b (https://kicad-pcb.org/libraries/klc/F5.2/) and the generator uses this value.

Please submit the resulting footprint to https://github.com/KiCad/kicad-footprints; we usually review the script changes here thoroughly and then trust the footprint to likely be correct and merge it first. Then cross-link the script PR and footprint PR.

Thank you!

mhrdmhrd:
Thanks for your time to let me know the background of bevel size.
The corresponding PR to https://github.com/KiCad/kicad-footprints is: https://github.com/KiCad/kicad-footprints/pull/2437

Thank you!

evanshultz:
The BGA generator now supports deriving the pad size from the ball diameters in IPC 7351B. 0.32mm nominal ball diameter isn't in the doc but 0.3mm and 0.35mm ball diameters bracket your pad size of 0.27mm. The doc says 20% reduction which isn't exactly 0.27mm, but the table doesn't use exactly 20% reductions either. So we should be OK.

Please remove the mask and paste margin. They default to 0.

cpresser:
> Edit: @cpresser I didn't realize putting `[1, 13]` would eliminate the entire row, but it does. I wonder if that's an accident. What if only the first and thirteenth balls should be removed? That seems impossible. I didn't look into the code, but from the outside does this look like a mistake in the generator to you?

Yep, that is unexpected. I am slightly bamboozled. 
Please open a new issue for that.

mhrdmhrd:
@evanshultz
Thank you for checking. I have removed mask and paste margin.

@cpresser 
Following code from ipc_bga_generator.py (formerly bga.py) seems to process "row_skips" parameter.
I think line 195 and 196 treats nested square brackets [[1, **13**]] and the part eliminate the specified columns by range. 

```
184     # Pads                                                                 
185     balls = layoutX * layoutY                              
186                                                
187     if rowSkips == []:                                                                  
188         for _ in range(layoutY):                                          
189             rowSkips.append([])                                                             
190     for rowNum, row in zip(range(layoutY), rowNames):
191         rowSet = set(range(1, layoutX + 1))                               
192         for item in rowSkips[rowNum]:
193             try:                           
194                 # If item is a range, remove that range
195                 rowSet -= set(range(*item))
196                 balls -= item[1] - item[0]
197             except TypeError:                              
198                 # If item is an int, remove that int
199                 rowSet -= {item}                      
200                 balls -= 1
```

cpresser:
I think the most important task is to document how the generator works.
And make sure that the documentation keeps track with changes. Out-of-date documentation is even more shitty than non existing docs.

PR #595, opened 2020-08-27
Added SOP-20 for HT12D

G-Pereira:
https://github.com/KiCad/kicad-symbols/pull/1932

PR #600, opened 2020-08-31
Fix to #595

G-Pereira:
Hi @chschlue the PR #595 had an error that I just committed the fix


chschlue:
Oops.
Missed the doc update.

chschlue (approved):

PR #604, opened 2020-09-03
Support hidden and deleted pins

evanshultz:
As discussed at https://github.com/pointhi/kicad-footprint-generator/pull/586 and specifically at https://github.com/pointhi/kicad-footprint-generator/pull/586#issuecomment-680057204, I've added support for deleted pins (pins which are not present but do not consume a pin number) and custom pad layouts like SOT23.

Here are some example footprints (from test_hidden_deleted_pins.yaml):
image

I had to change scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/soic.yaml just to conform to this new terminology.

@pointhi 
The node PadArray seems to barely resemble the original implementation. There are so many caveats with deleted and hidden pins now. Since the values were being validated already, I chose to use a negative number as a filter since negative pin numbers seemed like they should not be allowed. Are you OK with this or do you have some other direction you'd like to pursue?


scripts/Packages/Package_Gullwing__QFP_SOIC_SO/ipc_gullwing_generator.py
```diff
@@ -155,6 +163,15 @@ def __createFootprintVariant(self, device_params, header, dimensions, with_therm
         size_y = dimensions['body_size_y'].nominal
 
         pincount = device_params['num_pins_x']*2 + device_params['num_pins_y']*2
+        
+        if 'hidden_pins' in device_params:
+            pincount_text = '{}-{}'.format(pincount - len(device_params['hidden_pins']), pincount)
+            pincount = device_params['num_pins_x']*2 + device_params['num_pins_y']*2 - len(device_params['hidden_pins'])
```
cpresser, line 169: Why would you use `pincount` and overwrite its value in the next line?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/ipc_gullwing_generator.py
```diff
@@ -140,10 +141,17 @@ def deviceDimensions(device_size_data):
     def generateFootprint(self, device_params, header):
         dimensions = Gullwing.deviceDimensions(device_params)
 
-        if dimensions['has_EP'] and 'thermal_vias' in device_params:
-            self.__createFootprintVariant(device_params, header, dimensions, True)
+        if 'deleted_pins' in device_params:
+            if type(device_params['deleted_pins']) is int:
+                device_params['deleted_pins'] = [device_params['deleted_pins']]
```
cpresser, line 146: Why is this (conditional) re-formating not done for hidden_pins?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/test_hidden_deleted_pins.yaml
```diff
@@ -0,0 +1,150 @@
...
+# hidden pins remove a pin location and pin number (for example, going from pad 1 to pad 3 with pin 2 missing from the package)
+# deleted pins remove a pin location but not pin number (for example, pin 2 is missing from the package but the footprint has pad 2)
+# deleted pins do not support custom numbering schemes https://github.com/pointhi/kicad-footprint-generator/pull/371
+
+FileHeader:
+  library_Suffix: 'TO_SOT_SMD'
+  device_type: 'SOT'
+
+# test add_quad_pad_border() in quad_dual_pad_border.py
+LQFP-48_7x7mm_P0.5mm:
```
cpresser, line 12: Please rename the test-cases so its clear what they are testing.
And make them minimal, `size_source` is not needed here.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/test_hidden_deleted_pins.yaml
```diff
@@ -0,0 +1,150 @@
...
+  lead_len: 0.45 .. 0.75
+  pitch: 0.5
+  num_pins_x: 12
+  num_pins_y: 12
+  deleted_pins: [1,15,24,30,31,44]
+
+# test add_dual_pad_border_y() in quad_dual_pad_border.py
+SOT-23-6:
+  size_source: 'https://www.jedec.org/sites/default/files/docs/Mo-178c.PDF variant AA'
+  custom_name_format: 'SOT-23-6R_Hidden'
```
cpresser, line 55: why `_Hidden` if this test-case has no hidden pins.

scripts/tools/pad_number_generators.py
```diff
@@ -23,9 +24,22 @@
 
 def increment(pincount, init=1, **kwargs):
     i = init
+    j = init # pad number for deleted pins
+
+    if kwargs.get("deleted_pins"):
+        skip_pins = kwargs["deleted_pins"]
```
cpresser, line 30: I really don't like this naming. I think the variable should be called `deleted_pins` as well.

scripts/tools/pad_number_generators.py
```diff
@@ -6,7 +6,8 @@
 
 Available generators:
 ----------------------
-- increment: Is a simple generator that increments the pin number by one.
+- increment: Generates pin numbers increasing by one and also supports skipping
+            pin numbers (skipped pins return a bogus value of -1)
```
cpresser, line 10: This should say that skipping pins is based on the `deleted_pins` keywordargs

cpresser (commented):
This is just an initial reading over the code.
I have yet to use and understand `scripts/tools/custom_pad_layout.py`.

I hope to get a full review done on the weekend.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/ipc_gullwing_generator.py
evanshultz: `deleted_pin` and `hidden_pins` can be an `int` or `list` type. I need a list, so if I get an int I stick it into a list.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/test_hidden_deleted_pins.yaml
evanshultz: I recalled the `size_source` removal discussion from the BGA pad size PR. But I checked and the gullwing and SON generators fail the same way if this argument isn't given so I kept it the same (for now).

evanshultz:
@cpresser 
Thanks! I'll take a closer look at this and make any updates a little later.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/ipc_gullwing_generator.py
cpresser: my point is that you are doing this only for `deleted_pins`.
where is the conversion for `hidden_pins`?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/test_hidden_deleted_pins.yaml
cpresser: Ah, I was not aware of that. Perhaps put something else there. The datasheet-link to a real part is confusion.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/ipc_gullwing_generator.py
evanshultz: Line 169 is the actual number of pins/pads present in the device, taking into account the hidden (removed) pins. So it's just subtracting the hidden pins from the pincount calculated above. Line 165 is the 'full' pin count used if there are no hidden or deleted pins. Yeah, no need to do the math again. You're saying why don't I take the value from line 165 and, if there are hidden pins, just subtract the number of hidden pins. I just missed that. Will fix.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/test_hidden_deleted_pins.yaml
evanshultz: I changed them to a tide. On the gullwing and SON generators they are assumed to be different than a manufacturer-specified footprint, and thus we need a datasheet or package drawing. The BGA generator could be used for a generic footprint or manufacturer-specific one. But either way, we still need some documentation against which to verify the footprint before we could add it to the library. So I don't see this as a problem in the general sense as some kind of documentation is required, but it is a little weird for test footprints.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/test_hidden_deleted_pins.yaml
evanshultz: Whoops. Copy and paste. Good catch.

scripts/tools/pad_number_generators.py
evanshultz: I tried to abstract things. In the PadArray node, for example, I kept the exclude_pin_list variable name but I allow two entry points of hidden_pins and deleted_pins since there are some cases where the same list works and other places where I need to know if the pins are hidden or deleted. Similarly, I decided to use a general term under the hood that could be used for other purposes in the future while the parameter name is specific to the behavior the user requests. .I can change it if you like.

scripts/tools/pad_number_generators.py
evanshultz: Alright. Updated.

scripts/tools/pad_number_generators.py
cpresser: How about that:
```
def increment(pincount, init=1, skip_pins=[]):
```
That makes it more abstract, it does not need to know about the `deleted_pins` keyword.

evanshultz: I thought about that, but I will still have to pick this keyword up in `get_generator()` so I figured it was essentially the same thing one place or the other. Is it better to grab this keyword in one place or the other?

evanshultz:
The node `Pad` only accepts int and string types for the pad number. Just below line 244 in PadArray this variable is used to determine if the pad is created or not (`includePad`). Since PadArray now includes iterators and many ways to create the pad numbers, not just simply building a list from the pincount argument, perhaps we should validate the type of the `number` after line 244? Then I could use some silly thing from line 35 of `pad_number_generators.py` instead of `-1` which would get filtered out. And I don't have to check for a specific keyword on line 246 of `PadArray`. @cpresser @pointhi Thoughts?

KicadModTree/nodes/specialized/PadArray.py
```diff
@@ -106,8 +109,13 @@ def _initPincount(self, **kwargs):
             raise ValueError('{pc} is an invalid value for pincount'.format(pc=self.pincount))
 
         self.exclude_pin_list = []
-        if kwargs.get('exclude_pin_list'):
-            self.exclude_pin_list = kwargs.get('exclude_pin_list')
+        if kwargs.get('hidden_pins') and kwargs.get('hidden_pins'):
```
cpresser, line 112: That should be `if kwargs.get('hidden_pins') and kwargs.get('deleted_pins'):`

cpresser:
I did play around with the generator a little and don't quite understand the need for `add_single_pin_right_center`.

The SOT-23 can also be build with:
```yaml
# can we do a SOT-23 without the custom_pad_layout? Seems like that is a yes!
SOT-23:
  size_source: '~'
  custom_name_format: 'SOT-23'
  body_size_x: 1.6
  body_size_y: 2.9
  overall_size_x: 2.8
  lead_width: 0.3 .. 0.5
  lead_len: 0.3 .. 0.45 .. 0.6
  pitch: 0.95
  num_pins_x: 0
  num_pins_y: 3
  deleted_pins: [2, 4 ,6]
```
It seems to identical to the one you build with the custom layout.

scripts/tools/pad_number_generators.py
cpresser: I think it is better to extract the keyword in the calling function.
That way the `increment` is more abstract. Also the user/programmer can clearly see what it does when looking at the arguments.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/ipc_gullwing_generator.py
cpresser: I think my problem here is that the variable `pincount` has different meanings just a few lines apart. How about something like this?

```python
        # pincount value and pincount text change if there are hidden or deleted pins 
        pincount_full = device_params['num_pins_x']*2 + device_params['num_pins_y']*2
        if 'hidden_pins' in device_params:
            pincount = pincount - len(device_params['hidden_pins'])
            pincount_text = '{}-{}'.format(pincount_full, pincount))
        elif 'deleted_pins' in device_params:
            pincount = pincount - len(device_params['deleted_pins'])
            pincount_text = '{}-{}'.format(pincount_full, pincount))
        else:
            pincount = pincount_full
            pincount_text = '{}'.format(pincount)
```

cpresser:
> The node Pad only accepts int and string types for the pad number. Just below line 244 in PadArray this variable is used to determine if the pad is created or not (includePad). Since PadArray now includes iterators and many ways to create the pad numbers, not just simply building a list from the pincount argument, perhaps we should validate the type of the number after line 244? Then I could use some silly thing from line 35 of pad_number_generators.py instead of -1 which would get filtered out. And I don't have to check for a specific keyword on line 246 of PadArray. @cpresser @pointhi Thoughts?

I took a short look at `PadArray`. It seems like the `pad_numbers` variable is populated in lines 212-223. 

How about returning `None` from the generator? And just having a check like
```python
if number == None:
  continue
```
after line 244?

KicadModTree/nodes/specialized/PadArray.py
evanshultz: Good catch. Fixed.

scripts/tools/pad_number_generators.py
evanshultz: Yes, that's what I was feeling. Sorry for not being more clear and/or verbose. I originally had the pin-skipping increment generator as a separate function but decided from the outside it was similar enough to just combine them. I'll leave this alone unless I hear another comment from you.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/ipc_gullwing_generator.py
evanshultz: Fine with me. I don't mind overwriting variables but that works too. Done.

I believe finding the pincount with hidden or deleted pins needs to subtract the length of the list from `pincount_full` so I made that correction from what is shown above. Let me know if I'm being a bonehead and my commit to address this isn't right.

I think it's too much to make a full document and/or rules out of it, considering the contributors writing scripts, but I wonder if there's a reason to gather up coding style preferences. I'd rather include someone's contribution if it works rather than turn them away or make it a burden if they're just learning and could have a more rewarding experience. And if the scripts, in general, are stocked with the desired coding style then it's easier to use them as a seed for starting a new script and get it reviewed and merged faster. (Additionally, I generally think all our scripts are dramatically under-commented but I usually follow along instead of adding the volume of comments my brain desires.)

evanshultz:
Something is now broken when using deleted_pins. Let me sort this out and then I'll commit with the changes I described above, and also reply to your last two comments about the SOT-23 footprint and PadArray. Hang on...

scripts/tools/pad_number_generators.py
evanshultz: Currently, custom pad numbering doesn't work with deleted pins. If that's going to work in the future the refactoring will need to be done since the pins to be deleted will have to be removed for any generator function and not just the increment one. But that can happen when/if needed.

evanshultz:
Sorry about the delay. I didn't have as much time as I expected to work on this the last couple days. I had it figured out everything with deleted pins before and then forgot what I was doing. I've added more comments now to hopefully help me and everybody else who (sadly) has to deal with my ineptitude.

Good call on the SOT-23. I felt stuck on the deleted pins so I did the custom pad layout for SOT-23. Once the deleted pins was working I didn't realize the deleted pins meant the one custom pad layout wasn't necessary. Thanks! I didn't remove the custom pad layout file, yet, so please confirm and then I'll remove it in another commit and clean up the YAML file.

We are thinking similarly about filtering out pads that are deleted. I did it slightly differently so that it handles any other disallowed type for a pad number.

Anything else?

evanshultz:
@cpresser 
I removed the custom pad layout stuff. Hopefully it just makes things easier if that's done since it seems unneeded now.

@pointhi 
Do you have any issues with the PadArray changes?

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/ipc_gullwing_generator.py
```diff
@@ -14,6 +14,7 @@
 from footprint_text_fields import addTextFields
 from ipc_pad_size_calculators import *
 from quad_dual_pad_border import add_dual_or_quad_pad_border
+from custom_pad_layout import add_custom_pad_layout
```
cpresser, line 17: That needs to go as well.

cpresser (commented):
I checked you recent commits and skimmed through all changes.
Looks good! :heart: 
Just one import that you missed.

scripts/Packages/Package_Gullwing__QFP_SOIC_SO/ipc_gullwing_generator.py
evanshultz: You got me! Fixed.

evanshultz:
@cpresser 
When thinking back on this, I realize that the changes to PadArray may not have been necessary. It could have kept the parameter `exclude_pin_list` since it really doesn't need to know if the pads are hidden or deleted. Filtering out pad numbers which aren't int or str gives me the hook to support deleted pins, and the hidden pins work just by slinging a list from some external source to `exclude_pin_list`. There's a level of abstraction that really isn't needed and it makes things more opaque because the meaning and use of hidden pins and deleted pins is outside of PadArray. Adding the type filter I mentioned earlier if a good change, but what do you think about undoing the rest of the changes to that file?

cpresser:
Tbh, I would need to look at that again in detail. But the overall idea sounds promising. PadArray should be as agnostic as possible. Feel free to rework this, I will gladly review it.

chschlue:
@evanshultz @evanshultz https://github.com/pointhi/kicad-footprint-generator/pull/645#issuecomment-700878817
Was that change intentional?

evanshultz:
No. Well, yes. If a standard footprint name is used, which is built inside the script, the pincount is always changed to a string so it works. But I didn't think about custom footprint names. Do you want me to push a fix?

chschlue:
Yes, please.
I'd rather not change every custom name, especially if the value is fundamentally a positive integer.

evanshultz:
It was a positive integer until we adopted the IPC naming format for deleted and hidden pins. I think it's the right way to go, but I had overlooked custom footprint names.

I submitted https://github.com/pointhi/kicad-footprint-generator/pull/649, which I think should take care of things.

PR #605, opened 2020-09-04
Add scripts for the generation of keyboard switch footprints

perigoso:
These scripts generate footprints for common switches used in keyboard applications, they include user drawing to aid the layout.

image
image
image
image
image
image

I understand the scripts might not be in great shape, my python is rough. Open to some feedback.

There is an issue with these scripts (or KicadModTree itself) it can be visualized in the second picture, it seems the `offset=[x,y]` seems to not be taken into account like it should in `kicad_mod.append(Pad(number=2, type=Pad.TYPE_THT, shape=Pad.SHAPE_OVAL, at=[-3.81,-2.54], size=[4.46156,2.5],  rotation=48, offset=[0.980778,0], drill=1.5, layers=['*.Cu', 'B.Mask']))` any input?


perigoso:
Detected Issues:

- [ ] 3D model settings
- [ ] Second Reference Designator missing

PR #606, opened 2020-09-05
Rework dictMerge() so it copies the first argument

calebreister:
I've edited `dict_tools.dictMerge()` based on a peer review. It now makes a deep copy of the first argument instead of overwriting it.


calebreister:
I've made some more extensive changes to the script in #397, which is the version I use. That makes this PR obsolete.

PR #609, opened 2020-09-05
WIP: Adds Recom RAC20-xxSK and RAC20-xxDK footprints

Werni2A:
Both footprints are for AC/DC converters so this is kind of "dirty" to place them in a DCDC script. Should I rename the script to just "Converter.yml"?

[Datasheet](https://recom-power.com/pdf/Powerline_AC-DC/RAC20-K.pdf#page=6)


aris-kimi:
https://github.com/KiCad/kicad-footprints/pull/2265

Werni2A:
This footprints should be created with a more generic DIP pack generator.

See https://github.com/KiCad/kicad-footprints/pull/2265 for a few of the footprint's issues.

Werni2A:
The used DCDC_Converter generator is inappropriate for this footprints. In the future I will write a new one and adapt this footprints. Therefore closing.

PR #616, opened 2020-09-10
Add WLCSP-8_1.551x2.284mm_P0.5mm

aectaan:


scripts/Packages/Package_BGA/csp.yaml
```diff
@@ -465,3 +465,12 @@ WLCSP-16_1.409x1.409mm_P0.35mm:
   pad_diameter: 0.2
   layout_x: 4
   layout_y: 4
+WLCSP-8_1.551x2.284mm_P0.5mm:
+  description: "WLCSP-8"
+  size_source: "https://www.adestotech.com/wp-content/uploads/AT25SL321_112.pdf#page=75"
+  body_size_x: 1.551
+  body_size_y: 2.284
+  pitch: 0.5
+  pad_diameter: 0.2
```
cpresser, line 474: The ball is `0.275mm`, so 0.2mm pad_diameter seems fine.
image
I could not find a manufacturer recommendation for the pad size.

However, I would like to use the new feature (#592) of the generator (extracting pad_size from ipc_7351b_bga_land_patterns.yaml). But it has no entry for 0.275mm
@evanshultz Should we just add value to the definitions file? Does IPC say anything about in-between diameters?

cpresser: I found another resource here: https://www.analog.com/media/en/technical-documentation/application-notes/AN-617.pdf
image

It has 0.2mm pad for a 0.26mm ball. Which is pretty close to this one.

---

@evanshultz regarding the IPC bga generator
How about adding a fallback that calculates the pad-size by
`round(0.8 + ball_diameter, 0.01)`? In this case that would yield 0.22mm.

PR #625, opened 2020-09-12
Add 205338 series SMT RA header for Molex Pico-Lock (2mm pitch) family

justyn:
This PR adds the PCB headers for the 2mm pitch version of the Molex Pico-Lock family.

Currently the section "PicoLock_side_entry" in conn_molex.yaml covers only the 1.5mm pitch versions.

For the values I started with the existing PicoLock_side_entry values then updated them using a combination of:
* The datasheet ([here](https://www.molex.com/pdm_docs/sd/2053380002_sd.pdf))
* The Molex 3D model for the 2-circuit and 4-circuit connector ([4ckt version here](https://www.molex.com/pdm_docs/stp/205338-0004_stp.zip))

I then made some minor manual adjustments to the outline so that it fit more evenly around the edge of the connector when the 3D model was added.

See the below images of 3D model superimposed on the footprint, deliberately intersecting to show alignment. Note that the top and bottom of the connector are different sizes so the body outline covers the extent. The lower image shows the model flipped so that the bottom of the connector can be compared to the outline.

datasheet

4ckt-footprint

4ckt-fp-3d

### DIFF
--- a/scripts/Connector/Connector_SMD_single_row_plus_mounting_pad/conn_molex.yaml
+++ b/scripts/Connector/Connector_SMD_single_row_plus_mounting_pad/conn_molex.yaml
@@ -340,6 +340,33 @@ device_definition:
             #width_end: 4.5
             end_from_body_side: 1.95
 
+    PicoLock_2mm_205338_side_entry:
+        series: 'Pico-Lock'
+        mpn_format_string: '205338-{pincount:04d}'
+        orientation: 'H'
+        datasheet: 'https://www.molex.com/pdm_docs/sd/2053380002_sd.pdf'
+        pinrange: ['list', [2,4,6,8,10]]
+        text_inside_pos: 'center'
+        pitch: 2
+        pad1_position: 'bottom-left' # 'top-left' | 'bottom-left' -> pin 2 always to the right of pin 1
+        mounting_pad_size: [1.67, 2.9]
+        # x position mounting inner mounting pad edge relative to nearest pad center
+        center_pad_to_mounting_pad_edge: 2.22
+        # y dimensions for pad given relative to mounting pad edge
+        rel_pad_y_outside_edge: 8.15
+        rel_pad_y_inside_edge: 6.7
+        pad_size_x: 1.1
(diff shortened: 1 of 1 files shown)


Werni2A:
Hi @justyn nice work but shouldn't the pin numbering be inversed?

The datasheet shows pin (circuit) 1 on the rigth in top view whereas in your footprint pin 1 is on the left in top view.
image
image

justyn:
@Werni2A that is a great point, I can't believe I didn't spot that.

However, the same applies to the existing [1.5mm Pico-Lock 504050 parts](https://www.molex.com/pdm_docs/sd/5040500291_sd.pdf). Their pin numbering is also reversed (see images below).

If we add the 2.0mm pitch versions with the correct pin numbering it would be pretty inconsistent with the 1.5mm version. We presumably can't change the pin numbering on the existing footprints because it would break existing designs.

What is done in this situation? Add a new set of footprints for the 1.5mm versions with the numbering corrected, existing alongside the originals?

image
image

Werni2A:
That's a good point. I think the existing footprints should be corrected. Because of the breaking change it might be possible for Kicad V6 but I'm not a librarian, so it's up to them to decide.

justyn:
I've tentatively pushed a new commit to fix the ordering on these (2mm) parts to match the datasheet, but probably a plan for the 1.5mm parts is needed before this is applied.

It would be pretty hellish to have two sizes of the Pico-Lock family with opposite pin numbering.

justyn:
chschlue [indicated](https://github.com/KiCad/kicad-footprints/issues/2465#issuecomment-691679864) that the incorrect ordering on the 1.5mm parts is a severe bug, I have submitted fixes to the script values [here](https://github.com/pointhi/kicad-footprint-generator/pull/628) and the footprints [here](https://github.com/KiCad/kicad-footprints/pull/2466).

I assume then that this pull request for the 2mm parts can be merged (unless there are other issues).

justyn:
I've updated the screenshots to reflect the new rotated footprints with fixed pin numbering.

PR #628, opened 2020-09-13
Fix pin numbering in 1.5mm pitch Pico-Lock connectors (504050 series)

justyn:
The existing values produce reversed ordering for pad numbering when
compared against datasheet, and the physical circuit 1 indicator on the
connector itself.

Pad 1 is the right-most pad in the orientation shown on the datasheet
(insertion from top of diagram).

Because the smd_single_row_plus_mounting_pad.py script always
places pad 1 on the left, this patch changes the rotation of the generated
footprints by 180 degrees, fixing the pin numbering.

The connector family page is [here](https://www.molex.com/molex/search/deepSearch?pQuery=category%253A%2522PCB%2BHeaders%2522%2540orientation%253A%2522Right%2BAngle%2522%2540pitchmatinginterface%253A%25221.50mm%2522%2540productname%253A%2522Pico-Lock%2522), the datasheet [here](https://www.molex.com/pdm_docs/sd/5040500291_sd.pdf).

This fixes https://github.com/KiCad/kicad-footprints/issues/2465, PR submitted here: https://github.com/KiCad/kicad-footprints/pull/2466

image

image

### DIFF
--- a/scripts/Connector/Connector_SMD_single_row_plus_mounting_pad/conn_molex.yaml
+++ b/scripts/Connector/Connector_SMD_single_row_plus_mounting_pad/conn_molex.yaml
@@ -318,7 +318,7 @@ device_definition:
         pinrange: ['list', [4,5,6,7,8,10,12]]
         text_inside_pos: 'center'
         pitch: 1.5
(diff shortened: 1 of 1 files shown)


chschlue (approved):
Thanks.

PR #629, opened 2020-09-15
Add HVQFN-32

Kev-J:
Hi,

As discussed [on this kicad-footprint PR](https://github.com/KiCad/kicad-footprints/pull/2439#issuecomment-690460784), here is my PR to add the HVQFN-32 script.

Cheers,
Kevin

### DIFF
--- a/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/qfn/hvqfn.yaml
+++ b/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/qfn/hvqfn.yaml
@@ -144,3 +144,57 @@ HVQFN-24-1EP_4x4mm_P0.5mm_EP2.6x2.6mm:
   #pad_length_addition: 0.5
   #suffix: '_Pad{pad_x:.2f}x{pad_y:.2f}mm_HandSolder'
   #include_suffix_in_3dpath: 'False'
+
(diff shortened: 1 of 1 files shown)


cpresser:
During my review I learned that this part is also called `MO-220`
https://www.jedec.org/system/files/docs/MO-220K01.pdf
(either make an account or use bugmenot.com)


That footprint is already library as `QFN-32-1EP_5x5mm_P0.5mm_EP3.1x3.1mm`

For me it seems like the only difference here is the name. Or did I miss something.

cpresser (commented):
The content of this file looks good. All values match the datasheet.
Can you please remove the commented lines to reduce noise?

Kev-J:
Good to know. Yes, this footprint has many names (SOT617-1 as well).

Yes, the footprint is exactly the same. I'm no expert, but I assume that the HVQFN packages are the same as QFN with better heat sink. I created this one because of the review of my symbol (https://github.com/KiCad/kicad-symbols/pull/2884).

Maybe an alias would be appropriate but I don't know if it is possible for footprints.

Comments have been removed.

Cheers,

chschlue:
> Maybe an alias would be appropriate but I don't know if it is possible for footprints.

That's my preferred solution, too, but it requires application support.
Alias names would reside within the footprint file so they all have to be read and cached in advance to be able to search for them, for example.

Kev-J:
Yes, and the 3D package would be linked to the alias to support this present case: packages sharing the same footprint but not the same physical package.

cpresser (approved):

PR #633, opened 2020-09-18
Added definitions for Texas PWSON-N8 (WSON) Package

jamesrlucas:
Addition of definitions for Texas PWSON-N8 (WSON) Package with a 1.45mm x 2.4mm Exposed Pad (different size to the other Texas PWSON-N8 package already in the library)

Just a copy and paste, name change and thermal pad size change. I've left everything else as-is, as there's no other dimensional differences on the datasheet

Footprint PR is [Here](https://github.com/KiCad/kicad-footprints/pull/2223)


cpresser (commented):
I checked all values. They are good.
Can you please remove the unused lines? They just add noise.

jamesrlucas:
Done :)

cpresser (approved):

PR #636, opened 2020-09-22
Add HTSSOP-44 with TopEP

evanshultz:
Footprint PR: https://github.com/KiCad/kicad-footprints/pull/2468

https://www.ti.com/lit/ds/symlink/tpa3251.pdf


myfreescalewebpage:
Footprint is merged

PR #638, opened 2020-09-24
Add WQFN-20-1EP_3x3mm_P0.4mm_EP1.7x1.7mm_ThermalVias

cronJ:
As suggested in https://github.com/KiCad/kicad-footprints/pull/2111 I tried to create the footprint, by adding the necessary data into the YAML file.

I couldn't find a documentation of all the available properties, so there are some deviations from the suggested land pattern. There should be five thermal vias, but currently I only could implement four.

The EP should be 1.8x1.8 mm according to the data sheet, but when I check the footprint in the GUI it is only 1.7x1.7 mm. Maybe someone can give me a hint here.


myfreescalewebpage:
One comment right now, the name of the footprint has no ThermalVIa in the title and EP size to be fixed: "WQFN-20-1EP_3x3mm_P0.4mm_EP1.7x1.7mm_ThermalVias" => "WQFN-20-1EP_3x3mm_P0.4mm_EP1.8x1.8mm"

myfreescalewebpage:
Footprint has been merged

PR #641, opened 2020-09-25
Add Infineon_LFBGA-292_17x17mm

cronJ:
grafik

https://www.infineon.com/cms/en/product/packages/PG-LFBGA/PG-LFBGA-292-11/


cpresser:
The datasheet is weird. I cant see if the ball-size is 0.5 or 0.35mm.
image

Also, it seems this is a SolderMaskDefined example footprint. However, I see a postive mask_margin in the changeset.

chschlue:
The drawing is... highly original, but I _think_ they suggest 0.5mm mask, 0.35mm copper, making it NSMD.
(The hatched "blob" is supposed to be mask and if you look closely, the diagonal lines stop at the 0.5mm circles, so that's supposedly the mask cutout.)

cpresser:
I think you are right.. I checked the PDF again, one can zoom in:
image

IPC suggest a 0.4mm pad for a 0.5mm ball
image
But the numbers are pretty close.

@cronJ Please get rid of the mask_margin parameter. Then the mask will be controlled by the board settings.

cpresser (approved):

cpresser:
Is there also a footprint PR? If yes, please update it with the generator output and ping me so I can merge it.

PR #644, opened 2020-09-29
Add Microchip_FCVG484_19x19mm_Layout22x22_P0.8mm and

pthomas:
Microchip_FCG1152_35x35mm_Layout34x34_P1.0mm

These packages are defined for the PolarFire SoC's
in ug0902 here:
https://www.microsemi.com/document-portal/doc_download/1244577-ug0902-polarfire-soc-fpga-packaging-and-pin-descriptions-user-guide

---
I wasn't sure about the names, should the Microchip prefix be there? Should FCG & FCVG be there?
Here's the Microchip_FCVG484_19x19mm_Layout22x22_P0.8mm:
Screenshot_20200928_211917

And the Microchip_FCG1152_35x35mm_Layout34x34_P1.0mm:
image

### DIFF
--- a/scripts/Packages/Package_BGA/bga.yaml
+++ b/scripts/Packages/Package_BGA/bga.yaml
@@ -132,6 +132,30 @@ Microchip_TFBGA-196_11x11mm_Layout14x14_P0.75mm_SMD:
   paste_margin: 0.000001
   layout_x: 14
   layout_y: 14
+FCVG484_19x19mm_Layout22x22_P0.8mm:
+  description: "FCVG484"
+  size_source: "https://www.microsemi.com/document-portal/doc_download/1244577-ug0902-polarfire-soc-fpga-packaging-and-pin-descriptions-user-guide"
+  body_size_x: 19.0
+  body_size_y: 19.0
+  overall_height: 2.92
+  pitch: 0.8
+  ball_type: 'collapsible'
+  ball_diameter: 0.5
+  layout_x: 22
+  layout_y: 22
+FCG1152_35x35mm_Layout34x34_P1.0mm:
+  description: "FCG1152"
+  size_source: "https://www.microsemi.com/document-portal/doc_download/1244577-ug0902-polarfire-soc-fpga-packaging-and-pin-descriptions-user-guide"
+  body_size_x: 35.0
+  body_size_y: 35.0
+  overall_height: 2.8
+  pitch: 1.0
+  ball_type: 'collapsible'
+  ball_diameter: 0.65
+  layout_x: 34
+  layout_y: 34
+  row_skips: [[1, 34], [], [], [], [], [], [], [], [], [], [], [], [], [], [], [], [],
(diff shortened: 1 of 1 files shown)


myfreescalewebpage:
@pthomas comment about the script (I'm not 100% confident so if you have a reason why this is like that tell me):

Microchip_FCVG484_19x19mm_Layout22x22_P0.8mm
- [ ] pad diameter is 0.5mm +/- 0.05mm in the datasheet, not 0.4mm

Microchip_FCG1152_35x35mm_Layout34x34_P1.0mm
- [ ] overall_height should be 2.8 +/- 0.19mm
- [ ] pad_diameter: 0.64 +0.06/-0.14mm in the datasheet, not 0.51mm

@chschlue if you can have a look to :
mask_margin: 0.015 => this one is not clear for me and we find various values in the script for BGA packages. Is it computed from something else ?
paste_margin: 0.000001 => probably it should not be 0 ? Other BGA have the same value. Probably Ok even if I'm not sure what is the purpose of this field.

Joel

chschlue:
0.000001 is a workaround addressing a limitation of the footprint format (0 in the file means "inherit from board setup" so if you actually want zero, you need to use a very small value instead.) Other than that, I believe @cpresser is more knowledgeable WRT the BGA generator.

cp-aquila:
First, it needs to be decided if this is going to be a IPC-compliant footprint, or should it be build according to vendor spec. O even mix both (which I think is horrible).
In this case, I don't see why it should be vendor specific. Its does not have any special properties, and the pitch is rather coarse.

Thus, I would suggest to
* specify `ball_diameter` instead of `pad_diameter`.
* keep all margins at 0 to use board defaults
* that will generate a NSMD footprint, as recommened by the manufacturer

To answer the original question. @chschlue is correct. The `mask_margin` will directly be applied to the footprint. It is not set on a per-pad-level.

pthomas:
Thanks for looking at this everyone.

On Thu, Oct 1, 2020 at 6:45 AM cp-aquila <notifications@github.com> wrote:

> First, it needs to be decided if this is going to be a IPC-compliant
> footprint, or should it be build according to vendor spec. O even mix both
> (which I think is horrible).
> In this case, I don't see why it should be vendor specific. Its does not
> have any special properties, and the pitch is rather coarse.
>
> Thus, I would suggest to
>
>    - specify ball_diameter instead of pad_diameter.
>    - keep all margins at 0 to use board defaults
>    - that will generate a NSMD footprint, as recommened by the
>    manufacturer
>
> So if we do IPC-compliant and specify ball diameter what will the pad
diameter end up being? The tables lists the ball diameter as TBD, but I do
see in the drawing it shows it, so it would end up looking like this for
the 484:
  description: "FCVG484"
...
  body_size_x: 19.0
  body_size_y: 19.0
  overall_height: 2.92
  pitch: 0.8
  ball_diameter: 0.5
  mask_margin: 0
  paste_margin: 0
  layout_x: 22
  layout_y: 22

-Paul

cp-aquila:
You can omit the mask_margin and paste_margin parameters. Which is the same as having them at zero.

